PR #81, opened 2014-12-23
Now viewer list is loaded from method store.

rsutormin:
And arguments are constructed automatically based on method specs for these viewers.

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/css/kbaseNarrative.css
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/css/kbaseNarrative.css
@@ -1633,6 +1633,9 @@ For data in KBase workspace
 .modal-body {
     max-height: 100%;
 }
+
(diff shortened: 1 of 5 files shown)


dangunter:
Matt,

It's not clear to me whether landing pages are supposed to work in the bitches brew we have now of dev and prod. We are running a dev instance out of the prod workspace. The URLs are being constructed like this:

 link += "/functional-site/#/" + landing_page_url_prefix + "/" + o.ws_name + "/" + o.name

where the magical 'landing_page_url_prefix' is coming out of the method store. I copied this from previous code (by Roman I think). Given the cut-and-paste nature of this, I am inferring that the magic of constructing landing page URLs is just in people's heads, whereas it should properly be in some common piece of JS with some documentation. If this is the case, then please point me to the appropriate file and function, we should use it.

Anyways, my experience running this off my laptop is that landing page links may sometimes go to a page, but very little of that page actually loads. Maybe this is totally normal behavior from a laptop and it is only possible to load landing pages from actual narrative-dev instances..

PR #83, opened 2014-12-24
Continuation of pull request #80, this time against correct branch (develop)

dangunter:
See commentary in original pull request: https://github.com/kbase/narrative/pull/80


briehl:
Hey Dan,
I just tried to merge develop into this branch to catch it up before going the other way, and found a number of errors and unexpected files getting updated. Would you mind double checking things a bit? I don't want to clobber any actual code changes.

dangunter:
Yes, there were a number of fixes to things including improving our install process.

dangunter:
OK, I merged in the develop branch. I use the "curenv-install.sh" exclusively now. I think we need to make that the main install script and deprecate "install.sh".

Issue #116, opened 2015-01-08
Debug helper


briehl:
We need a way to help debug what gets sent to the NJS. In the short term (like, this evening), it'll be possible to add a way to view the result on the front end, either just dumping in the JS console or something slightly more heavy weight - there was talk of including a 'debug' mode that would expose a couple buttons or something that could be used to view the result from sending an app/method.

A *better* option (slightly longer term, maybe by the end of the week) would be to wrap up the functions that are used to fetch the specs and call the apps into a Python module that can be invoked right from the code cells in the Narrative. This would lose the UI elements (which would need to be further tested separately), but would let devs rapidly test things.

For tonight (1/7), I'll address the first, short-term solution.

briehl:
Another debug helper was added for submitted jobs that might cause an error on their way through the NJS (e.g. the NJS might accept the job, but fail before returning an id). In the Javascript console of your browser, a debug message is emitted with a job's JSON document before submitting it. I also put together a narrative that shows how to submit jobs directly to the NJS with a couple lines of Python:
https://narrative-dev.kbase.us/narrative/ws.3286.obj.1

I feel this is helpful enough for now, so I'm closing this issue. It might get reopened during next week's session.

PR #117, opened 2015-01-08
Add debug mode for app jobs

briehl:
This adds a couple of things, related to issue #116.

The first is a 'mode' configuration in config.json (currently set to debug).

The second is an additional cell menu option that appears in debug mode - at the top of the menu is a 'View Job Submission' option. Now, when an app job is run in debug mode, all of the results of that initial set up (all communication from the kernel) is stored in that app cell's metadata. Clicking that new menu item will do two things:
  1. This will dump the contents of that info into the Javascript console (including a few message strings)
  2. Create and run a code cell that contains the job info that gets returned on a successful job submission. Having this cell auto run registers that object as a variable, and prints it out in a nice way (it's a substantial hunk of JSON, usually).

That job info object has three main properties: 'app' is the document that gets sent to the NJS, 'app_state' is the initial state object that gets returned, and 'job_id' is the job id that gets registered with the Narrative.

Hopefully this will make some debugging easier until a more elegant debugger can be assembled.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -25,6 +25,7 @@
     }, 
     "git_commit_hash": "bfbb9a8", 
     "git_commit_time": "Wed Nov 12 15:34:42 2014 -0800", 
+    "mode": "debug",
     "name": "KBase Narrative", 
     "prod": {
         "awe": "http://140.221.67.185:7080", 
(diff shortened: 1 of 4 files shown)


briehl:
I'm gonna jump ahead and put this in, so I can get the Narrative redeployed tonight on narrative-dev.

PR #118, opened 2015-01-08
framework for data type icons. one for contigs and default

dangunter:
Modifies the "icons" for the data types to use images in `images/data-icons`. There is a JSON config file `data-icons.json` (in same dir as `config.json` et al.) that specifies the mapping, which is based on the bare type name such as "ContigSet".  If no match is found, a default double-helix icon is used, with the background color chosen as before (so at least different types have different backgrounds even if they have the same icon).

This could be merged right now, without breaking anything, but I want to wait until I have a few more icons defined.

Also: if the config is not read, it fails hard and stupid with an alert box and redirect to `/`. I couldn't figure out how to directly load our HTTP `error.html` ?!


briehl:
The error.html is another templated thing using Jinja2, so the Tornado backend populates it.

We should probably have some sort of "dead end" error page for major fatal errors (not loading needed configs, Workspace/Shock is down, etc.). Is there a KBase-wide error page we should redirect to?

kkellerlbl:
Maybe Erik has a generic error page?  There's a 404 page but that isn't appropriate for a backend error.

dangunter:
I will move the error discussion to another issue / pull-request since I think it might be confusing in retrospect to have it under the "data icons" commits.

Issue #123, opened 2015-01-09
Change narrative dev install script to curenv-install.sh


dangunter:
The current install script doesn't work if you already are in a virtual env, whereas curenv-install.sh works in both places. It is also cleaner and easier to use. So, rename install.sh to old-install.sh and curenv-install.sh to install.sh. Some of the options from install.sh no longer work, but it's not clear that anybody was using those anyways.

dangunter:
See branch https://github.com/kbase/narrative/tree/dev-install-script

briehl:
This has been merged in, but the deployment script still uses the old one for a little while. Docker seems to complain something or other. I'll fix it up this weekend.

PR #124, opened 2015-01-09
new install script and docs

dangunter:
As described in Issue #123 

This change means that the current "install.sh" is renamed to "old-install.sh"
and the current "curenv-install.sh" is renamed to "install.sh"

Docs are updated appropriately. The underlying reason is that it is less error-prone and easier in the long run to let the developer set up their virtualenv than to try to automagically create that for them.


scanon:
Is this mainly for developers?  I don't think we are using this in our deploy right now. 

Shane

PR #126, opened 2015-01-09
Added some docs and screenshots to help deal with app errors.

briehl:
As the title says, I wrote a help document that goes into some detail the points at which loading/running/retrieving apps can fail, and hopefully a little insight into how to understand what's failing and why.

It might be incomplete, so any feedback and additions are very welcome!


briehl:
Adding this to the branch for release now. It's all new, so there should be no conflicts.

PR #137, opened 2015-01-14
communities widgets updates and additions

teharrison:


briehl:
Hey guys, this merge is fine for now, but could you update your workspace links to use the endpoint in config.json? That's in the global namespace, so you can access it via javascript from

    window.kbconfig.urls.workspace

We're in the process of removing all hard-coded links so it's easier to configure for the CI and Next machines, so adding fewer would be helpful. Thanks!

teharrison:
will do.

Issue #125, opened 2015-01-09
Need page and handling of fatal UI errors


dangunter:
Need to have some sort of "dead end" error page for major fatal errors (not loading needed configs, Workspace/Shock is down, etc.). The occurrence of these errors should be logged to our global narrative logging service (the log proxy running on the narrative host).

dangunter:
Logging of errors (and fatal errors) is done in branch https://github.com/kbase/narrative/tree/dev-status, see commit 1b5d9dd40c7a2585a261ceddf0d4170ceaa4cc0f

briehl:
It looks like, for logging from the front end, we just have to call KBError(source, message) for non-fatal errors and KBFail(source, message) for fatal ones?

dangunter:
almost. KBError, yes. KBFatal is the equivalent. KBFail is the "generic" one they both call.

The method is actually commented in the code!

briehl:
Guess I should've read those instead of skimming their use!
Thanks for this!

PR #143, opened 2015-01-15
Log error/failure messages from UI through kernel

dangunter:
For issue #125


dangunter:
Crap went against master again

PR #151, opened 2015-01-17
Better job management/canceling support, Fixes to App and Method cells to deal with cancellations

briehl:
This has a number of changes to how App and Method cells behave. Now:
  - Running Apps and Methods have a "Cancel" button associated with them. Clicking that will restore the cell to its input state, and cancel (and delete) the running job
  - Deleting App/Method cells now prompts to delete, warning the user that it'll kill any running job (if they've been run before)
  - Deleting jobs now triggers the call to the right back end services to delete the running job itself.
  - Deleting a job from a running cell now unlocks that cell so its inputs can be used again.
  - Added new error checking for different states
    - A dangling job with no associated cell should properly error.
    - A dangling job on the front end that's been deleted on the back should properly error.
    - A job in an error state (not the above, but a runtime error) should now reflect an error on its associated cell (if present)
  - Fixed an error where running a method automatically made an output cell connected to incomplete data.
  - Refactored Jobs panel to only look up jobs that are incomplete (e.g. not 'error', 'completed', 'done', or 'deleted')

This pull request also has a couple of cosmetic changes:
  - Toolbars should only appear over IPython cells now (KBase cells have their own menus)
  - Styled app/method control buttons to be closer to the style guide / mockup.

### DIFF
--- a/src/biokbase/narrative/common/kbjob_manager.py
+++ b/src/biokbase/narrative/common/kbjob_manager.py
@@ -163,7 +163,7 @@ class KBjobManager():
         workspace = os.environ['KB_WORKSPACE_ID']
         return _method_get_state(workspace, token, URLS, self, method_info[0], method_info[1], method_job_id)
 
-    def delete_jobs(self, job_list):
+    def delete_jobs(self, job_list, as_json=False):
         """
         Delete all jobs in the list. They may belong to different services based on their prefix, currently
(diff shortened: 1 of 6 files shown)


briehl:
In the interest of time, I'm going to go ahead and merge this (with --no-ff so it can be rolled back easily), so I can redeploy quickly.

PR #171, opened 2015-01-21
Revised widgets

nconrad:
Add revised fba and model output widgets to use the new "kbaseTabTable.js" output widget.  This requires PR https://github.com/kbase/narrative_method_specs/pull/86

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "modules/ui-common"]
 	path = modules/ui-common
 	url = https://github.com/kbase/ui-common
+[submodule "modules/revised-widgets"]
+	path = modules/revised-widgets
+	url = https://github.com/kbase/ui-common/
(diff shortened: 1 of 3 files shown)


briehl:
It looks like you're trying to add a whole new submodule copying ui-common again. Is this correct? I feel like this is a bad idea (it's an extra 66 MB of data that's being copied).

I think it might be okay to provisionally accept, with the caveat that we really need to sort out the issues between the branches before Friday. We should only have one copy of ui-common as part of the Narrative.

nconrad:
That's correct.  We are only looking to have it provisionally accepted.  The submodule is called "revised widgets"

I would argue that a 20% increase in size (which could only affect deployment time) is not a good reason.  This method is already better than what currently exists, which is not having *anything* in sync.  This has resulted in numerous tickets and confusion on why data is displayed differently in the narrative vs the ~~workspace browser~~ ~~landing pages~~...  whatever will (or won't) exist next in ui-common. 

It was tested against narrative sandbox and I saw no major issues, although it is hard to say now.  Any testing there is appreciated.  It is the model and fba widget and we need people testing those so that any issues can be ironed out.

briehl:
Checking this branch out doesn't seem to break anything. I'll merge and deploy this on Narrative-dev now, and we can figure out where to go with the twin ui-commons tomorrow.

PR #174, opened 2015-01-21
Minimize app and method cells

dangunter:
Still need to do method cells, but app cells alone will be useful.
Also added "running" spinner that shows whether cell is minimized or not.
Also added "error" icon for failed runs.


dangunter:
OK, method cells are done too. Last thing is state saving, but let's make this a separate issue

PR #182, opened 2015-01-22
adding FBA output view back in

nconrad:


briehl:
I don't think your revised-widgets branch is up to date with the other path fixes. They should all be something like: `kbase/js/revised-widgets/src/widgets/modeling/`

PR #183, opened 2015-01-22
Increased height of data slide out

msneddon:
Part of NAR-439


msneddon:
whoops - misread ticket.  made height too big... arg.

msneddon:
Ok, should be right now.

nconrad:
For what it is worth, on a 15inch laptop, I think the slide-out panel is quite big now.  It takes up most of the screen...

PR #192, opened 2015-01-23
refresh button added to data panel slideout

msneddon:


msneddon:
see https://atlassian.kbase.us/browse/NAR-503

PR #201, opened 2015-01-24
There's now a little red badge in the Jobs tab header with a count of ho...

briehl:
This counts the number of incomplete jobs - those that are queued, in-progress, running, etc.


briehl:
This fixes https://atlassian.kbase.us/browse/NAR-432

PR #203, opened 2015-01-24
Suggested Next Steps after running an app or method

dangunter:
See JIRA issue https://atlassian.kbase.us/browse/NAR-423

After a person runs an app or method, adds "suggested next steps" to the bottom of the output boxes. 

Current list set of app/methods to list is hardcoded in function called getNextSteps() in both kbaseNarrativeAppCell.js and kbaseNarrativeMethodCell.js . Only a few apps and a few methods have anything at all right now.


msneddon:
By the way, we can encode this information in the method/app specs as well now (https://github.com/kbase/narrative_method_store/blob/master/NarrativeMethodStore.spec#L91), although none of the app/method specs have this information populated.  It's probably worth switching at some point so we can update the suggestions without redeploying the narrative.

PR #206, opened 2015-01-26
Switched out old widget for new in gene_domains, and added production server

jmchandonia:


jmchandonia:
This is partially redundant with #202, which also adds a link to kbaseDomainAnnotation

Issue #101, opened 2015-01-06
Data panel: filter apps from selected data


dangunter:
Clicking should highlight a data item and filter apps/methods by that type as input

briehl:
Done by Mike.

Issue #100, opened 2015-01-06
Data panel: "+/-" button should be turned to “...”


dangunter:

briehl:
Done already by Mike.

PR #208, opened 2015-01-27
widget updates

teharrison:


briehl:
I'm a little confused about the update to proxy_mgr.lua - this adds a maximum number of containers, right? So, if user #101 tries to log in, do they just get booted with an error? At the very least, shouldn't we expand that to some really high number?

teharrison:
My bad, the default should be insanely high for now (which would duplicate
the current state of no max).  The 100 is what I had when testing.

Will update the pull request to 5000.  This code change does not require
the nginx config be updated, that just allows the defaults to be set in the
config.  Will still work with old config.

teharrison:
Forgot to mention, the max is a soft cap and not a hard cap. User #100
would cause there to be no provisioned containers, but user #101 would get
a container, just has to wait on it building.

briehl:
I see, that's just how many are sitting around waiting. No problem, just clarifying that's what you wanted to pull.

PR #222, opened 2015-01-29
Websocket text change

briehl:
The IPython websocket error doesn't have the same meaning in the KBase Narrative, and needed to be updated. This PR does so, and removes the 'OK' button (since that doesn't really have meaning either - it's not like the user can just reboot their instance and reconnect to it).


briehl:
Also, this fixes JIRA NAR-444 (and probably others).

PR #227, opened 2015-01-30
Added a concatenated, minified form of all kbase client api libraries. T...

briehl:
...he full size version with compiler is in a separate PR for ui-common.

I shuffled the older clients into an 'old' subdirectory for now, in case this makes things crash and burn. Hopefully it won't.!


briehl:
This also contains a fix for https://atlassian.kbase.us/browse/NAR-579

PR #249, opened 2015-02-04
removed button to open the gallery

thomasoniii:
We're going to hold off on deploying the gallery until we get more icons and screenshots, so this just drops the button to open it from the UI.

Widgets, etc. still exist for future deployment.


briehl:
Oops, I think we both did this at the same time, so that's already solved (Paramvir came by and asked me to hide it). Thanks, though!

PR #253, opened 2015-02-04
Fixed bug in data list filtering when searching and switching between type filters

msneddon:


psdehal:
Do you really want that console.log statement in there?

PR #262, opened 2015-02-07
Job tracking and autosaving

briehl:
Changes to how job info is tracked, serialized, and saved.
- A job's last status should be tracked in the following structure in the Narrative object metadata (stored as a JSON string under the job_info key:
```
{
     'running': number,
     'completed': number,
     'error': number,
     'queue_time': number,
     'run_time': number
}
```
The 'running', 'completed', and 'error' properties all refer to the state of the jobs currently tracked by that Narrative - i.e., those that would pop up in the Jobs panel. 'queue_time' and 'run_time' is cumulative across all jobs that have ever reported those. NJS-based jobs report both, but UJS-based jobs (I think only the species tree methods? Maybe domains?) only report run time. All values should be >= 0.

- Autosaves are now done when a job's state causes a new cell to be created, whether it's output or error. This should eliminate the condition where a user watches a job finish, populate the screen, closes the page before saving, then comes back to find that his job info went away on the backend somewhere, and doesn't get to see results. Or, worse, the results returned get corrupted somehow.

- This also injects a generalized Narrative object (see kbaseNarrative.js) under the IPython namespace where the rest of the main code lives. This gives access to the following:
  - `IPython.narrative.saveNarrative()` to trigger a save. Soon, this wrapper will block saving unless the user gives a new Narrative a name other than 'Untitled'
  - `IPython.narrative.checkVersion()` will trigger a manual version check update
  - `IPython.narrative.narrController` is the main controller widget, and commands can be sent straight to it.
  - `IPython.narrative.readonly` is a boolean flag
  - And there's room for more in the future. Or this might get moved to a somewhat separate `KBase.narrative` namespace.


briehl:
Also noting here that when a Job finishes, the "status" row turns into "Status: Completed XX ago", eg. "Completed 8 days ago". The "Started: XX ago" becomes "Run time: XX", e.g. "Run time: 15.4 min".

PR #265, opened 2015-02-07
Nar 651 fix

briehl:
Fixes two things.
1. If the user has no valid narrative objects to look up, it just doesn't render anything.
2. Fixed an issue where the narratives panel was loading twice on page load.


briehl:
see https://atlassian.kbase.us/browse/NAR-651

PR #275, opened 2015-02-10
Read-only or "view-only" mode

dangunter:
Bunch of changes to implement a view-only mode, including the ability to toggle this mode even when the user has the appropriate permissions:

- check for permissions at startup, and periodically thereafter, allowing narrative to update its state "live" in response to sharing changes
- add functional (although not beautifully implemented) "copy" button that creates a copy of a read-only narrative and then provides a button to allow the user to jump to it. This required some refactoring of the copy-narrative logic that was in the narrative control panel; really that code needs a lot more cleanup
- add a small button/icon that lets users with read/write access to toggle to view-only mode. this was originally used for debugging, but playing with it makes me think it will be useful to reclaim screen real estate when narratives are being browsed


dangunter:
Bill, I'm going to wait for you to merge this. I have been staring at it for too long today, and may be missing something silly.

PR #277, opened 2015-02-10
Clarified the installation instructions

jmchandonia:
Clarified the installation instructions, so that somebody who hasn't installed Narrative before has a recipe to follow

Replaced obsolete repos, and added notes on installing a missing dependency.

Edited nginx default config so it will work with functional site and narrative simultaneously, and
has the same lua path as README-dev.md

### DIFF
--- a/docs/README-dev.md
+++ b/docs/README-dev.md
@@ -12,6 +12,7 @@
 ## Narrative development
 
 Last updated: Dan Gunter <dkgunter@lbl.gov> 7/31/2014
+Minor update by John-Marc Chandonia <jmchandonia@lbl.gov> 2/3/2015
 
 To report bugs, login using your KBase developer user/pass at: https://atlassian.kbase.us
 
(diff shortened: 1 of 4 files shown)


docs/README-dev.md
```diff
@@ -70,7 +71,7 @@ First, however, you'll need to get a necessary submodule initialized:
 If there are any problems with this step, run the following commands:
 
     cd modules/ui-common
-    git checkout hardy
+    git checkout narrative-dev
```
briehl: This subrepo still needs 'hardy' - it's not the same as what's deployed. Not yet, at least.

docs/README-dev.md
```diff
@@ -217,22 +221,30 @@ Next, bring in the Narrative and Bootstrap repos (as yourself):
     git clone http://github.com/kbase/bootstrap
     git clone http://github.com/kbase/ui-common
 
+# note that as of 2/10/15, narrative-test.kbase.us is using the
+# develop branch of both narrative and ui-common
```
briehl: The 'staging' branch of both narrative and ui-common is deployed on narrative-test.kbase.us

docs/README-dev.md
```diff
@@ -217,22 +221,30 @@ Next, bring in the Narrative and Bootstrap repos (as yourself):
...
+    apt-get install -y  python-software-properties
     add-apt-repository ppa:nginx/stable
     apt-get update
     apt-get install -y nginx-extras
     # verify this worked before going on
     exit
 
+## Install the right version of Nodejs (including npm)
+
+   sudo add-apt-repository ppa:chris-lea/node.js
```
briehl: There's an official ppa, now - see here: https://github.com/joyent/node/wiki/installing-node.js-via-package-manager

Also, that should only be for Ubuntu 12.04 builds. I thought I updated the instructions to grab a 14.04 Ubuntu image, though I could be wrong. Either way, 14.04 would be a better target.

PR #287, opened 2015-02-12
Added proper parameter checks for max/min ints and floats in parameters

msneddon:
related to https://atlassian.kbase.us/browse/KBASE-1586


briehl:
Do these get communicated in the spec documentation? Or do they just throw an error if a user tries something out of bounds?

msneddon:
The documentation does not get automatically updated (we should probably update the yaml for the construct tree methods), but the error message is fairly good I think and appears as the user enters in a number that is out of bounds.  Before these limits were just not checked on the front-end.

briehl:
Ok. Looks like this will fix an issue that can tear down services, so I'll merge it.

PR #295, opened 2015-02-13
Fix for rare problem with "My data" and "Shared with me" tabs. Lazy loading in genome widget.

rsutormin:


briehl:
This has a couple of critical (even blocker) fixes.
1. Fixes a problem where the data importing panel is unable to load.
2. Fixes an issue where the genome widget loads an entire genome before displaying. This can block or freeze browsers if a Narrative has several large genomes in it.

I feel this is definitely worth merging and testing today, and deploying tonight.

PR #296, opened 2015-02-16
Hide step errors

briehl:
This does two things.
1. Hides the 'step_errors' results (the little red '!' by the status) unless the status is an explicit error.
2. Fixes a critical issue where widgets occasionally wouldn't load.


briehl:
Forgot to mention that this addresses a blocker ticket: https://atlassian.kbase.us/browse/NAR-716

PR #297, opened 2015-02-17
Dropdown to filter by data type is now sorted

msneddon:
A minor feature that's been requested several times now.  See: https://atlassian.kbase.us/browse/KBASE-1607

Note that longer term the plan is to organize these types into groups to make it easier to find types when you don't know the name of a type beforehand.


briehl:
This has been merged and deployed onto narrative-dev.

PR #298, opened 2015-02-18
1.0.1 Production release

briehl:
This has a number of critical fixes from the past few days. These are covered in the release notes (RELEASE_NOTES.md), but I'll copy them here:
- Widgets
  - JIRA NAR-699 - profile viewer sometimes failed
  - JIRA NAR-700 - run button in custom input for "Create metagenome set" didn't work
  - Make the genome viewer widget do "lazy loading" - it shouldn't load the whole contig set until the user tries to look at contigs/genes
  - Fixed an intermittent issue where many widgets were unable to load and would cause the page to hang
- Data panel
  - Sometimes the "My Data" and "Shared with Me" tabs didn't load properly
  - The available types in those panels were unsorted and confusing
- Jobs panel
  - JIRA NAR-716 "Errors" shouldn't be shown on a job with non-error status


briehl:
These have been tested throughout the day on narrative-dev (version 1.0.1e). But some more eyes would be useful!

PR #301, opened 2015-02-19
added in lengths for transcriptomes; loads up locations to do cross chec...

thomasoniii:
added in lengths for transcriptomes; loads up locations to do extra step in transcriptome check.


briehl:
This is another critical patch to get both plant transcriptomes and genomes to view properly.

PR #302, opened 2015-02-19
Fix for search of meta-genomes in public data tab.

rsutormin:
Please change narrative config.json to test it (before new Search API goes into Prod) like:
-------------------------
        "search": "http://dev01.berkeley.kbase.us/search/getResults",
-------------------------


briehl:
For reference, the SSL'd endpoint is `https://dev01.kbase.us/search/getResults`

PR #308, opened 2015-03-09
Added base directory for Selenium test scripts.

briehl:
This adds a testing harness for the Narrative that uses Selenium to automatically start and run apps and methods. An example for an app and method is supplied.

Obviously, there's much more work to be done, but this is a start.


briehl:
This is very low impact, and includes a base directory for adding more tests to various functionality as we go. Pulling this in, but we can make changes as necessary.

PR #311, opened 2015-03-17
Fixes for some narrative bugs 

rsutormin:


briehl:
I'll pull this now to get it deployed on narrative-dev, but I suspect Nomi and the docs team will have suggestions for some of the text changes.

PR #312, opened 2015-03-18
NAR-783 Change update button text

briehl:
This changes the update button text to read "Get new version!" which is more actionable, and clearly prompts the user to click there, as opposed to the more passive "New Version Available!" text.


briehl:
This update is safe, and only affects the text of the update button. It also has a bugfix where an error while updating wasn't properly being logged. Merging.

PR #315, opened 2015-04-06
Modify save error messaging.

briehl:
This change does two things.

1. It pulls out some hacky things that were injected into the IPython notebooks savewidget.js, and puts them where they belong.

2. It addresses a special error condition we've come across (https://atlassian.kbase.us/browse/KBASE-1843) where a Narrative object that is too large gets rejected by Nginx with a 413 error (Request entity too large). Currently, the maximum Narrative object size is set to 4 MB, which is pretty substantial, but we should have a discussion to decide what's actually reasonable. The Workspace can handle objects of nearly any size - several megabytes is no big deal - but this does create a performance cost on the front end.

Partly, I'm of the "caveat emptor" mindset that we should let the user do what they like and absorb the consequences, but this might also put undue pressure on the rest of the system to be performant in a way that isn't fair. For example, it might make sense to append large TIFF images of slides or tissue sections (for example), or videos of colony growth, but those should probably not be part of the Narrative object itself.

Likewise, if a user fetches a large data object via the API and looks at it in a code cell (the method I used to replicate the error and deal with it), that should probably not be saved separately. Then again, users will be able to do all sorts of things with code cells and shouldn't really be stopped by maximum Narrative sizes.

I think it's still up for discussion what an appropriate max size should be for a Narrative object, but these code changes will support it. Currently, this is set as an Nginx parameter and needs to be configured in kbaseNarrative.js. The next PR will factor that out into an external configuration. I think it's worth refactoring the config file itself all together, which is why that's not part of this.

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/kbaseNarrative.js
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/kbaseNarrative.js
@@ -454,9 +454,9 @@ Narrative.prototype.init = function() {
         // currently - 4/6/2015 - there's a hard limit of 4MB per KBase Narrative.
         // Any larger object will throw a 413 error, and we need to show some text.
         if (data.xhr.status === 413) {
-            errorText = 'Due to current system constraints, a Narrative may not exceed 4 MB of text.<br>' +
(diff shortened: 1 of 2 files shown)


msneddon:
A small change- everything still seems to load up and work as expected so I'm pulling this in.

PR #317, opened 2015-04-18
Require.js migration

briehl:
Sorry for the beyond-the-pale huge commit. But this is like pulling off a bandage - it's gonna hurt either way, and yanking it off all at once will help deal with changes quickly.

**Goal of these changes:**
The main goal in these edits was to migrate all KBase front-end code to be loaded asynchronously via Require.js. You can read more about Require here: http://requirejs.org/. This gives us a few huge benefits.

1. Ability to manage load order more specifically, and ensure that events happen at the right time.
2. Load only the Javascript files (i.e. visualization widgets and supporting code) needed in the current view.
3. Enforce modularization of visualization code that will eventually allow 3rd party (and 1st party!) developers to write and deploy widgets without needing to compile the whole Narrative apparatus.
4. Assist with unit and integration testing by enabling test directives that only load explicitly necessary code.
5. Helps with the path toward migrating all visualization widget code out of the narrative repo and into somewhere more global (ui-common, or maybe another yet-to-be-created spot).

Encouraging developers to write widgets as asynchronous module definitions (AMD) lets us manage dependencies in a clear, sustainable way. This and similar methods (CommonJS, Browserify, etc.) have been a web standard for several years now, and a version of this is going to be a component of ECMAScript 6.

Doing this migration was relatively easy, yet tedious, but required a number of accessory fixes to ensure that things work right. These are detailed below.

**List of changes:**

**KBase-developed code changes:**

1. Wrapped all used widgets in this repo with define() statements. This lets them be easily loaded via Require.
2. Fleshed out a config file (currently named "narrative_paths.js") with the list of path names for each widget to be asynchronously loaded. This also contains the initial load order for the Narrative app itself.
  * I was hoping to avoid this, but the version of IPython that the Narrative uses (1.x branch) doesn't load most of its with Require. To ensure the load order is right, the notebook/main.js and custom.js are no longer loaded inline, but via Require
  * This is changed in the more recent versions and later updates will deal with that.
3. Rewrote the Gruntfile compilation script to use a Require.js directive. 
  * This now compiles only the core Narrative components into a minified, revision-tagged file.
  * It also emits a source map - this lets the web browser's console unravel minified code, so it's actually possible to trace errors in production.
4. Refactored the main kbaseNarrative.js
  * Factored out global error functions and startup activities - button bindings, etc. - to the kbaseNarrativePrestart module.
  * Factored the remaining kbaseNarrative module into an object that obeys load order.
5. Refactored the kbaseNarrativeDataCell. This was having issues with a global variable that it uses to load all viewers once.
6. Refactored the Method, App, Output, and Viewer cells to load their inner-widgets asynchronously, but due to how they currently sit in Narratives, they have to be loaded inline. This is done with the Require shim apparatus: http://requirejs.org/docs/api.html#config-shim
7. A few viewer widgets are also currently loaded via shim (FBA model widgets and Metagenome widget dependencies).

**IPython code changes:**

1. Extracted Narrative changes to IPython code. Most of these were in notebook/main.js to deal with load order issues. Those have been removed.
  * There still exist other changes to IPython code to support the Narrative, but these are mostly cosmetic, mainly changing text in popups dialogs.
2. Removed notebook/js/main.js and custom/custom.js from the inline load order. These are now loaded via Require.

**External code changes:**

1. Temporarily removed datavis.js. This module isn't currently being used, and contains a lot of duplicate code of various different versions - including jQuery and Bootstrap - that create problems when loaded up. It needs to be recompiled/adjusted to work well.
2. Added a separate underscore.js dependency that other widgets were using.
3. Added the Require DomReady extension - this forces Require to wait until the DOM is loaded (e.g. all IPython code and other dependencies) before executing.
4. Wrapped dataTables.bootstrap.js to play nice when loaded with Require as part of a minified package.
5. Commented out the jquery-ui tooltip code since it conflicts with Bootstrap (this is temporary, pending examination of how we're using jquery-ui and whether we need a version update.)
6. Extracted CSS code from the KNHX tree code into the kbaseNarrative.css stylesheet. This was being loaded via document.write, which (1) is a stylistic no-no, and (2) didn't work when loaded asynchronously.

**Finally...**
While I feel this will help make testing easier, there are no automated tests included here yet. That should be the next focus. In the meantime, this was tested using the methods done leading up to the February demo - running lots and lots of Narratives and examining each viewer and app. That's really not a perfect metric, and I'm sorry for that. Better regression testing will come soon.

And, again, sorry for the monstrous PR. But the best way to get this done without too many code conflicts is all at once.


briehl:
Hey, all ( @dangunter @msneddon @thomasoniii @nconrad @rsutormin @teharrison - you guys will probably be affected the most ), I'd appreciate a few looks over this if you can. It's a big scary pull, but I think it'll help put us on the path to cleaning up a bunch of code in this repo and ui-common, and getting a testing framework set up.

@thomasoniii - this might also deprecate the kb_define workaround you put in, so the widgets in ui_common/hardy could all just use the define() command instead.

As an aside - this doesn't actually affect the code in the ui-common branch subrepos. The widgets from ui-common/hardy show up okay, and the ones from ui-common/master are included by shim in the narrative_paths.js file.

Thanks!

msneddon:
I started doing some tests the other day, and so far everything I looked at seemed to work correctly.  The narrative starts up as expected, I'm able to do basic things like upload/download data, save a narrative, add data to a narrative, view that data, and run a few methods with correctly rendered output widgets.

I still wanted to poke around a few more methods and apps.  It would still be good though if a few other people take a look at widgets they care about too.  If a few more people have a look and say ok, I can merge this in early next week.  Sound like a reasonable plan Bill?

briehl:
Yup, sounds great. That's why I poked people. :)

rsutormin:
Looks good to me. I haven't tested all widgets though. Only those I'm familiar with (upload, download, genome annotation, proteome cmp, trees, genome sets). We probably should have some testing scenario for each widget.

thomasoniii:
Some cursory inspection looks like it's okay. I have questions about the
future, though -

Yes, if the narrative has moved to a using require everywhere, then we
could get rid of the kb_require nonsense for inside of there...BUT, how
will this work for landing pages? AFAIWA, angular doesn't quite play nicely
with require. So maybe we want the hack in there for those purposes?

Likewise, where should widgets and things be living at this point? The
hardy branch of ui-common was started solely to be a place for me to put
widgets that had been require'ed w/o messing up master. So I could
presumably merge that back into master and have everything in master (and
then we could remove redundant widgets from one of the repos)....but I
don't know if that's going to goof up anything else depending upon the
widgets in ui-common.

Presumably if the narrative is finally making a stand on require, then all
other locations using those widgets should demand it as well (landing
pages? Is there anything else?), but I think that's outside of my scope
right now.

Maybe we toss the hardy branch and just move the master copies of the
widgets into the narrative repo? I'm sure this requires further discussion
to get it done right.

briehl:
I'd like to get all widgets loaded with Require, everywhere. This should let us essentially store the widgets anywhere we want and merge their Require config with each page's config as needed (which can be done on page load, as long as we know where that config file is stored).

I wanted to do this and get it stable with the Narrative first, since it's pretty well encapsulated, then move on to the landing pages, etc.

For getting this working with Angular, there are a couple of approaches. We can go all in with it and wrap all the Angular code into AMDs to get loaded through Require like [here](http://www.sitepoint.com/using-requirejs-angularjs-applications/) or [here](http://marcoslin.github.io/angularAMD/#/home).

I feel the big benefit we get from using Require for the widgets is to only load the widgets we need. So another option is to use a config file that points to all widgets, and load them in a require() block as needed. Something like:

```
angular.module('directive') {
.directive('my-directive', function($location) {
    return {
        link: function(scope, element, attr) {
            require('myWidget', function() {
                $(element).myWidget({ ... });
            });
        }
    };
});
```

Which should do the lazy loading for us. For the record, @eapearson is already doing similar things on the main page dashboard.

briehl:
And to answer your other question - once we get this change worked out (along with other changes to the ui-common repo), then AMDified widgets should all live on the main branch, whatever that winds up being. There's discussion on a different PR [here](https://github.com/kbase/project_guides/pull/43) about how to structure code development/deployment.

msneddon:
Ok, a few people have taken a look now and all seems good.  The travis-ci build failed just because the travis config file is missing on this branch and travis isn't fully configured yet.  I'm pulling this in.

PR #319, opened 2015-05-04
Updates to text files

briehl:
Updates release notes to be caught up with the backlog of issues currently resolved, updated Docker installation directions in the developer deployment doc.


briehl:
These are safe and low-impact. Just catching up some documentation. Going ahead and merging.

PR #318, opened 2015-04-28
Base image fix

briehl:
This addresses https://atlassian.kbase.us/browse/NAR-391 where the base Docker image was failing to build. There were a couple of main errors going on here:

1. The R repository used was set to .org instead of .edu, so it was defaulting to Ubuntu, meaning that most packages were out of date.
2. The automatic pip installation through apt (apt-get install python-pip) results in an unusable pip. Updating it with easy_install fixes it for now, will revert once the official Ubuntu fix is made.
3. An earlier bug (that was actually fixed! but masked by the above two) was fixed. The public ANL Ubuntu mirror was taken down - this was replaced by the PNL mirror.

PR #320, opened 2015-05-08
fix CRAN URL in Dockerfile.base

kkellerlbl:
A bum URL was in the dockerfile for CRAN at berkeley.  It should be to cran.cnr.berkeley.edu.


briehl:
Yup! Fixed in the other PR that's not accepted yet. :) See #318

kkellerlbl:
Oops, sorry!  I'll just close this.

PR #322, opened 2015-05-11
First pass at getting Travis-CI set up for the narrative.

briehl:
This includes a .travis.yml file, and modifications to the Makefile and Gruntfile to handle testing. Since tests all run against PRs, and it's easy enough to update a PR before it's accepted/closed, this will be ongoing for a little while.

This one will be a little tricky because it contains multiple languages to test. This sets the base .travis.yml file to use NodeJS for testing, and will (eventually) install the Python dependencies in the script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ branches:
     - master
 before_install:
   - git clone git://github.com/kbase/bootstrap
-  - pip install -r bootstrap/kb_python_runtime/python-pip-list-narrative
+  - sudo pip install -r bootstrap/kb_python_runtime/python-pip-list-narrative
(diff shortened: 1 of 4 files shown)


dangunter:
Looks like it works. Have you considered merging this one, to keep the PR smaller, and expanding on it with future ones?

briehl:
Yup, that was my plan. Travis-CI was a little wobbly yesterday, so I went home before this finished. I just wanted to get it to turn green before accepting the PR, so I didn't have to make a new one with each change.

All these changes do are locally build the Narrative in the test environment, then run stub test scripts that just return 0. Those'll get fleshed out in another PR.

PR #323, opened 2015-05-18
Fix error text when a narrative is unavailable

briehl:
See https://atlassian.kbase.us/browse/KBASE-2042

When a Narrative is unavailable - whether it doesn't exist, or a user doesn't have permission to see it - this template is displayed. This fixes a typo and a couple out of date links on that page.


briehl:
Merging this so it can be deployed on narrative-dev for testing - it was previously tested on a local VM with no trouble. Also it's only textual changes, so it's pretty safe.

PR #324, opened 2015-05-18
Added Travis-CI and Coveralls badges to the readme.

briehl:
Minor fix. Now that Travis and Coveralls are set up to deal with the Narrative, this adds a status badge for each of the branches of note - develop, staging, and master.


briehl:
This is pretty tiny and only affects the readme structure. Accepting it, but it can be revised later if folks don't like it.

PR #325, opened 2015-05-21
Updates the narrative method store client.

briehl:
This updates the Narrative Method Store client to expose three new methods: validate_method, validate_app, and validate_type.

This is pretty harmless, as it doesn't affect any existing methods. However, those are only available on ci.kbase.us right now.


briehl:
Merging this now so narrative-dev can get updated tonight.

PR #326, opened 2015-06-25
Add ci and next sections to config.json

kkellerlbl:
It might be helpful to have separate config sections for -next and -ci.  I'm not sure if this is the best way to do that.  :)


briehl:
I'm planning on doing a broader refactor of the configuration. After chatting with Shane a bit, it would be useful to have that config.json generated from a standard deploy.cfg file that'll work with your deploy tools. There might be some other fiddly bits that are needed, too, especially with the ui-common refactor moving along.

Either way, this'll be useful right now. Going ahead and merging.

PR #327, opened 2015-07-02
Update output cells with horizontal scrollbar

briehl:
If an output cell's content gets too wide (e.g. the abundance data table), it should be embedded in a scrollable table. Right now, it just flows on past everything.

So far, this only affects the abundance data table, but it could hit others. So, instead of fixing it there, I added the fix to the output widget container.

However, this exposes an issue with widgets that make use of the Bootstrap grid layout, but don't embed those in a .container classed div. They all get a tiny scrollbar. Those widgets will get fixed in a separate PR since I think they mainly live in ui-common anyway.


briehl:
This is a minor enough change, and there are no complaints, so I'm bringing this into develop for testing on narrative-ci.

PR #328, opened 2015-07-10
add mesa-common-dev to package list

kkellerlbl:
One of the R packages requires OpenGL headers, which are in the mesa-common-dev package.  Added the package.  I also moved the R build to be last; the hope is failures in previous steps abort before doing the R step, which takes an insanely long time.


briehl:
Yup, good idea. Not sure why mesa-common-dev wasn't in there before. But thanks for adding it!

kkellerlbl:
It was in there before but somehow disappeared.  I think it is still in the master version.

PR #329, opened 2015-07-11
Config change in docker.lua

kkellerlbl:
There may have been a change in how the VolumesFrom config option is interpreted by the lua code.  It used to be that an empty string would work, but the variable is supposed to be an array, and the empty string is now an error.  json.util.null populates VolumesFrom with an empty array.


briehl:
This is for a recent version of Docker that we need. Merging!

PR #330, opened 2015-07-17
Move bootstrap deps

briehl:
This moves the files that the narrative docker image build was dependent on from the bootstrap repo. Pointers from the Dockerfile and the buildNarrativeContainer.sh script were updated as well.

I also modded the install script to use the narrative directory as a variable. Too many times, I've run into problems with creating a clone called 'narrative-fork' or 'narrative-test' and it mucks with the install process.

Solves JIRA ticket NAR-809


briehl:
This is obsoleted by other changes, and certainly by upcoming changes with the Jupyter Notebook update.

PR #331, opened 2015-07-18
Nar 804 race condition

briehl:
This addresses NAR-804 - a race condition failure that sometimes happens on Narrative startup.

The issue is that the KBase code is loaded asynchronously, but still requires that IPython code is present, and certain events are fired at the right time. Specifically, it was set that the Narrative code waited for the first status_idle.Kernel event to be kicked off before starting up. This ensured that everything was in place before starting, including auth info and workspace info.

This PR fixes the problem by doing two things.
1. Load the KBase startup code in custom.js before running the IPython main function. This registers the needed event listeners at the right time.
2. Use the notebook_loaded.Notebook event, which is fired once when the notebook is successfully loaded.

This will all be obsolete when the upgrade to IPython 3 is complete, since that loads everything asynchronously and allows the dependencies to be much more explicit. But this should patch things for a release.

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/custom/custom.js
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/custom/custom.js
@@ -47,7 +47,7 @@
  */
 
 console.log('Loading KBase Narrative setup routine.');
-$([IPython.events]).on('app_initialized.NotebookApp', function() {
+$([IPython.events]).one('notebook_loaded.Notebook', function() {
     console.log('Performing narrative startup');
     require(['kbaseNarrative'], function(Narrative) {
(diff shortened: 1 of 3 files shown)


kkellerlbl:
I can merge this unless anyone else wants to review it.

kkellerlbl:
Actually, is most of 314-369 just change in indentation?  It looks like a big change but from what I can tell it looks like just the elimination of a block.  If it's more complicated than that I should read it more carefully before merging.  :)

briehl:
Yup, that's all whitespace. That's a great trick, Sam!
The main change there is that the encapsulating event that triggered that block was moved to a separate module, which removes a few chained events. The rest is just unindenting that block.

PR #337, opened 2015-07-28
WIP: Preliminary IPython 3x upgrade

briehl:
This is on the path toward updating the Narrative to be backed by IPython 3.x, and, eventually, newer forms of Jupyter (though that'll take a heavier refactor of the front end to keep up with their push toward TypeScript and PhosphorJS).

Right now, this updates the backend to extend the right classes that allow communication with the KBase Workspace instead of the local filesystem. It also includes an updated login/logout system that persists (or destroys) KBase auth credentials. Finally, it's a path toward both updating the front end to work with the newer IPython, and a refactor of the front end to do things more sensibly.

Still remaining to be done:
* Testing! The backend tests are on the way, but incomplete. The frontend tests are nearly non-existent.
* Updating CSS - the styles that KBase extended from have changed pretty dramatically
* Reconcile Require config with ui-common - this will let us deploy ui-common separately, and not a separate thing that's part of the Narrative deployment
* Others i'm forgetting as part of this ticket: https://atlassian.kbase.us/browse/KBASE-2245


briehl:
If there are no complaints, I'd like to merge this in (to its own separate branch) so I can keep working without appending more commits to this PR.

briehl:
Merging ahead so it'll be easier for others to help put this together.

PR #338, opened 2015-07-28
Duplicate a narrative method with parameter values

msneddon:
I've been working out how to automatically create a method with parameter values set in the Narrative via the JS api, and have a potentially useful feature here to copy a narrative method including parameter values.

This could be useful because we can copy narrative methods even after they've been run.  So it allows a user to essentially rerun something without reentering all the parameters.

This could probably be done for apps too, but is slightly trickier because we need to find a way to ensure the the app input widgets are all fully loaded before we update state.

Thoughts on this feature?


briehl:
I think it's a good idea. There's copying built in to IPython (which we're hiding...), but it might not easily apply to KBase cells with inline parameters.

I'm working on bringing back the IPython menubar and all that it supports in a sensible way, but we'll still need something (for the near future) that effectively copies populated method cells. This is a good start.

PR #340, opened 2015-07-29
API function for running a method cell

msneddon:
This small change adds a runMethod function to kbaseNarrativeMethodCell so that we can run a method cell without user interaction.  This is a step towards being able to automatically create a method, set parameters, and run the method, which will be useful for some of the feature-value user stories.


briehl:
Looks good to me.

nconrad:
This sounds pretty cool!

PR #341, opened 2015-07-30
UI widgets for the expression sprint

psnovichkov:


briehl:
These are all the expression/feature-value things we talked about before, right? And the Highcharts package?

msneddon:
Pavel can confirm, but yes- these should just be the new feature-value widgets and the Highcharts package dependency.

briehl:
Ok. I just want to make a note here that we're (slowly) moving toward fetching those through bower/npm as part of the build process. But for this rapid prototyping sprint, I'm okay with keeping it.

Hopefully, by the end of next week, we can refactor some of that out.

msneddon:
This is out of date now, so I'm closing this PR.

PR #342, opened 2015-07-30
Two fixes related to bugs in build-featureset method.

rsutormin:
1) Fix for new lines in text area method parameters;
2) Fix for null values in method parameters under ref transformations.


briehl:
Yup, you showed me this today. Thanks, Roman.

PR #343, opened 2015-07-30
Minimize methods on run and add summary text

msneddon:
This update consists of:

1) When a method is run, it is automatically collapsed, and 'replacement' text is shown that allows us to briefly summarized the step the user performed.  The replacement text is pulled from the replacement text field in the narrative method specification (for example, here: https://github.com/kbase/narrative_method_specs_ci/blob/master/methods/reannotate_microbial_genome/display.yaml#L14).  This was already a feature in the narrative method store, so it will just work if you update the specs.  The cool thing is that the replacement text uses http://handlebarsjs.com so you can include parameter values in the replacement text (and do things like have conditional expressions, show the length of an array input, etc)

2) When an app cell runs, the method input fields are collapsed.  I choose to not yet add the replacement text in the case of Apps, because the 'Steps' are displayed more prominently, but that is something we can easily add.  Thoughts on this?  Also, should we replace the App header when it is collapsed with some other information?

3) Added user names to the timestamp when methods and apps are run, so you can see who executed a particular method/app.  The user name links to the user profile page.  I think this small feature will be very useful when multiple people are editing a single narrative.

4) When working on the above, there was some minor bugs that I found in updating the error state of methods/apps which I fixed.  There are still issues here I think in connecting the jobs panel to the state of the apps/methods in the narrative cells, but at least now things are somewhat improved.

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/css/kbaseNarrative.css
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/css/kbaseNarrative.css
@@ -1623,7 +1623,7 @@ div#notebook_panel {
 }
 
 .app-panel-heading {
-    padding-top: 20px;
+    padding-top: 10px;
 }
 
 .app-panel-heading h1 {
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/narrative_core/kbaseNarrativeAppCell.js
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/narrative_core/kbaseNarrativeAppCell.js
@@ -178,7 +178,7 @@
                               .append('Run')
                               .click(
                                   $.proxy(function(event) {
-                                      var submittedText = "submitted on "+this.readableTimestamp(new Date().getTime());
+                                      var submittedText = "&nbsp;&nbsp; submitted on "+this.readableTimestamp(new Date().getTime());
                                       if(this.auth()) {
                                           if(this.auth().user_id)
                                               submittedText += ' by <a href="functional-site/#/people/'+this.auth().user_id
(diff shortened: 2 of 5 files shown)


msneddon:
Here is an example screenshot of running the 'Annotate Microbial Genome' method.  Notice that the cell is collapsed, the text is replaced with a string that includes the name of a parameter value, and the user name is included after the timestamp.

screen shot 2015-07-30 at 4 36 16 pm

msneddon:
And here is an example of app steps collapsed after submitting an App:

screen shot 2015-07-30 at 4 30 33 pm

aparkin:
Nice

Sent from mobile device and I have clumsy fingers.




On Thu, Jul 30, 2015 at 4:54 PM -0700, "Michael Sneddon" <notifications@github.com> wrote:












Here is an example screenshot of running the 'Annotate Microbial Genome' method.  Notice that the cell is collapsed, the text is replaced with a string that includes the name of a parameter value, and the user name is included after the timestamp.

briehl:
Looks great! That's a nice feature we've talked about for a while. Are there stubs for all of the method specs to know what their templates should be? What happens if there's an empty string or something there?

msneddon:
Good questions.  

The 'replacement-text' field in method specs is optional, so if it isn't there then the field will just be missing in the specification returned by the narrative method store.  There aren't stubs in place in the existing methods.  We should just add them to methods as needed following this example:   
https://github.com/kbase/narrative_method_specs_ci/blob/master/methods/reannotate_microbial_genome/display.yaml#L14

If there isn't a 'replacement-text' defined in the spec, it will default to displaying the method name, as it did before.  Here is where the decision is made: https://github.com/msneddon/narrative/blob/msneddon-minimize-methods/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/narrative_core/kbaseNarrativeMethodCell.js#L499

So if the replacement-text is missing, or empty, or the method hasn't been run yet, then the minimized view will just show the Method name.

briehl:
I see. I missed that part in the code. Thanks! This is worth merging and testing on ci.

PR #348, opened 2015-08-02
Missing dependencies in method cell

briehl:
The method cell had some missing dependencies that weren't always loaded up at the right time, as found on narrative-ci. This should fix it.

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/narrative_core/kbaseNarrativeMethodCell.js
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/narrative_core/kbaseNarrativeMethodCell.js
@@ -10,7 +10,7 @@
  */
 
 (function( $, undefined ) {
-  require(['jquery', 'kbwidget'], function($) {
+  require(['jquery', 'kbwidget', 'kbaseAuthenticatedWidget'], function($) {
     $.KBWidget({
         name: "kbaseNarrativeMethodCell",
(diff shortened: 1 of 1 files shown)


src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/narrative_core/kbaseNarrativeMethodCell.js
```diff
@@ -10,7 +10,11 @@
  */
 
 (function( $, undefined ) {
-  require(['jquery', 'kbwidget'], function($) {
+  require(['jquery', 
+           'handlebars', 
+           'kbwidget', 
+           'kbaseAuthenticatedWidget',
+           'kbaseNarrativeCellMenu'], function($, Handlebars) {
```
msneddon: Why is the "Handlebars" parameter passed into the widget? This gives me an error locally because the Handlebars parameter overwrites the global Handlebars JS and is left undefined when this widget is created. If I just remove 'Handlebars' as a function parameter, it works as expected.

And sorry for not updating the require dependencies... I'm still getting used to using require in widgets, and figured it just wouldn't work if they aren't correct.

I also now realize that we need to add the kbaseAuthenticatedWidget dependency in kbaseNarrativeAppCell. I can do that in a separate PR.

msneddon: Ok, here is the App Cell dependency update:
https://github.com/kbase/narrative/pull/349

PR #349, opened 2015-08-04
updated app cell dependencies

msneddon:


briehl:
Great. Thanks! But there are also a couple other dependencies - the kbaseNarrativeOutputCell (at line 948, and one other) and kbaseAccordion (line 151). These should be added, too.

msneddon:
Ah ok.  I'll add those too.

msneddon:
I assume we don't need to include kbaseNarrativeError because that widget is not invoked here, just passed on to the NarrativeOutputCell widget?

msneddon:
also removed the Handlebars input parameter to the method cell widget...

PR #350, opened 2015-08-04
Added support for placeholder text in Textarea method parameters

msneddon:
Previously placeholder text was only supported in text field parameters, not textareas.


briehl:
Looks good.

PR #353, opened 2015-08-05
Fix compiler loadtime

briehl:
Should actually fix the page load timing. Works on narrative-dev with lots and lots of page refreshes.


msneddon:
I think there is still something in a require config that is searching for static/narrativeMain.js?

I'm getting a 404 on trying to find that resource on the backend and in the browser:
2015-08-11 13:57:49.048 [tornado.access] WARNING | 404 GET /static/narrativeMain.js (127.0.0.1) 0.63ms

Would any of the tests catch that?  Because it did pass tests.

briehl:
Grr. I think some commits got crossed that shouldn't have been. Thanks for catching that.

briehl:
Ok, that should be fixed now. The problem was I was only testing the compiled version, not the uncompiled. The compiler saw what the paths were just fine, but they were off in the main page.

msneddon:
Great.  It looks like it all loads correctly now, and basic things work, so I'm accepting.

PR #354, opened 2015-08-05
JavaScript client for KBaseFeatureValues service + 2 widgets that do not have plot viz

psnovichkov:


briehl:
Looks good.
Just making a note here that we need to update the process for adding new KBase clients. That needs an overhaul overall (in how we build and use clients), but there is a separate building and distribution process.

It's just really broken right now. Hopefully it can be cleaned up in the next sprint of the "easing development" campaign...

PR #356, opened 2015-08-07
New subdata autocomplete method parameter type

msneddon:
This introduces a new type of parameter that allows autocomplete of fields based on subdata selection of a workspace data object.  This is useful, for instance, for selecting features from a genome or elements of a model.  It is generalized (built on the workspace generalized subdata query), so you should be able to select anything in a list/map in a WS data object, and use those values to populate the dropdown.  The WS data object can either be one selected in another parameter, or a fixed object(s) (by WS reference id) which I think will be useful for instance if we want to force selections from a controlled vocabulary.

The Narrative Method Store was updated to support the new parameter, and I setup an example method specification in CI here.  You can see in the spec how the subdata fields are specified.  https://github.com/kbase/narrative_method_specs_ci/tree/master/methods/example_subdata

There are a variety of options for display described in basic detail here:
https://github.com/kbase/narrative_method_store/blob/develop/NarrativeMethodStore.spec#L212

And here are some screenshots from the example.

After a user selects an input model, the second parameter allows search and autocomplete of the set of reactions in that model.  Notice that we can also specify some of the fields for a description, so it doesn't just have to be a list of IDs.
screen shot 2015-08-06 at 5 05 56 pm

The parameter also allows you to group several selections in a single input box rather than separate lines for each selection, which may sometimes be useful:
screen shot 2015-08-06 at 5 06 51 pm

Finally, you can select fields from an arbitrary object.  In this case I'm pulling something from the GramNegModelTemplate which is in a public workspace.
screen shot 2015-08-06 at 5 08 17 pm

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/function_input/kbaseNarrativeMethodInput.js
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/widgets/function_input/kbaseNarrativeMethodInput.js
@@ -4,7 +4,8 @@
  */
 define(['jquery', 'kbwidget', 'kbaseNarrativeInput', 'kbaseNarrativeParameterTextInput',
     'kbaseNarrativeParameterDropdownInput', 'kbaseNarrativeParameterCheckboxInput',
-    'kbaseNarrativeParameterTextareaInput', 'kbaseNarrativeParameterFileInput'],
+    'kbaseNarrativeParameterTextareaInput', 'kbaseNarrativeParameterFileInput',
+    'kbaseNarrativeParameterTextSubdataInput'],
     function( $ ) {
     $.KBWidget({
         name: "kbaseNarrativeMethodInput",
@@ -52,6 +53,7 @@ define(['jquery', 'kbwidget', 'kbaseNarrativeInput', 'kbaseNarrativeParameterTex
             for (var i=0; i<params.length; i++) {
                 var paramSpec = params[i];
                 var $stepDiv = $('<div>');
+
                 // check what kind of parameter here.
                 if (paramSpec.field_type === "text") {
                     var textInputWidget = $stepDiv["kbaseNarrativeParameterTextInput"]({loadingImage: this.options.loadingImage, parsedParameterSpec: params[i], isInSidePanel: this.options.isInSidePanel});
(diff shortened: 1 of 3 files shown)


psnovichkov:
Mike,

Nice! 

Few  comments:
 
1. To visualize a selection, I have to set selection_description, but the way the elements from this array are combined is hardcoded… It is always good to have a default solution, but to make it even more powerful, in addition to selection_description, it might a good idea to have something like ‘custom_selection_description’, like:

‘custom_selection_description’:
{
	‘pattern’: ‘Gene %s encoding %s’,
	‘items’ : [’name’, ‘function']
}


2. As we discussed before, it does not cover the use case that might be very common. Example: ExpressionMatrix… I want to select a subset of genes present in the matrix, but in the selection_description I want to show gene function that I have to grab from the Genome object referenced by this matrix. Thus:  "id from one object, description from another (or the same) object(s)".
Just looking at the spec.json  (https://github.com/kbase/narrative_method_specs_ci/blob/master/methods/example_subdata/spec.json <https://github.com/kbase/narrative_method_specs_ci/blob/master/methods/example_subdata/spec.json>) I would expect that you load several arrays (one for each property), and then join them by index. If instead I can specify the ‘key’ and ‘value’, we can easily support this use case .

The design could be:

      "subdata_selection": {
	“selection_id”: {
        	"parameter_id" : “expression_matrix_ref",
		 “subdata_included”: “data/row_ids/[*]/id”,   // Only one value! It is kid of  primary key
	},
	“selection_description”: {
        	"subdata_included" : [{
        		"parameter_id" : “genome_ref",
			label : ’name’,
			id: “features/[*]/id”,
			value: “features/[*]/name"
		},{
        		"parameter_id" : “genome_ref",
			label : ‘function’,
			id: “features/[*]/id”,
			value: “features/[*]/function"
		}],
		’selection_description’: [’name’, ‘function'']
	}
      }


3. Not clear why do you need “path_to_subdata” if you already have it in “subdata_included” ?

Best,
P.

nconrad:
Very nice.  I like the second one a lot.

msneddon:
@psnovichkov To answer your questions:

1) Great idea!  Yes, the description in the dropdowns should be templated. We already are using Handlebars for the replacement text of methods, so this is easy to add- and we'll get all those nice handlebars features for free.  I'll add that to this PR.

2) I know what you are trying to do :)  And I agree that it may be common.... but I think fetching extra information from a third object even from another parameter is actually fairly tricky and I think we should wait until we have more time to think it through.  First, it requires two subdata queries and some matching based on some shared field.  We don't strictly validate sub object references, so the IDs won't necessarily match up.  Second, I think in general not all subids will come from the same object (for instance in an arbitrary feature set, there may be features from multiple genomes).  I'm not sure the best way to specify that, and it could even require fetching from a large number of WS objects.  Third, I think there are many more edge cases and error handling logic we would need to add to support this well.  So I like the idea, but I think we should push it back.  We should also have not just the expression/feature example, but a few examples to make sure we can generally support most of the closely related user stories.

3) It is because there are actually two paths to keep track of - there is the WS subdata query so that you only fetch a subset of the object, and then there is the path to the array/map that contains the elements you want to populate.  Those don't have to be the same thing, for instance, if the description is not at the top level in that list.  This is also part of the reason for grouping all the "subdata_included" paths into one simple list- it matches the format of the WS query API.  The set of subdata paths have to be compatible together, and should be grouped together.

msneddon:
Ok @psnovichkov - we can now specify the descriptions using Handlebars templates.  For instance the template specified here:
https://github.com/kbase/narrative_method_specs_ci/blob/master/methods/example_subdata/spec.json#L76

Gets rendered like so:
screen shot 2015-08-07 at 10 14 40 am

briehl:
Looks great! Any more edits you want to make here? Safe to merge?

msneddon:
Nothing more right now. I think it's good and should be safe to merge.

psnovichkov:
Perfect!
P.

PR #359, opened 2015-08-11
Added Narrative API call to create and autorun a method

msneddon:
This change allows us to add a method, populate parameters, and run it through an API call.  This is useful for widgets that need to trigger some additional narrative action, like computing some statistics on data selected in a widget using an existing narrative method, or calling a method to create say a FeatureSet based on features selected in a widget.

This is an important because it means we still get nice usability without having to create heavyweight widgets that do analysis internally, making it easier for people to see what is happening in a narrative.

You can test it out directly from the JS console like so:
```
var method_id = 'example_subdata';
var parameters = {
        'input_model':'model',
        'input_rxns':['rxn1','rxn2'],
        'input_cmpds' : 'cmpds1,cmpds2',
        'input_template' : ['something']
    };
IPython.narrative.createAndRunMethod(method_id, parameters);
```

Note that in the future we probably want to add a way for KBase widgets to notify when it is finished rendering/loading, so that we can update the state safely.  Since each widget potentially does something different and loads in multiple async steps, and because widgets can be arbitrarily nested, I'm not sure exactly how to do that.  Here I just keep checking until the method input widget is defined and ready, which worked properly for all the methods I tested with.

### DIFF
--- a/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/kbaseNarrative.js
+++ b/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase/js/kbaseNarrative.js
@@ -406,5 +406,55 @@ Narrative.prototype.saveNarrative = function() {
     IPython.notebook.save_checkpoint();
 };
 
+/**
+ * @method
+ * @public
+ * Insert a new method into the narrative, set it as active, populate the
+ * parameters, and run it.  This is useful for widgets that need to trigger
+ * some additional narrative action, such as creating a FeatureSet from 
+ * a selected set of Features in a widget, or computing a statistic on a 
+ * subselection made from within a widget.
+ */
+Narrative.prototype.createAndRunMethod = function(method_id, parameters) {
+    //first make a request to get the method spec of a particular method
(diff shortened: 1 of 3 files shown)


msneddon:
I also just added a minor fix that reinitializes the 'delete cell' warning modal every time it shows.  There is a bug that detaches events associated with the modal after it has been shown a couple times.  This is really annoying when testing automatic method running...  

Probably somewhere we or bootstrap modals are using jquery remove instead of detach, or something like that: http://stackoverflow.com/questions/559147/are-events-lost-in-jquery-when-you-remove-an-element-and-append-it-elsewhere

briehl:
This is a great start. I'm a little concerned about relying on timeouts to wait for the input widget to be ready - the parent widget should probably accept a callback, or return a promise - but I think this is worth testing on CI.

Also noting here that the narrative controller kinda desperately needs a refactor - those should be exposed as methods instead of requiring events to be triggered.

The bug in the second part is pretty weird, since it only happens after the second or third use. I suspect it's like you said - some strange mismatch between Bootstrap's use of jQuery and ours. If this fixes it, great!

PR #363, opened 2015-08-12
Updates to feature-values widgets; Example data tab works in CI

msneddon:
Some minor edits to the expression widgets for basic expression matrix data viewing and exploration. This PR also updates the example tab so we can add example expression matrix data later, and so example data loads in CI or Next.


briehl:
Looks good.

PR #365, opened 2015-08-12
Narrative should fetch widget config from main site

briehl:
(the other half of https://github.com/kbase/ui-common/pull/153)

These changes allow the narrative to update its Require config using paths from the ui-common root. It expects to see a file there called 'widget-paths.js' that hold some configuration for those paths. On page load, it tries to fetch that (and silently continues if it cannot), then updates those widget paths so that they become absolute. So, for example:

```
'KBaseFBA.FBAModel'                     : 'src/widgets/modeling/KBaseFBA.FBAModel',
```

becomes (in production)

```
'KBaseFBA.FBAModel'                     : 'https://narrative.kbase.us/src/widgets/modeling/KBaseFBA.FBAModel',
```

and Require can load it up from there.

The reason for this is that we need a way to keep widgets only in ui-common so we can continue to build and deploy them from that one repo without an explicit subrepo in narrative that needs to get constantly updated.

Note that this only works for the FBA widgets right now (see that ui-common PR), but is a solid proof of principle that'll extend to the rest of them.


briehl:
After an offline discussion with Mike, we decided to merge this and get it tested on narrative-ci. We can roll back later and try to reconcile the subrepos if necessary.

PR #366, opened 2015-08-12
Typo fix

msneddon:
Arg!  Corrected a typo in an expression widget, of course just a minute after the other PR went in...


briehl:
D'oh! I should've caught that when I read through it, too...

PR #369, opened 2015-08-13
Updated submodule pointers and some paths

briehl:
This should include recent feature-value widget updates and some path adjustments to their location.


briehl:
merging this now to get it checked on ci.kbase.us

PR #370, opened 2015-08-13
From develop -> staging to test on next.kbase.us

briehl:
Merging all recent changes against the develop branch to staging so the narrative can be deployed in the next environment.


briehl:
This will pretty much need to be done by hand.

briehl:
Of all the things to cause a conflict - a little typo in the RELEASE_NOTES that has since been corrected...

PR #372, opened 2015-08-17
Update to work with Jupyter

briehl:
This is a first step toward updating the Narrative to work with Jupyter. It includes an installer for the Jupyter Notebook and ipywidgets that requires an existing virtualenv. It also cleans out some of the duplicate and unnecessary changes from the ipython 3.2 update branch.

Finally, this moves all code that was under /src/notebooks/ipython_profiles/profile_narrative to just /kbase-profile (though that name might change). It's a little bit easier to navigate like that...


kkellerlbl:
There are a bunch of files like this one with funny uuid-like filenames:

kbase-profile/security/kernel-1d8025b8-d935-4593-aea1-07a7891e89a0.json

Is that expected from the new profile?

briehl:
No, those shouldn't be there. I'll add them to the gitignore.

briehl:
Made a bunch of changes to keep up with what's been done in the develop branch.

briehl:
Sheesh. Realized this didn't get merged in. Since it's on a separate working branch, this is safe. Merging...

PR #377, opened 2015-08-22
Minor ui-common submodule update

msneddon:
Forgot to update to the latest commit in ui-common.  This is important for the updates to the chart widgets used for expression visualizations.


msneddon:
I'm pulling this in myself now so we can get it on CI for the weekend, but easy to revert if something goes wrong.

PR #378, opened 2015-08-22
Enable browsing of genes with basic stat and heatmap calculated for a set of conditions

psnovichkov:
Also, colorGenerator is optimized for the kbaseExpressionHeatmap


msneddon:
Let's not merge this in directly again.  We should pull into a common branch we've been using for feature-values, and add this together with all feature-value changes together into Narrative develop.

rsutormin:
I'm closing this PR since it was merged into https://github.com/msneddon/narrative/tree/msneddon-fv-widgets (which is supposed to be a target for future PR's in current feature_value sprints).

PR #379, opened 2015-08-25
No longer auto-quit when the Workspace service is unavailable

briehl:
See JIRA ticket: https://atlassian.kbase.us/browse/KBASE-2626

When the Workspace service is unavailable (due to misconfiguration, downtime, cosmic rays, etc.), a newly started Narrative will just bail out, and shut its IPython process right back down. A little annoying, maybe, but not really a problem unless you're running this as a process in a Docker container. Then it takes the container down with it, since the process finishes. Even worse, if you're provisioning a bunch of containers at once, and trying to maintain a queue of loaded containers, that'll cause some serious churn as they start up and shut down repeatedly.

This PR fixes that but inserting an error into the log, but it keeps IPython alive. Trying to fetch any narratives will likely result in a 404 (a restriction due to a number of reasons - mainly that this version of IPython doesn't deal with those types of errors very well. Fixed in the ease-dev-campaign branch that works against Jupyter 4.0), until the Workspace service comes back.

A follow up PR will amend the drop down in the browser to be able to shut down the browser. There's some finicky code restructuring that needs to take place to get that button to work without the rest of the Narrative apparatus being started up.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,11 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the IPython Notebook (more notes will follow).
 
+### Version 1.0.4
+__Changes__
+- Added widgets and methods to support feature-value analyses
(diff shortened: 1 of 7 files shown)


msneddon:
Looks good, tests pass, and seems to be working locally.

PR #382, opened 2015-08-26
Docker

scanon:
This may still need some work but I would like to start the review.


scanon:
Bill: I think the PR is okay to merge.  We may find some unexpected side affects on CI, but I think they can be quickly addressed.  Give me a heads up before you merge so we can test things out on CI.

briehl:
Sounds good, Shane. Let me try running the build process locally (or maybe on narrative-dev) to see if there are any kinks, then I'll merge it so see it on CI.

briehl:
Okay, I got most of it to build, then it choked here:

```
Step 7 : RUN git submodule update --init --recursive && rm -rf .git/modules/modules
 ---> Running in 7f9e103f48dc
INFO[0015] The command [/bin/sh -c git submodule update --init --recursive && rm -rf .git/modules/modules] returned a non-zero code: 1 
```

Looks like something may not have been moved into the image during the build? Here's the log from the failed process:

```
vagrant@precise64:~/narrative-docker$ sudo docker logs 7f9
fatal: Not a git repository: /home/vagrant/narrative-docker/.git/modules/modules/revised-widgets
Unable to find current revision in submodule path 'modules/revised-widgets'
```

PR #391, opened 2015-09-03
Staging the 1.0.5 release

briehl:


msneddon:
Tested and works on next, so good to go.

PR #397, opened 2015-09-23
Added kbaseBlastOutput widget code for Visualizing Blast output (GenomeUtil.BlastOutput)

pranjan77:


msneddon:
Looks good locally, so I'm pulling this in so we can test in CI.

A couple minor comments for @pranjan77 - we should add a scrollbar to the sequence alignment tab so that the cell only expands to a set size (could be adjustable), instead of expanding to fit all alignments, and we should add some kind of legend in the graphical alignment tab to indicate what the colors mean.

pranjan77:
Ok I will add scrollbar to sequence alignment tab and legend to graphical alignment tab in next iteration...

PR #400, opened 2015-09-24
Changes in narrative back-end supporting async methods of services...

rsutormin:
registered as docker images running through AWE.


briehl:
Will this disrupt any of the existing apps/methods? From the code change, it doesn't look like it, but I'd just like to check with you to be sure.

PR #403, opened 2015-09-28
Fix installer problem

briehl:
The installer wasn't installing the biokbase stack the right way. This should fix that (and install the right requirements as well).


briehl:
Merging this now so it can get built on CI. It's a minor enough change that it can be easily rolled back.

PR #404, opened 2015-10-06
Improved FeatureSet widget error handling

msneddon:
When features are not properly linked to a Genome, or cannot be found in a Genome, provide an error/warning per feature rather than an error that shows nothing.  This is useful for allowing the featureset object to still be inspected and for debugging.


msneddon:
Also added a one liner fix that should allow newline characters to pass correctly in textarea parameters.

PR #415, opened 2015-11-03
first version of support for new GenomeAnnotation and Assembly objects

realmarcin:


src/biokbase/narrative/common/kbtypes.py
```diff
@@ -738,12 +767,24 @@ class ContigSet(tls.Unicode, TypeMeta):
         class v1_1(tls.Unicode, TypeMeta):
             """Type spec for the "ContigSet" object"""
             info_text = "KBaseSearch.ContigSet-1.1"
+    class Assembly(tls.Unicode, TypeMeta):
```
mlhenderson: KBaseSearch.Assembly is not registered.  There is a separate ContigSet and Genome defined in Search for legacy reasons having to do with the Genome and ContigSet objects defined in KBaseGenomes.  The new Assembly and GenomeAnnotation types do not need separate Search types defined.

src/biokbase/narrative/common/kbtypes.py
```diff
@@ -738,12 +767,24 @@ class ContigSet(tls.Unicode, TypeMeta):
...
+        class v1_0(tls.Unicode, TypeMeta):
+            """Type spec for the "Assembly" object"""
+            info_text = "KBaseSearch.Assembly-1.0"
     class Genome(tls.Unicode, TypeMeta):
         """Genome type"""
         info_text = "KBaseSearch.Genome"
         class v2_0(tls.Unicode, TypeMeta):
             """Genome object holds much of the data relevant for a genome in KBase"""
             info_text = "KBaseSearch.Genome-2.0"
+    class GenomeAnnotation(tls.Unicode, TypeMeta):
```
mlhenderson: KBaseSearch.GenomeAnnotation is not a registered type and does not need to be defined here.

realmarcin: Thanks, I was a little overzealous there it turns out ...

So these will never be registered types?

realmarcin: I made a new PR removing this unnecessary code for Search.

This should be ready to deploy on CI now ...

mlhenderson:
Ok, let's merge this in and take a look on CI Narrative.

PR #420, opened 2015-11-05
Installer fixes

briehl:
Fixes problems with the local installer. It was trying to activate a virtual environment with a relative path while in a separate directory.


briehl:
Has a few fixes that are needed. Tests pass (this branch just tests installation and running right now....), so it's safe to merge.

PR #425, opened 2015-11-18
Updated config module

briehl:
The 'narrativeConfig' module now has two new utility functions:

`get`: returns any root-level key
`url`: returns anything formerly under kbconfig.urls.

as an example:
```javascript
define(['narrativeConfig'], function(Config) {
    var loadingImage = Config.get('loading_gif');
    var wsUrl = Config.url('workspace');
});
```

This PR applies this use to all of the core widgets. Trying to get away from relying on global variables to be present.


briehl:
Went through an applied the Config module to every module that needs it except for the 'function_output' directory. Also updated them all with 'use strict' and fixed issues that caused.

briehl:
The last several commits are catching up with the develop branch, and fixing tooltips across the page to use the new tooltip config. That config use looks like:

```javascript
// delay before showing tooltip in ms
Config.get('tooltip').showDelay
// delay before hiding tooltip in ms
Config.get('tooltip').hideDelay
```

This will all get moved to a KBase-specific tooltip utility function eventually, but for now this should help things be consistent.

briehl:
Finally tests on Travis pass!
Right now it just tests the building process and a few of the basic modules (loads the config, main, and login modules). That'll grow as time goes on.

Eventually, we'll get the selenium tests working in there, too.

PR #426, opened 2015-11-19
Set of changes to fix malformatted tooltips on the data list panel

eapearson:
The main problem was that the jquery-ui tooltip plugin was being loaded after the bootstrap one, so tooltips, at least on the data list panel (I didn't check elsewhere), were rendered in the jquery-ui style, which was in the incorrect format and also did not place tooltip content as html, so character entities were displayed as plain text. 
The fix for this was to set jquery-ui as a dependency of bootstrap, and to load bootstrap in the main loading script rather than jquery-ui.
Secondarily, there were minor display problems with the data list tooltip. On the first item of the list, the tooltip would display above the item, and was mostly hidden inside containing element. The solution was to set the tooltip to try to display above the item, but if that would cause the tooltip to bump into the container, display it below. To enable this, the "viewport" property needed to be set to a selector which would match the container for the data list. Because there is no specific id on this container, the selector used is not very solid - it selects the first container within the left column, so if that order changes, or if the class names change, this will break. Seems like it will be solid for now, though.
Finally, since setting the tooltip to "auto" will lead to arbitrary locations for the tooltip, the style overrides (basically setting the background color to blue from the default black) needed to be extended to all the tooltip orientations. This was accomplished by copy/pasting the relevant bootstrap styles into the kbase stylesheet, removing extraneous styles, and overriding just the necessary colors. This is rather order-dependent (kbase after bootstrap), so I also added a specializer ancestor class (notebook_app on the body seems to be the most appropriate one).
A note -- I was confused at first because bootstrap is not made available through the stock stylesheets, rather a generically named "style.css" within jupiter. This does raise the issue of what might be different about it, and being able to keep this stylesheet, the bootstrap javascript which is loaded from a bower-installed package, and kbase-ui in sync.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,71 @@ kbase-extension/kbase_templates/old/
 kbase-extension/static/components/
 node_modules/
 install.log
+/nbproject/private/
+/src/build/
+/venvs/
+.projectKnowledge
+/nbproject/
+static/dist
+
+# From https://github.com/github/gitignore/blob/master/Python.gitignore
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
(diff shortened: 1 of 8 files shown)


eapearson:
The last two commits are related to fixing the dropdown on the method parameter widget. The problem was (probably) that the select2-bootstrap styles were for bootstrap2, and since jupyter builds in bootstrap 3.3.5, we needed to swap in select2-bootstrap-css, which is a fork of the select2 styles specifically for bootstrap 3. The dropdown style is just slightly different, but bootstrap 3 compliant.
There was a spacing issue with the required/accepted icon to the right of the dropdown/select - this was due to a style that went missing the purpose of which was to shove the hint text over a bit to make space for the icon, which itself was squished out of its column by the select control.

briehl:
Great!

The 'style.css' ships with Jupyter - it's a compilation of Bootstrap 3's css (unmodified, I think, though I'm not certain) and other styles that Jupyter elements use.

Just so we know what's being overridden and where, it's probably best to keep Jupyter style changes separate - in kbase-extension/static/custom/custom.css, these'll get loaded in the right order.

While I'm commenting, the custom.js file in that same location is used to override Jupyter-specific JS code. It's currently being used to adjust a few different behaviors, text things, and other KBase branding problems.

PR #431, opened 2015-11-24
Fixed issue with narrative names not always appearing.

briehl:
Finally tracked this down. It was partly my mistake, partly some missing docs.


briehl:
Gonna go ahead and merge this if there are no complaints.

PR #432, opened 2015-11-24
Jupyter add trees

briehl:
Adding back missing tree rendering code. Oops.


briehl:
Also gonna merge this if there's no complaints - other fixes crept in here as well, including a bugfix that can happen when a shared user tries to rename the narrative, and a little docs update.

PR #433, opened 2015-11-24
Improving cell mechanics and controls, normalize cell layout, fix eternal scrollbar

eapearson:
A set of changes mostly from the last two days, relating to the changes we reviewed yesterday - fixing layout issues in the narrative main cell container and cells themselves, exposing the toolbar permanently, moving the collapse control and title up there. This solves the problem of "shifting cells", in which cells shift down when moused over because the toolbar is shown. Just from my working with rapidly adding and removing cells, it is quite effective for working with notebooks. Not finished, of course.
Also included is the work from this morning, which was figuring out the problem with the narrative windows permanently displaying a scrollbar with a good squidgen of free play. It ends up that the problem was due to the jupiter method for resizing the main area -- by calculating the window height, header height and subtracting. There were some problems -- the header id was not set on the page, and the jQuery method used by jupiter was not returning the outer height of the header, but the inner. Adding the header id on the nav-bar header, setting some styles to override the jupyter header styles and restore some bootstrap, and an override of that method in custom.js to get the correct height ... fixed the problem.


briehl:
Great!

PR #434, opened 2015-11-24
Few ui fixes

eapearson:
The most useful change is to disable the keyboard globally for a narrative. I don't know the best place for this, but I just set it in the Narrative constructor. To ensure that it is not enabled, i disabled the code in the enabling method. This may not be the best place or approach, but I was stuck with an immediate need -- I could not log in because the login dialog was having its keyboard stolen by jupyter. I had also noticed a largely unreproducible (because I haven't tried very hard) bug in input cells in which the input/dropdown/search control (select2) would lose keyboard to jupyter and the whole cell or narrative would get screwed up.

Smaller fixes:
I had wiped out part of a previous change when merging in from upstream, restored that (scrollbar issue.)

The border around the text cell render area (the main display area) was shifting 1px when the cell was selected, put in place extra css to ensure that the border is retained (and transparent) when unselected. There is generally so much css fighting against each other that sometimes it takes an unfortunate amount of effort for simple styling!


briehl:
I tried to go through and disable the keyboard shortcut thing whenever any text box was active, but there are probably too many side cases that are easy to miss. We also don't really have anything designed around the edit vs. non-edit modes, so it'll just be more confusing than anything. Jupyter users might miss those shortcuts, but we can address that later, once we have some use cases to sort out, I think.

And yes, there's lots of little css to fight with!

PR #436, opened 2015-11-30
Changed proxy config to force DNS lookups - should occur every 5 minu…

danielolson5:
…tes with this mod.

### DIFF
--- a/src/nginx/default
+++ b/src/nginx/default
@@ -137,7 +137,8 @@ server {
 	listen 127.0.0.1:65001;
 
 	location / {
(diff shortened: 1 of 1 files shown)


kkellerlbl:
Any objections to merging?

scanon:
I haven't used this form before but I trust Dan that it works.

kkellerlbl:
The actual config file used is actually in deploy_tools, I think; the config was overwritten earlier this morning.  Given the issues we saw in production today I wonder if this isn't the right solution anyway.  (See KBASE-3122.)

PR #438, opened 2015-12-10
Job panel fixes

briehl:
This PR addresses a few fixes
  1. The dev-mode tree page now shows the actual narrative names as well as narrative object ids and creators.
  2. Creates a semi-general Util class (includable with `define(['Util'], function(Util){})`) that has only a few pieces of functionality for now. Other bits of functionality that has been copy/pasted around various widgets by various people will eventually wind up there, or maybe in some suite of utility modules.
    * timestamp management functions
    * uuid generator
  3. Creates a bit more interactive Bootstrap Dialog module with getters and setters for each main element (title, body, list of buttons for the footer, show and hide functions). These just lightly wrap the Bootstrap dialog component, and let Bootstrap do most of the work.
  4. Refactors the jobs panel in a couple ways
    * Uses the BootstrapDialog module - this fixes the problem of the dialog losing its functionality after being used to delete jobs once or twice, and makes the management code a bit more tidy.
    * Uses the Util module to deal with timestamp manipulation. That could probably be done a bit better, and will get cleaned up as things go on.
  5. Includes test specs for the Util and BootstrapDialog modules.


briehl:
Also adds in fixes to the app/method cell re-run feature, and fixes cell state saving on manual save.

briehl:
Migrated ENIGMA widgets and their Plot.ly dependency to the Jupyter branch. If tests pass here, it should be ready to merge...

briehl:
Merging this, since we're still banging on the update...

PR #439, opened 2015-12-10
ENIGMA-KBase integration widgets and method inputs to work with 3 data types: GrowthMarix, WellSampleMatrix, and ChromatographyMarix

psnovichkov:
it includes plotly library used to implement widgets


briehl:
Looks good. This is all pretty self-contained, and will be a good exercise for seeing how well Plot.ly works with the rest of the Narrative.

PR #440, opened 2015-12-11
Put in 're-run cell' button

briehl:
This PR fixes JIRA issue KBASE-1360.
The re-run button had wandered off with recent changes. This replaces it with a few other state changes.

  1. When a method/app cell finishes running, the Cancel button is replaced by a 'Edit and Re-Run' button. Clicking this unlocks the input fields and re-enables running the cell.
  2. Fixes other state management problems - lingering run icons and lingering error state borders.
  3. In a method cell, the generated output is placed in a new cell. But in an app cell, the generated outputs are placed directly below each method. When re-running an app cell, the previously generated output viewers are removed. The data objects stay behind, however, and can be viewed by clicking on their names or dragging and dropping.
  4. After running, if next steps are available, they are only shown once. Previously, running a method/app cell multiple times might add the next steps block multiple times.


aparkin:
Yay!!!!

briehl:
Any other comments? Otherwise, I'd like to merge this and see if it behaves on CI.

PR #441, opened 2015-12-17
Update catalog client to latest api

msneddon:
Adds a couple useful methods, most importantly to list running builds.


briehl:
Great!

PR #442, opened 2015-12-17
Add ui-common to dependencies to trigger a CI build

briehl:
@kkellerlbl - this should trigger a CI build of ui-common when something changes there, right?


kkellerlbl:
I believe so, yes.  I'll merge when the tests finish (I know, the change shouldn't impact tests).

PR #444, opened 2015-12-18
Changes/additions to support the SDK

msneddon:
This PR improves the register SDK module widget, updates the JS catalog and NJS wrapper clients, adds a method to show the live logs of running (or recently completed) SDK module jobs, and adds a simple widget for the new Report data type.


briehl:
Great! Once the check finishes, let's get this on CI so we can test it there and move on to next.

PR #446, opened 2015-12-21
Dockerized Narrative + Jupyter

briehl:
This PR gets everything mostly running when building the Narrative in a Docker container with Jupyter 4.0.6 as the backend. It seems to run as usual.

Still todo:  
  1. ~~There's a bug where it needs the tree page to be accessed first to set up auth credentials internally. It should check for the cookie every time, so we can link straight into narratives.~~
  2. ~~The URLs are different. The previous version had hacked up the Tornado server to provide access to different url routes... we're not doing any more monkeypatching, so we'll have to approach this some other way - either through some form of proxying or something else.~~
  3. ~~Loading Narratives with the old format (non-updated WS object) doesn't show a title anymore (thought that was fixed...).~~
  4. The endlessly spinning data list at startup is much more prevalent, which might make it easier to diagnose, but is still a pain. We'll have to work on overhauling that widget much sooner than anticipated. For now, a refresh seems to fix it.


briehl:
Also noting here the significant changes:

* No longer inside a virtualenv in a docker container. No more environment recursion.
* Added /narrative to the base url in both the Docker form and local deployment. Local deployment doesn't require a virtualenv, but it's really a good idea to use one.
* Rewrote Dockerfile setup. Now, there are three layers expected (@kkellerlbl @scanon should be aware). The build_narrative_container script takes these into account, but we should probably put the versions somewhere, and maybe even store these in the kbase dockerhub.
  1. kbase/narrprereq = the old narrbase - this contains all base requirements for that image, including the requirements file from narrative/src. Still takes ~25-30 min to build.
  2. kbase/narrbase = a middle layer that builds in Jupyter 4.0.6 and IPywidgets 4.1.1. This builds pretty quick, so when those get updated it should be an easier process.
  3. kbase/narrative = As usual, but also includes data_api#develop. I'd like to include the data_api dependencies in at the narrprereq level, though, to get it to build faster, but that hasn't happened yet.

briehl:
Fixed problem number 1 - The little loading page that pops up while setting up a Docker session now calls the https://narrative-XXX.kbase.us/narrative/login to check if it's up instead if calling your narrative page directly. If so, then it redirects to the narrative.

I think this'll wind up having two effects - first, it fixes that auth problem (if in a little bit of a roundabout way - another solution below), and second, it fixes the problem where you might wind up seeing that page fail, but the reason for the failure is you don't have permission to see that narrative, NOT that the NI itself is having problems. Now it should properly redirect to the "You don't have permission!" page.

Another solution would be to invoke that login page from the Lua code when a new session gets activated. It's just a simple HTTP request that forwards the auth token. That might make more sense for the architecture, but it means we will need to maintain that behavior if we ever move to using Jupyterhub to set up containers.

Either way, the current solution doesn't live in this repo - it's a ui-common fix.

briehl:
Fixed the name-not-showing problem, and a new one where the copy button didn't work in a read only narrative.

briehl:
There's still some strageness with the tests failing - it can't seem to find a requirejs dependency. Other than that, most everything here should be up to date.

briehl:
Huh, that fix to the auth problem didn't seem to stick overnight. Being disconnected didn't trigger a narrative shutdown, and loading up a narrative gave the same 403 error. Maybe injecting through the Lua stack is the right solution - I'll add that after the demos.

briehl:
This should be more stable now. I added an explicit server handler for deal with the URLs we're used to for narratives. It's a little clunky right now and includes behavior that should be migrated somewhere more general, but it works. This should solve the problem of disappearing auth on a page refresh, and the redundant '/notebooks/' in the url.

The front end tests are still broken somehow, so those should be fixed before accepting this.

PR #447, opened 2016-01-04
Fixing front end testing harness.

briehl:
This will (hopefully) fix the issues with front end tests not finding the right files.


briehl:
Looks like it worked! Silly bug.

PR #448, opened 2016-01-05
Migrating old code.

briehl:
Doing this now to tear off the band-aid.

This PR migrates the front end code from that obnoxiously long path: `/src/notebook/ipython_profiles/profile_narrative/...`
to
`/kbase-extension`

Under that directory, things have been modified ever so slightly.  
  * Jupyter Config files are all flat at the top of that directory
  * Jinja2 templates (overrides to the Jupyter notebook templates) are in `/kbase-extension/kbase_templates`
  * JS code is under `/kbase-extension/static/`
    * the `/kbase-extension/ext_components/` dir is the spot where Bower drops in external components (there are a few that aren't in Bower, but are also included here.
    * the `custom/` dir is for specific custom extensions to Jupyter, e.g. overriding their code for nefarious KBase purposes
    * the `/kbase-extension/static/kbase/` dir tree should be the same as the old `/src/notebook/ipython_profiles/profile_narrative/kbase_templates/static/kbase` dir.

For reference, if you want any files to be available on the web server they need to be under `/kbase-extension/static`. They'll be available as `http://localhost:port/narrative/static/....`.


briehl:
Woohoo! Tests pass again! This should be ready to merge.

PR #449, opened 2016-01-07
improve markdown cell behavior

eapearson:
double click expands closed cells - before it was putting the cell into edit mode without actually toggling the visibility, which was weird. this is accomplished in the double click event on the cell bind_events.
toggling no longer interferes with edit / render state - on the other hand toggling put the edit/render cell areas into unexpected states. this fixes that by avoiding jQuery hide/show, instead adding display="none" to hide, and simply removing it (and thus returning to unmodified state) to show. 
Will be nice to get this working consistently across cell types, but this was targeted to have double click work more or less for markdown cells, in which it already does something. I'm not sure double click does anything interesting elsewhere.

### DIFF
--- a/kbase-extension/static/custom/custom.js
+++ b/kbase-extension/static/custom/custom.js
@@ -168,13 +168,16 @@ define(['jquery',
             var $cellNode = $(this.element);
             switch (this.getCellState('toggleState', 'unknown')) {
                 case 'closed':
-                    $cellNode.find('.inner_cell > div:nth-child(3)').hide();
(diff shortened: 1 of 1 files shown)


briehl:
Double click will probably have similar behavior in code cells, which is probably fine. Let's merge this and see how behavior changes there.

eapearson:
The way the double click behavior propagates works for code cells, but not for app/method cells, although I don't know how it might interact with code cells in actual usage. It really just triggers the toggle for a closed cell (if it can detect that it was closed, but that is another issue...). Rather than expand this to cover those cases, or to solve the problem of double clicking open cells, I thought it better to propose this little fix without further ado.

PR #452, opened 2016-01-08
Fixed undeclared variable issue in FileInput parameter widget, modifi…

briehl:
…ed front end test runner.


briehl:
Skipping the tests since this is a really really minor fix.

PR #455, opened 2016-01-13
Changes related to SDK job progress panel on the bottom of method input cell.

rsutormin:


briehl:
What does this do? It looks like a bunch of code was copied from the jobs panel to the method panel.

briehl:
Okay, merging so we can test. Adding a note that Roman and I will work together later to break out job handling tools into a module. This'll probably be useful for uploading and downloading, too.

PR #461, opened 2016-01-14
Reverting changes in job panel made for method input widget but not required anymore.

rsutormin:


briehl:
Great. Thanks, Roman.

PR #462, opened 2016-01-14
File upload input parameter: cleanup of shock nodes switched off, debug messages added.

rsutormin:
See https://atlassian.kbase.us/browse/KBASE-3365 and https://atlassian.kbase.us/browse/KBASE-3267 for details.


briehl:
We should probably limit the console.log stuff in production, but this will be useful for testing.

PR #466, opened 2016-01-15
Add kbase-ui to DEPENDENCIES

briehl:


briehl:
Being a terrible person and skipping the travis test - this won't break the test, but is necessary to build kbase-ui in CI.

PR #468, opened 2016-01-20
narrative_version file and uploader fixes

briehl:
* Puts narrative_version file in the right place (might get overwritten depending on the build order - KBase-ui should come first, then narrative... needs testing in CI)
  * Fixed uploaders for Chromatography, Well Sample, and Growth data.
  * Added a number of console.log checks for the progress of data panel issues that can be hard to trap at startup.
  * A few updates and code maintenance.


briehl:
Tests pass, merging this to get it up on CI so @psnovichkov and @aekazakov can test their uploaders. New documentation on the uploader config will be forthcoming...

PR #470, opened 2016-01-22
Bugfixes

briehl:
At the leads call this morning, we decided to put off any deployment of new code on CI of kbase-ui and narrative until tomorrow. So this PR is just a work in progress until then.

* Suppressed "Notebook converted" dialog (https://atlassian.kbase.us/browse/KBASE-3396)
* fixed tests
* removed unused jquery event "workspaceUpdated.Narrative".


briehl:
This should now:

* hide the "Notebook converted" dialog. When loading a narrative on the back end, it removes the orig_version key from the model
* run backend tests properly - something changed so that an unauthenticated user looking up permissions no longer throws a ServerError
* use Bluebird (the Promises/A+ spec-compliant promise library) for direct AJAX calls (client calls will come later)
* no longer reference the deprecated IPython object, but the new Jupyter object
* no longer crash when loading on a slow connection due to dependency issues
* load data types that have no associated viewer
* no longer reference 'functional-site' in urls (needs some more rigorous testing in place with other kbase-ui code)
* normalize most modules to the camelCase naming convention

briehl:
Should also no longer select multiple cells on startup.

briehl:
I'd like to get this in, since it fixes a crash bug on CI.

PR #471, opened 2016-01-25
Fixes for file upload in narrative

olsonanl:
This PR includes the following changes to fix the file uploads in the narrative.

Updated Shock client code in kbase-api-client.js. Since the minification procedure appears to not be automated at this point I have copied in the kbase-api-client.js from ui-common (commit hash in that commit) and applied the shock fixes. This fix appears to alleviate the random failures that the shock widget showed before by not loading the chunks explicitly into memory, instead passing a slice to the underlying transfer object to transfer. 

The updated Shock client code current exists only in this file. We need to define the canonical home of that file, move it there, and make a clean definition of how the rolled up Javascript API bundle is created.

Changed file upload widget. This widget now delays the actual transfer of files until the Import button is selected, and chains the creation of the transform request to be performed after the file transfers have completed. Multiple file transfers will execute in parallel. 

Removes registration of files with UJS. Because files are not transferred until a transform operation is initiated we will not have a proliferation of unwanted files in Shock. I do need to add cleanup of shock nodes for cancelled import operations.


briehl:
Tagging @rsutormin and @mlhenderson . Mind going over these updates?

rsutormin:
It looks like these changes are done with relation to data import panel instead of supporting all functionality in "kbaseNarrativeMethodCell.getParameters" method (https://github.com/kbase/narrative/blob/jupyter/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeMethodCell.js#L260-L264). It's absolutely critical to to that because we need to be able to support file-upload parameters in normal narrative method cells (not just in import tab). And this change seems to break this support. Here is an example of dynamically registered (with help of SDK) method doing upload like I explained: https://github.com/msneddon/pe_read_uploader_test/blob/master/ui/narrative/methods/upload_pe_reads/spec.json#L21 . You should be able to find it in Narrative Method Panel as "Upload PE Read Library" ("PEReadUploaderTest" module). As you can see it uses file upload parameter in normal method cell (outside of import panel).

olsonanl:
I get the idea of what you are saying but this doesn’t give me a clear path forward. If we want a different functionality for import than for narrative based uploads we should fork the widget since the upload-on-import model is much cleaner for the import case.

What is the intended call sequence for widget in narrative method? None of this is at all clear.

On Jan 25, 2016, at 3:13 PM, Roman Sutormin <notifications@github.com<mailto:notifications@github.com>> wrote:


It looks like these changes are done with relation to data import panel instead of supporting all functionality in "kbaseNarrativeMethodCell.getParameters" method (https://github.com/kbase/narrative/blob/jupyter/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeMethodCell.js#L260-L264). It's absolutely critical to to that because we need to be able to support file-upload parameters in normal narrative method cells (not just in import tab). And this change seems to break this support. Here is an example of dynamically registered (with help of SDK) method doing upload like I explained: https://github.com/msneddon/pe_read_uploader_test/blob/master/ui/narrative/methods/upload_pe_reads/ spec.jso n#L21<https://github.com/msneddon/pe_read_uploader_test/blob/master/ui/narrative/methods/upload_pe_reads/spec.json#L21> . You should be able to find it in Narrative Method Panel as "Upload PE Read Library" ("PEReadUploaderTest" module). As you can see it uses file upload parameter in normal method cell (outside of import panel).

msneddon:
To provide a little bit of context, one of the ideas that has been floating around this last week is to use the SDK infrastructure to implement uploaders and downloaders (which could in theory be run as a Narrative method directly, or used to populate the import panel- that needs some UX thought).

The advantages of migrating transform methods to the SDK are that dependencies would be nicely isolated for each uploader, fixes could be released for each uploader independently, and changes wouldn't require AWE worker redeploys.  We would also get the built-in SDK logging, error handling and metrics for free.  The "PE uploader" was a first test to show just that this was feasible without changing anything else.  No other non-uploader methods yet use the "File" input type.

With that in mind, it would be good to keep compatibility with both narrative methods and the import panel for "File" parameter types if at all possible.  If we switch to a mechanism that starts the file transfer only after clicking on import, then this just means we should update the Run button action for narrative methods so that the behavior is consistent.

I am wondering though- by switching the behavior, does that fix a bug or improve stability, or is it just a UX enhancement?  If it is an important performance fix, we should just make sure the change is also added to the "Run" button logic soon so we can experiment with uploaders as SDK methods.  But if it is primarily a UX feature, we should probably split this into two PRs so we can get the shock clients interaction bugs fixed and deployed now, and give more time to figure out if the change makes sense and how we should update the logic that runs methods in general.

danielolson5:
This is addressing both UX and stability.  Since this is fixing issues in production I think we should merge this. i agree with @msneddon, the Run button logic should change for uploaders.  I think we should handle that seperately as uploader as sdk methods are a separate feature.

scanon:
@msneddon,

Please merge it.  We can make fixes to make it work with the SDK.  Upload from the SDK is not a top priority at this point, so it is okay if this breaks temporarily.

msneddon:
Sounds good.  Let's get this merged in.  Unfortunately it looks like there's now a merge conflict, hopefully minor.  @olsonanl can you please help resolve the conflict?

I also noticed that the submodule tags changed.  Was that intentional?  Just double checking since it is easy to accidentally commit if your branch is a little behind what's upstream.

Thanks.

olsonanl:
I didn't intentionally change submodule tags; I'm guessing was a result of the narrative build process.

olsonanl:
I created PR 476 with the conflict resolved. I don't know how to force a reset of the ui-common versions; if you can tell me how I will do that.

briehl:
You can just make sure you're in sync with the upstream branch and run `git submodule update`.

Also, as long as your changes keep in the same branch, the PR will automatically update, you don't need to submit another one.

olsonanl:
I did the submodule update and still have this in a diff against upstream/develop. I don't know if that is expected or not.

diff --git a/modules/revised-widgets b/modules/revised-widgets
index 43cf81b..0065399 160000
--- a/modules/revised-widgets
+++ b/modules/revised-widgets
@@ -1 +1 @@
-Subproject commit 43cf81b46d4a387ca3bcedfe14f71e589eac3e3f
+Subproject commit 00653997142afd190f2338a3354ab9d0a13d453b-dirty
diff --git a/modules/ui-common b/modules/ui-common
index 6fcc4e5..093da42 160000
--- a/modules/ui-common
+++ b/modules/ui-common
@@ -1 +1 @@
-Subproject commit 6fcc4e59eea90aed869f69ae65606e775bd7f6ba
+Subproject commit 093da42f6359c5d5f830f547f8824bd4fe0925d0-dirty

briehl:
That looks out of date. Can you do a pull from upstream/develop into your branch? Here's the process I usually do:

```
git pull upstream develop
git submodule update
git commit                         #that should revert your change to the modules tags
git push
```

olsonanl:
Hm (the change in kbase-client-api.js is more cleanup I don’t yet want to commit) -

mcswl120:narrative olson$ git pull upstream develop
From https://github.com/kbase/narrative
 * branch            develop    -> FETCH_HEAD
Already up-to-date.
mcswl120:narrative olson$ git submodule update
mcswl120:narrative olson$ git commit
On branch develop-2016-0125-1
Your branch is up-to-date with 'origin/develop-2016-0125-1'.
Changes not staged for commit:
modified:   kbase-extension/static/kbase/js/api/kbase-client-api.js
modified:   modules/revised-widgets (modified content, untracked content)
modified:   modules/ui-common (modified content)

Untracked files:
kbase-extension.tgz
kbase-extension/static/kbase/js/api/fixed.js
kbase-extension/static/kbase/js/widgets/function_input/parameter_input/kbaseNarrativeParameterFileInput.js.original
kbase-extension/static/kbase/js/widgets/function_input/parameter_input/old
venv/

no changes added to commit

mcswl120:narrative olson$ git add  modules/revised-widgets modules/ui-common
mcswl120:narrative olson$ git commit
On branch develop-2016-0125-1
Your branch is up-to-date with 'origin/develop-2016-0125-1'.
Changes not staged for commit:
modified:   kbase-extension/static/kbase/js/api/kbase-client-api.js
modified:   modules/revised-widgets (modified content, untracked content)
modified:   modules/ui-common (modified content)

Untracked files:
kbase-extension.tgz
kbase-extension/static/kbase/js/api/fixed.js
kbase-extension/static/kbase/js/widgets/function_input/parameter_input/kbaseNarrativeParameterFileInput.js.original
kbase-extension/static/kbase/js/widgets/function_input/parameter_input/old
venv/


On Jan 27, 2016, at 11:59 AM, Bill Riehl <notifications@github.com<mailto:notifications@github.com>> wrote:


That looks out of date. Can you do a pull from upstream/develop into your branch? Here's the process I usually do:

git pull upstream develop
git submodule update
git commit                         #that should revert your change to the modules tags
git push

briehl:
I'll go ahead and merge this, and I can fix the tags outside. Not a big deal.

PR #472, opened 2016-01-26
Reformat Narrative metadata

briehl:
**DO NOT MERGE** this is pending some dashboard support changes

This is related to a couple JIRA tickets. The main one for tracking this task is here:
https://atlassian.kbase.us/browse/KBASE-3388

And the issue it exposed is here:
https://atlassian.kbase.us/browse/WOR-226

The short version is that Workspace object metadata has finite size, and each key value pair (KVP) has a smaller finite size. For convenience, we stored the count of methods and apps used in each narrative as a dictionary in the metadata, represented as JSON in the structure. With the new SDK going on, the number of distinct apps and methods has grown, and we found that adding many unique methods and apps (>40) would overflow the metadata KVP size limit, and cause the Narrative to fail to save (it actually crashed pretty hard, and that's been solved in WOR-226).

The right solution here was to flatten that dictionary of methods and apps into separate keys in the metadata. See KBASE-3388 for implementation details. A test was then added to ensure that the total metadata size stays below the maximum allowed, and collapses app/method keys into smaller groups if it hits the limit.

No test was put in for individual size, because the max size of method and app ids is limited by their location in the filesystem to 255 bytes, well below the limit we have.


briehl:
A side note - this means the dashboard area that shows which apps and methods are being used will need a rewrite. That should probably be done in concert with merging this change so CI isn't too disrupted. It doesn't crash the dashboard page, but it does make an empty box where the cell counts should go.

src/biokbase/narrative/narrativeio.py
```diff
@@ -32,6 +33,9 @@
 
 obj_ref_regex = re.compile('^(?P<wsid>\d+)\/(?P<objid>\d+)(\/(?P<ver>\d+))?$')
 
+MAX_METADATA_STRING_LEN = 900
```
MrCreosote: FYI, the limit is for bytes, not characters. If you have strings with non-ASCII characters you could pass the check below but exceed the workspace limit.

briehl:
Ok, this is ready to merge - the front end changes have been made on CI, so now we need to integrate and test.

PR #474, opened 2016-01-26
viz for SDK method Evaluate Flux Balance Analysis Against Gene Expression

nconrad:
Looks like this so far:

screen shot 2016-01-26 at 11 40 10 am


briehl:
Looks good!

PR #475, opened 2016-01-26
Fix cell selection issues

briehl:
Clicking buttons on unselected cells (specifically the run buttons or the delete button) would often cause behavior to be based on the other selected cell. For example, a different cell would get deleted, or output cells would appear in the wrong place.

This fixes that by properly binding each KBase widget to the right cell. Instead of just assuming it's selected and calling `Jupyter.notebook.get_selected_cell()`, call  
`Jupyter.narrative.getCellByKbaseId(id)` or  
`Jupyter.narrative.getCellIndexByKbaseId(id)`.

The ids are in the $elem bound to the KBase widget.


briehl:
Merging this now to get it tested on CI.

PR #477, opened 2016-01-27
Restores submodule tags to be up to date.

briehl:


briehl:
That should do it. Once it passes tests, I'll merge.

PR #478, opened 2016-01-28
Data List Adjustment

briehl:
Made a number of changes to the Data and Narratives panels.

Narratives panel:  
* Started using Bluebird Promises (actual A+ spec compliant library) for various workspace calls
* **TEMPORARILY** added new workspace client direct to repo - this needs to find a home in kbase-service-clients-js
* Now using get_permissions_mass to look up all narrative permissions at once (well, up to 1K at a time)
* Some general cleanup

Data panel:
* Rewrote majority of loading logic - some rendering was mixed up with data loading which could cause problems.
* Rewrote logic to use Bluebird promises
* Rebuilt spinner panel to be more informative (includes little bits of text at each stage)
* Stopped some cases where the refresh could happen more than once on startup.

Data slideout:
* Made all tabs (except import) lazy-loading. Should cut down pretty dramatically on startup WS calls. Each tab should only do its first load when it first appears.
* Removed a bunch of duplicate code from a few widgets.

Still to do:
* Test pretty thoroughly on CI. These changes are big, and though I've gone through several narratives and several test users with them, the big test will be seeing what happens to others.
* Fix styling. The loading segments tend to make the panels bounce around in size. Needs to be ironed out.


briehl:
Gonna go ahead and merge this to update CI.

PR #479, opened 2016-01-28
Updated list_objects calls to use min/maxObjectID

briehl:
…  instead of skip/limit params.

Also includes a little cleanup to the Public slideout tab to use Promises.

Update - this is going to take a little refactor of the My Data and Shared With Me panels. Right now, there's no guarantee that they get everything for large (>10K elements) workspaces, or for users who have lots of workspaces.

### DIFF
--- a/kbase-extension/static/kbase/css/kbaseNarrative.css
+++ b/kbase-extension/static/kbase/css/kbaseNarrative.css
@@ -476,7 +476,7 @@ p#site-title {
 
 .kb-overlay-dimmer {
     position: fixed;
-    z-index: 1020;
+    z-index: 10;
     right: 0;
     bottom: 0;
     opacity: 0.5;
--- a/kbase-extension/static/kbase/js/api/NarrativeManager.js
+++ b/kbase-extension/static/kbase/js/api/NarrativeManager.js
@@ -235,12 +235,9 @@ define(['jquery'], function( $ ) {
                     
                     var narrativeObject = {
                         nbformat_minor: 0,
-                        worksheets: [ {
-                            cells: cell_data,
-                            metadata: {}
-                        }],
+                        cells: cell_data,
                         metadata: metadata,
-                        nbformat:3
+                        nbformat: 4
                     };
                     
                     // setup external string to string metadata for the WS object
--- a/kbase-extension/static/kbase/js/kbaseNarrative.js
+++ b/kbase-extension/static/kbase/js/kbaseNarrative.js
@@ -313,7 +313,7 @@ function($,
         $('#notebook').append($versionModal);
     };
 
-    Narrative.prototype.saveFailed = function() {
+    Narrative.prototype.saveFailed = function(event, data) {
         Jupyter.save_widget.set_save_status('Narrative save failed!');
         console.log(event);
         console.log(data);
--- a/kbase-extension/static/kbase/js/util/bootstrapDialog.js
+++ b/kbase-extension/static/kbase/js/util/bootstrapDialog.js
@@ -6,7 +6,8 @@ function ($) {
 
     /**
      * options:
-     * { title: string,
+     * { 
+         title: string,
      *   body: jquery node
      *   buttons: array of jquery nodes
      *   closeButton: boolean, default false
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeAppCell.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeAppCell.js
@@ -15,6 +15,7 @@
   require(['jquery',
            'narrativeConfig',
            'util/string',
+           'util/bootstrapDialog',
            'kbwidget',
            'kbaseAuthenticatedWidget',
            'kbaseAccordion',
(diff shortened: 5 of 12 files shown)


briehl:
Rewrote the widget that creates the 'My Data' and 'Shared with Me' panels. It should now properly fetch the user's available data using the new mechanism. We should, at some point, rethink how that panel's expected to work - maybe start it empty, and force the user to search for stuff they own / have shared? Though I'd prefer to have it pre-populated with their data.

cshenry:
Speaking as someone who has 300K objects now… I would love if it started empty and forced you to do a search. Otherwise, it just takes forever to load for me and locks down everything else while it’s working on it.

-- 
Chris Henry
Sent with Airmail

On January 28, 2016 at 11:34:43 PM, Bill Riehl (notifications@github.com) wrote:

Rewrote the widget that creates the 'My Data' and 'Shared with Me' panels. It should now properly fetch the user's available data using the new mechanism. We should, at some point, rethink how that panel's expected to work - maybe start it empty, and force the user to search for stuff they own / have shared? Though I'd prefer to have it pre-populated with their data.

briehl:
Yeah, I was just thinking about that on the way home. There are a few power users out there that would benefit from it.

One (maybe easy?) change might be this - if you have > 10K objects (more than can be fetched in a single request), just grab the first chunk, and have a message saying that you're a super duper user and should use the search bar.

Then it's a matter of wiring the search bar to lookup things from the workspace instead of filtering what's already been fetched. I *think* that's part of this 0.4.0 update.

MrCreosote:
>> Then it's a matter of wiring the search bar to lookup things from the workspace instead of filtering what's already been fetched. I think that's part of this 0.4.0 update.

Well, there's this:
https://github.com/kbase/workspace_deluxe/pull/33
https://github.com/kbase/workspace_deluxe/blob/dev-candidate/workspace.spec#L1291

Which may or may not be what you're looking for. The `list_objects` function is also sort of a search function - it has a number ways you can slice your data up.
https://github.com/kbase/workspace_deluxe/blob/dev-candidate/workspace.spec#L1043

briehl:
Okay, with this last one, a few heavy changes - something to iterate on, at least.
1. Limit total data fetched to 100,000 objects (by max object id of each ws, so real value might be smaller). Added a prompt to tell the user to filter by type or narrative if they're at that limit. Search still only filters currently rendered objects, though, I'm not sure if just searching by prefix will be as useful as a general string search (and I think that might make Mongo cry).
2. Made all list_objects requests done in serial. Bluebird's Promise.reduce is awesome.
3. Added a progressbar. Could be replaced by the old spinner if there's only one request, but I think this'll be useful. Min value is 5%, so it always looks like it's doing something.

briehl:
@cshenry - you're kind of the power user here, would you mind checking this out on narrative-dev and see if it works for your case? I can also just put this on CI and see how it goes. It has fixes that need to get merged there anyway.

cshenry:
The problem is I’m only a power user in the production environment… I don’t have a ridiculous number of objects in any of the other workspaces.

-- 
Chris Henry
Sent with Airmail

On January 29, 2016 at 10:49:15 PM, Bill Riehl (notifications@github.com) wrote:

@cshenry - you're kind of the power user here, would you mind checking this out on narrative-dev and see if it works for your case? I can also just put this on CI and see how it goes. It has fixes that need to get merged there anyway.

briehl:
Of course. Didn't think of that. Nevermind, then. It's also not having the right behavior now, and needs a little adjustment.

Thanks anyway.

scanon:
I w

MrCreosote:
> 100,000 objects

Did you mean 10K? That's what we talked about yesterday, just wondering if this was a typo or you changed the limit.

> I'm not sure if just searching by prefix will be as useful as a general string search (and I think that might make Mongo cry)

If by general string search you mean search based on any given substring of the object name, as far as I know there's no way to do that in MongoDB without performing a regex on every entry in the index, which would very likely be slow. Prefixes can do a range query on the index which is a lot faster, obviously.

If we really really need it I can do some timings and see how bad it is (or possibly isn't).

Also, I think you might be underestimating the utility of prefix based searches - that's what most search engines (and a lot of other applications) do when they provide search suggestions based on what's been typed so far.

> The problem is I’m only a power user in the production environment… I don’t have a ridiculous number of objects in any of the other workspaces.

kbasetest has a zillion objects on ci.

scanon:
Sorry about that.

briehl:
I set it to 100,000 for testing, to see how performance is, especially against the kbasetest account on CI. It's not bad, but we can bring it down in production. There are still a few kinks that need to be worked out with how it handles the type-specific searching, though - that [WIP] tag is still up there :)

The switch to using just prefix search will be a UX change, though, since that's not what's done right now. I personally find it more useful to do a general string search, but that might just be me - I'm kinda prototyping here without user testing. I'll put in prefix searching and see how it fares on CI.

MrCreosote:
I suppose you can keep the substring search for loaded objects and use prefix searching for loading more objects.

briehl:
Made a few more changes. The filters should work more or less correctly now, and it'll fetch up to the configured maximum objects, then cancel all remaining unfired requests.

The only thing that's a sticky point is the text filtering. It'll only filter what's loaded into the panel. But if the user has > 100,000 objects of the same type in a single Narrative, that'll be the only problem. As that will only be a problem for kbasetest on CI (and maybe for Chris in production), I vote for putting off any more changes until it becomes an issue.

Once this passes tests, I'd like to get this deployed on CI, and make further edits there.

PR #482, opened 2016-02-02
Made initialization dependent on kernel connection, some ux fixes.

briehl:
A few things:
* Some updated release notes
* Made the blinking kernel button listen to the right kernel_busy event
* Tied KBase-layer startup to kernel connection. This fixes a few out-of-sync problems.
* Added tests for kernel connectivity and some rational behavior to a few kernel commands (Jobs panel, error logging)


briehl:
Merging this to fix problems on CI.

PR #485, opened 2016-02-03
Fix read only mode

briehl:
The read only mode toggle button works again. it still remains in view only mode until page reload, even if share settings change while viewing the page.

Additionally, you can now slide the left control panel out and back. It might be worth decoupling that from read only mode and still allowing usage. Honestly nothing much would change there, except for allowing existing run buttons to be pressed, and keeping the "View-only mode" message from showing. The only work to be done there would be where and how to define the toggle button.


briehl:
Tests pass, merging.

PR #486, opened 2016-02-03
A bunch of docs updates

briehl:
Re-arranged the docs folder, tried to keep like documents together. E.g. the `install` directory has all installation and deployment docs, the `developer` directory has information relevant for developers, like the method running flow, and some nascent docs on Javascript helper widgets.


briehl:
Oops, this also fixes a problem where reloading a narrative won't always populate input widgets with the right data objects they should have access to.

PR #487, opened 2016-02-03
add model comp

mdejongh:


briehl:
Looks good. Skipping the check to get it out on CI faster. This won't affect the auto-testing anyway.

PR #492, opened 2016-02-04
Update new fba model comparison widget

briehl:
It now uses the narrativeConfig module to fetch url configs, Bluebird for Promises, and the util/string module for uuid generation.

Some docs on those here: https://github.com/kbase/narrative/blob/develop/docs/developer/widget_helper_modules.md

And Bluebird is used throughout kbase-ui - this is just an early attempt to keep caught up.

Also, it looks like a lot of changes below, but most of it is my editor rearranging whitespace a bit.


briehl:
This isn't triggering a Travis-CI test for some reason. We updated the default repo, so maybe it needs a push to start the tests up again? I'll go ahead and merge this, and see what happens.

PR #493, opened 2016-02-04
Fix for bug in kbaseNarrativeMethodCell.getNextSteps() blocking output.

rsutormin:


rsutormin:
@msneddon, this is what we discussed few hours ago.

briehl:
Yup, that should fix it. If data.next_steps isn't present, it shouldn't throw the error.

PR #495, opened 2016-02-05
Methods Panel Update

msneddon:
This updates the methods panel to add several new features and provide a minor performance improvement:

1) instead of pulling complete specs of methods to filter by type, uses the new NMS brief info object.  So much less data has to be pulled and filtering by type is faster.  Specs are pulled when a method is added or a spec is requested.
2) Custom method icons are now displayed if they are registered with the SDK method or the NMS (also pulled out that code into a util function, with a few new tests, so this isn't copied in multiple places)
3) Links in the method panel and method/app cells are updated to point to the new App Catalog
4) Support for favorites is now in the methods panel.  Your favorites will be sorted to the top of the list, ordered by the date that you added a method to your list of favorites.  Within the panel, you can also directly add/remove favorites by clicking on the star.

### DIFF
--- a/kbase-extension/static/kbase/css/kbaseNarrative.css
+++ b/kbase-extension/static/kbase/css/kbaseNarrative.css
@@ -2262,10 +2262,11 @@ div[class="tooltip-inner"] {
     font-size: 10pt;
     margin: 2px;
     margin-left: 10px;
-    cursor: pointer;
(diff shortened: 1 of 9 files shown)


msneddon:
I forgot to mention that with the decreased focus on old style Apps in favor of SDK methods, this update also moves the Apps to the end of the list.

PR #500, opened 2016-02-08
Fix problem with renaming old narratives.

briehl:
Put in a check in save_narrative for old versions. Old versions had a segment like this:
```
{
    'worksheets' : [ {'cells': [ { list of cells...} ], 'metadata' : {} ]
}
```
new ones look like this:
```
{
    'cells': [ {list of cells} ]
}
```
basically, it skips the metadata. This is all automagically dealt with by the Jupyter loading machinery... unless only the name is changed on save. Then we have to deal with it manually. This change adds a check/fix in the KBase narrative save process that handles this.


briehl:
This affects older narratives, and only renaming them, not saving.

PR #501, opened 2016-02-08
App Catalog slideout added to the narrative

msneddon:
This PR ports over the app catalog search and organization page to the Narrative.  It adds back the informational slide-out from the methods panel, where it renders the catalog browser.  Links from the browser still link out to module/method pages in kbase-ui.  Clicking on methods/apps closes the slideout and inserts them in the narrative.  Refresh buttons were removed from the catalog browser itself and now respond to refresh and toggle through dev->beta->release in the main methods panel.


kbase-extension/static/kbase/js/widgets/narrative_core/catalog/catalog_util.js
```diff
@@ -0,0 +1,108 @@
+define([
+    'bluebird',
+    'jquery'
+], function (Promise, $) {
+    'use strict';
+
+    function CatalogUtil() {
+
+        // edited from: http://stackoverflow.com/questions/3177836/how-to-format-time-since-xxx-e-g-4-minutes-ago-similar-to-stack-exchange-site
+        this.getTimeStampStr = function (objInfoTimeStamp) {
```
briehl: This is in util/timeFormat already, I think. If not, that should be updated!

kbase-extension/static/kbase/js/widgets/narrative_core/catalog/catalog_util.js
```diff
@@ -0,0 +1,108 @@
...
+                }
+                if(categories[i]=='importers') {
+                    return true;
+                }
+            }
+            return false;
+        };
+
+
+        this.initLoadingPanel= function() {
```
briehl: There's a similar loading panel in util/display that might help. If not, this could move there, too.

kbase-extension/static/kbase/js/widgets/narrative_core/catalog/kbaseCatalogBrowser.js
```diff
@@ -0,0 +1,1105 @@
...
+                        }
+                    })
+                    // [NARRATIVE_EDIT] - catch to fail
+                    .fail(function (err) {
+                        console.error('ERROR');
+                        console.error(err);
+                    });
+            },
+
+            // warning!  will not return a promise if the user is not logged in!
```
briehl: If it helps some logic flow, you can just return an empty promise (sounds kinda sad, I know...) with Promise.try(function() {});

kbase-extension/static/kbase/js/widgets/narrative_core/catalog/catalog_util.js
msneddon: Good call!  I didn't have these util methods (or at least I didn't know where they are) in kbase-ui, but I'll update this in the narrative.  I can probably get rid of this util code altogether.

msneddon: turns out I'm not even using the timestamp code in the catalog browser.  I'll just remove this.

briehl: Eventually (in the magical after-contractor's-meeting time) we should get the kbase-ui and narrative util modules in sync.

msneddon: Ok, for now I was able to remove the catalog_util dependency completely, and just use the util/display.js already in the Narrative.

kbase-extension/static/kbase/js/widgets/narrative_core/catalog/kbaseCatalogBrowser.js
msneddon: Thanks. I was wondering about how to handle that situation. I'll update that in the kbase-ui version of this.  In the Narrative version, that warning doesn't apply because right now you must be logged in.

msneddon:
Thanks for the comments @briehl!  I've managed to remove the dependency on the catalog_util file altogether, so I just removed it.  I also switched over to using the same filter categories approach as in the method panel, so at least if we have to update the categories we can do so from a single place.

Let me know if there are other issues.

briehl:
Looks good.

PR #503, opened 2016-02-09
Another pass at styling

briehl:
* Removed padding between widgets and cell boundaries.
* Put method subtitle back where it belongs.
* Got rid of automatic black borders around tables.
* Added a left-side border for selected method/output cells.

### DIFF
--- a/kbase-extension/static/custom/custom.css
+++ b/kbase-extension/static/custom/custom.css
@@ -158,12 +158,9 @@ span#notebook_name:hover {
     color: #888;
 }
 
-
 .notebook_app .btn.active {
     box-shadow: none;
 }
-
-
 
 /* Fooling around with the narrative cells */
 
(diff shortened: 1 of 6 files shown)


msneddon:
I just tried this out locally to check out the new styling updates (it keeps looking better!), and noticed that running on sdk methods gives me an error in kbaseNarrativeMethodCell.js, line 887, where it is trying to call hide() on $staticMethodInfo that is no longer defined.

briehl:
D'oh! Thanks for checking.

briehl:
I guess that brings up another thing - running tries to minimize the input area, but there's no way to restore it now, since the only minimize/maximize control is on the whole cell. This might become an issue.

briehl:
Thanks for the help @msneddon . Merging this now.

PR #504, opened 2016-02-10
Data and Methods side panels now dynamically resize

msneddon:
We should have just done this a while ago!  This turned out to be relatively simple fix.  Added a callback option to the side control panels widget that is triggered when a panel is minimized/maximized.  This is now wired to setting the height of the data and method list, so these lists now dynamically resize to fill up the free space in the side panel.

This isn't the more complicated 'responsive' fix to adjust things everytime the browser resizes or on different screen resolutions, but it's much better than what we have now.

This also tweaks a style so we never get multiple scrollbars appearing in the side panel.


briehl:
Nice! I'll have a follow on in a minute that'll just plain hide the side panel.

PR #507, opened 2016-02-10
More side panel improvements

msneddon:
The changes from my PR last night had a side effect of removing scroll bars in the narrative and jobs panel.  I realized that the cleanest way to fix this is to just finally make the heights responsive to the window height, which turned out to be fairly straightforward given the new methods for setting the height of the control panel widgets.  So that finally works now!

I also made a separate minor styling change by removing the titles to the 'Narrative' and 'Jobs' panel.  I think it looks better, but could still use some styling tweaks (for instance, the jobs list seems compressed slightly more now).  We can always revert that or work on styling when we find time later.


briehl:
Hey Mike, your changes last night cause the side panel to still have a maximum height that can go off the end of the page. So if your browser is too short, you won't be able to scroll to the bottom of whatever panel's active. Does this fix that?

briehl:
That did fix it. Looks great!
I think removing the titles to the Narrative and Jobs panels is fine, but it leaves kind of a gap. I'll see if I can fix that. But let's merge this first.

PR #523, opened 2016-02-16
widget for sequence comparison

levinas:


kbase-extension/static/kbase/js/widgets/function_output/kbaseSeqCompView.js
```diff
@@ -0,0 +1,201 @@
...
+            return this;
+        },
+
+        loggedOutCallback: function(event, auth) {
+            this.token = null;
+            this.render();
+            return this;
+        },
+
+        uuid: function() {
```
briehl: There's now a separate module, `util/string` you can use as a dependency that provides uuids. See https://github.com/kbase/narrative/blob/develop/docs/developer/widget_helper_modules.md#utilstring

kbase-extension/static/kbase/js/widgets/function_output/kbaseSeqCompView.js
```diff
@@ -0,0 +1,201 @@
...
+            	}, function(data) {
+            	    container.empty();
+            	    container.append('<p>[Error] ' + data.error.message + '</p>');
+            	});
+            };
+            ready();
+            return this;
+        },
+
+        getData: function() {
```
briehl: The need for the `getData` function died with the old card landing page layout. Unless it's used internally, you don't need this function.

kbase-extension/static/kbase/js/widgets/function_output/kbaseSeqCompView.js
```diff
@@ -0,0 +1,201 @@
...
+        ws_id: null,
+        ws_name: null,
+        token: null,
+        width: 1150,
+        options: {
+            ws_id: null,
+            ws_name: null
+        },
+        loadingImage: window.kbconfig.loading_gif,
+        wsUrl: window.kbconfig.urls.workspace,
```
briehl: Depending on the global window.kbconfig variable has been deprecated. There's now a helper module called `narrativeConfig` that you can use as a dependency to get urls and such (just add it to the list in the `define` statement). See here for a couple examples: https://github.com/kbase/narrative/blob/develop/docs/developer/widget_helper_modules.md#narrative-config

levinas:
Thanks @briehl for your comments with examples. I've edited the code to use the narrative utils.

kbase-extension/static/kbase/js/widgets/function_output/kbaseSeqCompView.js
```diff
@@ -0,0 +1,187 @@
...
+            this._super(options);
+
+            this.ws_name = options.ws_name;
+            this.ws_id = options.ws_id;
+            if (options.report_name && options.workspace_name) {
+                  this.ws_id = options.report_name;
+                  this.ws_name = options.workspace_name;
+            }
+	    this.ws_id = 'ecoli.dnadiff.output';
+	    this.ws_name = 'fangfang:1454986986211';
```
briehl: It looks like you're setting this.ws_id and this.ws_name here explicitly, which will override the user options...

levinas:
I forgot to take out the hardcoded example. Now removed. Thanks for spotting that.

briehl:
Yup, that's why it's nice to have extra eyes on things. :) merging.

PR #536, opened 2016-02-18
Stage version 2.0.3 for deployment on next

briehl:
Just setting this up for now. It'll get accepted at ~4:45 pm PT, but any other changes to develop will wind up here.


briehl:
As in kbase-ui - accepting this now since there are no other pending PRs for develop, and not enough time to deploy and test them on CI anyway.

PR #543, opened 2016-02-23
Allow methods to specify no output widget

msneddon:
Adds a simple check to not show an output widget if the output widget specified is: 'no-display'.

Tested this locally with Megahit, and seems to all work as expected without side effects.


briehl:
The build is failing because qgrid can't be installed. This won't be a problem on CI, since that dependency is included in a subimage. Merging.

PR #549, opened 2016-02-24
One more correction in fix for bug with sharing SDK jobs.

rsutormin:


rsutormin:
@briehl, do you think __proxy_client() was made hidden for security reason?

briehl:
Yeah, we hid it so it would have less of a chance of being exposed.

briehl:
Originally, it was only meant to be used by the kbjob_manager module, so I made it only accessible there.

PR #557, opened 2016-02-25
Added Shutdown menu option

briehl:
Also tinkering with templating dialog HTML.


briehl:
Stupid qgrid breaking testing still...

PR #559, opened 2016-02-26
Fixes to side public tab

msneddon:
Public data from workspaces were not showing- this fixes the basic problems.


briehl:
Thanks, Mike!

PR #560, opened 2016-02-26
fix for missing pathway widgets

nconrad:


briehl:
Hm. Looks like Travis-CI is just clogged and not running tests for us. Happy Friday!
Merging this anyway, since it's minor.

PR #562, opened 2016-02-26
Public and Example data config

briehl:
This grew a bit more than I expected...

This changes the hard-coded Public and Example data panel sources to be configurable. It also rewrites the narrativeConfig module to fetch the new config file. This file is expected to be in the webserver's root/data_sources_config.json , but if that can't be found, it fetches a default from inside the narrative container.

The new config file for the data panels migrates the configuration JSON into an external file with blocks for CI, next, and prod deploys, which can be updated independently.

We're amassing a few front-end config files in the narrative, so those have all been moved to a new directory: `kbase-extension/static/kbase/config` (from `kbase-extension/static/kbase`). However, and here's some bold text, this means that **ALL EXISTING LOCAL NARRATIVE INSTALLATIONS WILL NEED TO BE REINSTALLED**. If you have a narrative running, just quit it, and run `scripts/install_narrative.sh` to update things. As long as you're inside a running virtualenv, you don't need any parameter flags for that script. If you're NOT running inside a venv, use --no-venv to do the update (or just install from scratch).


briehl:
Travis-CI is still choking. Should I accept this to get it on CI so we can test the deploy post-process setup? Tests pass locally, FWIW.

msneddon:
My vote is to get it in so we can test on CI now.

briehl:
Ok. Let's see what we can break!

PR #563, opened 2016-02-26
Fixed issue with Promises generated while looking up data source conf…

briehl:
…ig were crossing wires.


briehl:
Getting this in CI now, so we can do further deploy_tools testing while Keith is still sitting behind me.

PR #565, opened 2016-02-29
tweaks to display plant specific genome information, inline in the ex…

thomasoniii:
Plant genome information needs to be displayed slightly differently, so I added some modifications to kbaseGenomeAnnotation.

For our purposes, in this widget, we can assume that all plant genomes are also transcriptomes (which is handy since that way we only need to modify two tabs instead of 3 as there is no config information).

If it loads up a genome in the "Plant" domain, it does a few things -
* On the overview, it'll break up the total genes count into two separate Genes and Transcripts counts.
* it drops the length field on the genes tab, and only loads genes of type 'locus'
* it adds a new CDS tab, which is identical to the genes tab, but with genes of type 'CDS'.

All sections are helpfully commented with "XXX plants baloney", for easy future refactoring. Because I hate adding conditionals to my objects in this manner, so the work arounds implemented are baloney.

I'd say that at some point, this widget should probably be refactored into 2 (or maybe 3) separate widgets (plants, non-plants transcriptome, non-plants non-transcriptome) and maybe all wrapped up inside a factory widget that returns the appropriate sub-widget after some type checking. But that's for another day.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/kbaseGenomeAnnotation.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/kbaseGenomeAnnotation.js
@@ -4,15 +4,15 @@
  * @public
  */
 
-define(['jquery', 
+define(['jquery',
         'narrativeConfig',
-        'ContigBrowserPanel', 
-        'kbwidget', 
-        'kbaseAuthenticatedWidget', 
+        'ContigBrowserPanel',
+        'kbwidget',
+        'kbaseAuthenticatedWidget',
         'kbaseTabs',
         'jquery-dataTables',
-        'jquery-dataTables-bootstrap'], 
-function($, 
+        'jquery-dataTables-bootstrap'],
+function($,
          Config,
          ContigBrowserPanel) {
     $.KBWidget({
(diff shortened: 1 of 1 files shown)


briehl:
Thanks for working on this, Jim. I have a couple questions, though I'm pretty sure you might not have the answers :) (tagging @mlhenderson @jkbaumohl @pranjan77 @r2sunita)

Will all plant genome objects, regardless of source (uploading, taking from GenBank, or fetching from search) have 'Plant' in their domain? What happens if they don't?

Similar question for CDS - I recall all genomes having CDSes that are separate from other feature types. Should those be included there, or does that have a different, more specific meaning for plants?

samseaver:
I can answer both: the RNA-Seq, FeatureValues, and Metabolic modeling pipelines require the Genome object to have locus and CDS features. I'm not aware if any individual method requires both to be present, but as the Genome object itself is used by multiple methods/pipelines, we need it to be "complete".

Genomes that do not have the key word "Plant", nor the keyword "viridiplantae" in their taxonomy field, will not be recognized in the metabolic modeling pipeline as plants.

briehl:
Okay, well, let's get this on CI and see if any supposed-plant genomes violate those rules for the viewer. :)

mlhenderson:
Adding this as a requirement will not work for any user data uploaded
through the Genome uploader.  Reference data loaded from the Central Store
and present in KBase Search will not meet this requirement.  Data uploaded
via transscript likely would meet this requirement, but since that
overloads the Genome type, users will not know the difference.

The real problem here comes down to the data types themselves.  We can't
solve that today, but be very careful about introducing bugs here.  This
would have to be tested on a variety of data in the system.  It is
dangerous to make assumptions about the content of the data right now.

briehl:
@mlhenderson The change here is pretty lightweight. The main bug that might happen here is that the "plant" object may not have the "Plant" string in its domain, and would be treated as a prokaryote for viewing. Might be slightly confusing for the user, but it shouldn't break anything.

But, again, that's why I want to see it with various genome objects on CI.

PR #567, opened 2016-03-01
Bumped version to 2.0.6, updated release notes.

briehl:


briehl:
Very very minor change, merging now to get deployed on narrative-dev to solve another problem.

PR #570, opened 2016-03-02
Fix job runtime being displayed improperly.

briehl:
Was using the wrong time format api call to make the time string.
Also always adds the started time (for now). In some cases, the time it was finished doesn't get added to the status.

It should also show the actual time, not just the relative time, as a tooltip on mouseover. Working on that.


briehl:
Tooltips were already there, they just didn't have the fancy Bootstrap styling. Mousing over the started time will give a real timestamp.

PR #572, opened 2016-03-02
Fix missing handlebars dependency for subdata widget.

briehl:


briehl:
This'll get tested in place on narrative-dev. Merging now and skipping Travis-CI.

PR #579, opened 2016-03-03
Update build_narrative_container.sh

briehl:
A couple updates here for the container build process.
The script now takes two optional flags.
`-e | --env` = set the environment. Choose from ci, next, prod, or appdev (default is still ci)
`-d | --dev_mode` = set developer mode to either true or false - this actually takes the strings 'true' or 'false'

example = 
```
./scripts/build_narrative_container.sh -e prod -d false
or
./scripts/build_narrative_container.sh --env appdev --dev_mode true
```

The configuration update still takes place in the Python script that was `kb-git-version` and is now `kb-update-config`. The same inputs there still apply, and it also has the new flags to support the build script's change. But the deploy_tools' post-process script will probably need to be updated.


kkellerlbl:
I made the needed changes on ci, so merging.

PR #580, opened 2016-03-04
Update build_narrative_container.sh

kkellerlbl:
Do a mkdir in case $WEBROOT_DIR doesn't exist.


briehl:
This code doesn't even get called during the test build, so let's skip it.

PR #582, opened 2016-03-04
Some commented debug code in config files, updated data source config

briehl:
data source config now includes RNA-seq reads files in the Public data slideout.


briehl:
Don't merge yet. There's more change a-comin'

PR #584, opened 2016-03-08
Fix problem with red circle + button still appearing in data panel in…

briehl:
… read only mode.

See https://atlassian.kbase.us/browse/KBASE-3817


briehl:
Also put in a fix for https://atlassian.kbase.us/browse/KBASE-3779 (hopefully). The config was still pointing to the /functional-site/ path for some silly reason.

PR #587, opened 2016-03-21
Homology SDK method update

hyoo:
- bug fixed and improved Homology method output widget
- added another input type (kbaseNarrativeParameterAjaxTextSubdataInput.js), which basically loads dropdown options via ajax call.
- using new input method, homology method UI can add public genome name as an input.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/kbaseBlastOutput.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/kbaseBlastOutput.js
@@ -1,443 +1,459 @@
-define(['jquery', 'd3', 'kbwidget', 'kbaseAuthenticatedWidget', 'kbaseTabs', 'jquery-dataTables', 'jquery-dataTables-bootstrap'], function($){
-	$.KBWidget({
-		name: "kbaseBlastOutput",
-		parent: "kbaseAuthenticatedWidget",
-		version: "1.0.0",
-		ws_id: null,
-		ws_name: null,
-		token: null,
-		width: 1150,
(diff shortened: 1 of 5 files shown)


kbase-extension/static/kbase/js/widgets/function_output/kbaseBlastOutput.js
```diff
@@ -1,430 +1,459 @@
...
+                    ready(data)
+                },
+                function (data) {
+                    container.empty();
+                    container.append('<p>[Error] ' + data.error.message + '</p>');
+                });
+            return this;
+        },
+
+        getData: function () {
```
briehl: This is an old bit of code that isn't really necessary anymore. Unless you're using it internally, the getData() function isn't needed.

kbase-extension/static/kbase/js/widgets/function_output/kbaseBlastOutput.js
```diff
@@ -1,430 +1,459 @@
...
+            this.render();
+            return this;
+        },
+
+        loggedOutCallback: function (event, auth) {
+            this.token = null;
+            this.render();
+            return this;
+        },
+        uuid: function () {
```
briehl: There's a utility library that handles UUIDs now. You can include `util/String` as a dependency, and that provides the `uuid` function. There's an example in the docs here: https://github.com/kbase/narrative/blob/develop/docs/developer/widget_helper_modules.md#utilstring

briehl:
Looks good! Thanks for the updates.

PR #588, opened 2016-03-24
provide redirect to auth page if there is an auth rejection in the lua startup code

eapearson:
The lua code was simply returning 401 for auth failure, although there had been code in the past to redirect to a "login page" to force the user to authenticate. The new code routes to the current kbase-ui login widget (#login) with the required routing parameter (nextrequest). In addition, the configurability of
the auth endpoint was removed -- simple string substitution is not really adequate for building the url. Well, it could be, but it is just a single case and one cannot really reconfigure the structure of the auth endpoint without knowing how and probably changing the function.

Also, fixed some bugs in the same function (use_proxy) which probably did not manifest themselves in a measurable way, because they were related to locks, which are released automatically after a period of time. Also, removed dead and commented out code, and added comments to help (at least me) follow the logic.

### DIFF
--- a/docker/proxy_mgr.lua
+++ b/docker/proxy_mgr.lua
@@ -1051,7 +1051,7 @@ use_proxy = function(self)
     if target == nil then
         session_lock = locklib:new(M.lock_name, lock_opts)
         elapsed, err = session_lock:lock(session_key)
-        if elasped then
+        if elapsed then
             target = session_map:get(session_key)
         end
         -- still missing, assign container to session
--- a/docker/proxy_mgr.lua
+++ b/docker/proxy_mgr.lua
@@ -69,7 +69,7 @@ local use_proxy
 -- that we use to track docker containers. The ngx.shared.DICT
 -- implementation only supports basic scalar types, so we need a
 -- couple of these instead of using a common table object
--- session_map maps a session key (kbase token userid) to a whitespace seperated list of:
+-- session_map maps a session key (kbase token userid) to a whitespace separated list of:
 --          1. ip/port proxy target (eg. '127.0.0.1:49000')
 --          2. docker ID
 -- docker_map maps a docker ID to a whitespace seperated list of:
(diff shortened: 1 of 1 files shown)


briehl:
Looks good to me. Should we try putting this on narrative-dev before it gets on CI? I don't think the Jenkins/deploy tools process automatically rebuilds the Lua/Nginx stack. But I could be wrong! @kkellerlbl ?

kkellerlbl:
The Jenkins build currently reloads nginx, but I don't think that re-reads the lua code.  If it's easy to test on -dev before putting on -ci that's great, but I don't think it's essential.  As long as I know that it's happening I can kick nginx on ci after the build completes.

briehl:
I think my main question is whether it redeploys the Lua code to the location that nginx reads it from (wherever that is in the sites-enabled config). If it does that every time (and it would need a service restart), then it'll work fine.

kkellerlbl:
I think it looks at the deployed paths.  Here's what's in sites-enabled/default.

`lua_package_path "/kb/deployment/services/narrative/docker/?;/kb/deployment/services/narrative/docker/?.lua;;";
`

I think nginx might need a hard stop and start in order to read in the new lua code (since the old deployment dir gets renamed, but nginx might keep the old files open and just re-read them on restart), but that's easy to do anyway.

eapearson:
The lua code is the only bit that is deployed to /kb/deployment -- ```/kb/deployment/services/narrative/docker/```

If the lua code is copied there and nginx restarted, the lua code is active.
E.g. here is how I test the lua changes from within the working vm:

```
cd /vagrant/narrative/docker
rsync -avP resty *.lua /kb/deployment/services/narrative/docker/
service nginx restart
```

(well, and occasionally after many iterations I have to burp docker.)

kkellerlbl:
So, let's merge?

briehl:
Yup.

kkellerlbl:
Jenkins just finished, and I just did a restart of nginx.  I was able to log in successfully afterward.  Erik can you check that the change you made did what you wanted?

eapearson:
So far so good; Safari, Chrome, Firefox on Mac work as expected. Windows - Firefox, Chrome ... er, not so good on IE 11 and Edge. I'll have to dig into that now ...

The canonical test is to sign in, open a narrative from the dashboard, sign out in the first tab, then refresh the narrative. It used to give a blank page, which was a 401, now it should route to the login page, you log in, and then it redirects back to the narrative.

PR #592, opened 2016-04-01
Fixed problem with checkboxes not always loading from save with prope…

briehl:
…r state.


briehl:
Should fix https://atlassian.kbase.us/browse/KBASE-3947

PR #595, opened 2016-04-12
Make sure code cell generated JS loads each time

briehl:
A first pass at getting code cell generated widgets to appear on page reload. This is almost as equally unsecure as rendering JS in markdown, but it'll be easier to modify it to only affect KBase cells from the back end (the kernel loading sets the trust before loading the page).

Also includes some other fixes that aren't in the main develop branch yet - a deadlock that can occur on page refresh, and some updates to the genome viewer.


briehl:
This now also contains a widgetmanager module and accompanying test suite. You can invoke the widgetmanager with:

```Python
from biokbase.narrative.widgetmanager import WidgetManager
wm = new WidgetManager()
wm.show_output_widget('kbaseTabTable', obj='MyModel')
```
...etc.

Still needs a few docstrings so the Jupyter Pager will play nice, but it's a start.

eapearson:
Let's get this merged! (I can't)

briehl:
Ok!

PR #596, opened 2016-04-14
Added exporter class

briehl:
also a few minor fixes related to hosting our mishmash of javascript outside of a Tornado server.


briehl:
Okay, made some changes to make the templates "good enough" for demoing now. When we get closer to a publishing sprint, we can pick those up later.

PR #601, opened 2016-04-15
Vis cell refactor chaplin

thomasoniii:
I am become death, destroyer of worlds.

Only 170 files changed. No big deal.

Also points to new revised-widgets and ui-common submodules (revised-widgets+chaplin and chaplin, respectively), which also contain the mods for refactoring.

In my testing, it seems to mostly work most of the time for most things. input widgets may not be properly loading their state, and things which rely upon the non-AMD modeling widgets probably don't work well at all - I did work towards refactoring them, but haven't gotten 'em operational.

To do your own migrating of your own widgets - use the make-not-a-plugin.pl script in the root of ui-common#chaplin. It'll mostly handle most cases most of the time.

### DIFF
--- a/kbase-extension/static/custom/custom.js
+++ b/kbase-extension/static/custom/custom.js
@@ -688,7 +688,7 @@ define(['jquery',
         // Kickstart the Narrative loading routine once the notebook is loaded.
         $([Jupyter.events]).on('notebook_loaded.Notebook', function () {
             require(['kbaseNarrative'], function (Narrative) {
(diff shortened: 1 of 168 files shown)


briehl:
Okay, I went ahead and merged this in my fork, in the vis-cell-refactor branch. Resolved conflicts, fixed some problems. Non-AMD widgets are still broken, but *most* things work again.

I'll add a brief punch list of problems in my PR, which may override this one.

PR #602, opened 2016-04-16
Integrated Jim's and Erik's changes

briehl:
After some wailing and gnashing of teeth, I think I have most blocking kinks worked out. Most viewers seem to work again, aside from any non-AMD ones. They also work in the new offline mode, too, which was a monster to manage.

Here's some punch list things to roll in to the next sprint doc:

  * State serialization is broken now because it relied on the $elem[widgetName]('getState') idiom. Need to get a better way to get that.
  * App cells are still broken since they still do the $.kbaseNarrativeAppCell thing to get instantiated
  * A number of widgets use that string accessor thing internally for manipulating sub-widgets. E.g., I saw several uses of: `$elem.kbaseTabs(); $elem.kbaseTabs('addTab', {blah});`
  * There's a tiny fork in the Python API - `WidgetManager.show_output_widget` shows the kbaseNarrativeOutputCell, and `WidgetManager.show_external_widget` loads the CDN info and cell.
  * All existing widgets need to be tested. With the API calls, it'll be easier to test them automatically now, but we still need to set that up.


thomasoniii:
Awesome. Questions below. This is all into the vis-cell-refactor branch
now, right?

State serialization is broken now because it relied on the $elemwidgetName
<http://'getState'> idiom. Need to get a better way to get that.
App cells are still broken since they still do the $.kbaseNarrativeAppCell
thing to get instantiated
A number of widgets use that string accessor thing internally for
manipulating sub-widgets. E.g., I saw several uses of: $elem.kbaseTabs();
$elem.kbaseTabs('addTab', {blah});

These things *should* work. Is your concern that they're still using the
old style, or that it's not working? Do you have an example narrative
and/or app to test with? I'd want to make sure I'm hitting against the same
thing you were looking at.

It's possible that the plugin syntax was relying on the side effect of
something being loaded globally previously and I accidentally removed it,
but I'd think that if it was loading in advance before, it should still be
loading now, so I'm dubious about that being the culprit.

I had also updated the two ui-common subrepos with the changes - the
modules/ui-common repo (#hardy branch) can safely be replaced with chaplin
in any context that we want to use the refactoring.

But the modules/revised-widgets branch (I *think* this was a tag of
develop, but it might've been master) still needs work. That's where we've
got the modeling widgets that aren't AMD modules, as well as the
slightly-but-not-quite redundantly defined tabs and table widgets. I did a
lot of work towards refactoring those, but I know they're not working
properly.

I don't have a good idea about how much effort it'd take to bring the
revised-widgets repo up to speed, or who else is going to be affected by
it. I think it's mostly Neal's code, but I don't know beyond that. At least
hardy has effectively devolved into my playground, so I'm quite confident
that chaplin is migrated and can be dropped in when we're done.

-Jim...

nconrad:
yeah, revised-widgets was mostly a joint effort between me, Chris, and Matt.  It's pointed to a commit on the develop branch.  

I was thinking once the new widget spec is finalized/stable, I could start work on porting `src/widgets/modeling` to a repo, following the spec/documentation?   Otherwise, ya'll are welcome to do that work, but please leave the actual code as is or make a PR.

briehl:
@thomasoniii:
Those things don't work - they seem to expect something loaded globally (I think), and are trying to execute the jQuery plugin. Since they aren't still loaded as jQuery plugins, they don't want to play. I tested by just opening an empty narrative and clicking on either an App from the method panel (any of the things with the three-boxes icon at the bottom of the list), or on a few different objects in the Data panel - specifically GenomeComparison and ProteomeComparison objects threw problems with kbaseTabs.

For the state saving and loading function, the problem is that since those pieces of code are no longer bound as jQuery plugins, it tries to execute them directly, and returns the entire cell DOM node instead of the state object that the function returns, so saving a Narrative spazzes out as it tries to embed the cell's DOM node within the cell's metadata.

Honestly, I think if we're going to do the refactor, it's worth going all the way with it as soon as we can. Otherwise, when will those changes get pushed through? And how long will we have to support multiple widget code styles? There are really only a few places (though they're quite major) in the Narrative code base where these things are invoked that way, and I think they could be updated with a little work.

@nconrad:
Yup, once the dust settles and there's a clear path to documentation, migrating the modeling widgets to a separate repo is a good idea. There might be some intermediate work done to prop those widgets up in the meantime, but if so, it'll hopefully be minimal and won't disrupt the existing code. That's on the head of ui-common#develop, right? Not some older commit?

nconrad:
@briehl, ok, cool.  No, it's pointing to a particular commit. https://github.com/kbase/narrative/tree/develop/modules  I don't think you can point to head and I don't think we would have wanted to because development in another repo would become part of the narrative at deploy time (assuming narrative submodules are updated).

thomasoniii:
Yes, I agree that just getting everything done sooner rather than later is
the best course of action. I strictly dropped in the backwards
compatibility mode to get *something* working, since before I did I hadn't
successfully pulled anything up yet. Now we've got a better plan of attack
to know the more specific places that need updating, so yanking those out
and continuing to lumber forward seems wise.

The version of revised-widgets that was hanging off of
vis-cell-refactor-chaplin (and I assume now off of vis-cell-refactor)
should be an updated version of the revised-widgets tag that
narrative#develop was pointing to previously. Unless I screwed something
up, of course. 😇

So presumably we're at least partially the way there on that repo.

briehl:
Yup. Everything seems to be lined up right. So, should we just merge this PR in, so we're all on the same, partially broken, page? It's still separated from the develop branch, so there's no danger of conflict there, yet.

briehl:
I'll go ahead and merge this now so Paramvir can set up his demo things. I'll be doing another PR later today with some doc changes in a couple manager modules.

PR #605, opened 2016-04-19
Minor changes to widgetmanager API and tests.

briehl:
* renamed `get_widget_inputs` -> `print_widget_inputs` because that's what it does.
* Made a few things private that won't likely need to be exposed easily
  * Methods: `check_tag`, `load_all_widget_info`,
  * Attributes: `version_tags`, `default_narrative_input_widget`, `default_narrative_output_widget`, `cell_id_prefix`
* Added docstrings to all methods so the Jupyter Pager will be useful (e.g. `biokbase.narrative.widgetmanager.WidgetManager?` does something helpful!)


briehl:
Gonna go ahead and merge this, since it has some useful updates and doesn't disrupt anything.

PR #607, opened 2016-05-03
Updating transform options for uploading ontology objects

samseaver:


briehl:
Hey Sam, that won't actually work - each parameter in optional arguments needs to be a map that looks like this:
```
{
    "type" : "string" | "int" | "shock" (shock's only useful in the url_mapping),
    "value" : "some_name",
    or
    "param" : "parameterIdFromMethodSpec"
    "optional" : false | true
}
```

So your change should look something like this: (including the whole `transform` block):
```
"transform": {
    "output_file_name" : {
        "type" : "string",
        "value" : "Toy",
        "optional" : false
    }
}
```

samseaver:
I'll fix that, but I was working under the assumption that this block is fed directly into the configuration for the transform repository, so I mirrored the format I used in my local testing.

samseaver:
I should have been looking closer at the others for examples, but this gives me a follow-up question, I see there are sub-objects such as "default", is there any documentation on what needs to go into this block?

briehl:
Yeah, that needs to get written up in a formal document. There's inline comments in the import tab code, that go like this:

```Javascript
/**
 * mapping has five possible keys:
 * type = string, int, boolean
 * param = which method parameter name to map to
 * value = an explicit value to use (takes precedence over 'param' key)
 * optional = true or false
 * default - this is its own block with some similar keys:
 *     param - as above, maps to a method parameter
 *     value - as above, takes precedence as above, too
 *     prefix - a prefix string to put in front of the string
 *     suffix - a suffix string to append to the string
 * the 'default' parameter takes effect only if no value can be resolved
 * from the other options.
 */
```

So both `param` and `value` can be present (`value` takes precedence), and `default` is its own option with similar behavior.

I'll find some time to write this up this evening and put it in this repo's docs directory.

samseaver:
How's this?

briehl:
I think you're missing a } below line 534.

briehl:
Other than that, I think it should work. I just copied the change into my local deploy and it ran fine (do you have one of those? It's probably worth setting up for testing things like this)

samseaver:
I got it to work in my local deploy of narrative, but not in narrative-ci, I take it this isn't something that Jenkins automatically deploys?

briehl:
No, it should show up in CI on a redeploy. What's the timestamp on your Narrative? Maybe you need to update it? Or maybe it didn't deploy.

briehl:
Yup, it deployed. Last deploy was the version from this morning at 10am.

samseaver:
Ya, timestamp was out of date, updated and it works. Thanks!

PR #609, opened 2016-05-05
Working job manager

briehl:
Some big changes.
Job management is now done mostly on the kernel side. One goal this sprint was to get that entirely on the server-side, but that won't happen this sprint, since it requires some big changes to UJS.

Instead, here's the flow for starting a new job:
  1. User executes a job (with the new MethodManager.run_method() function - will be plugged into UI input widgets soon, or runnable in just a code cell).
  2. Inputs get sanitized, validated.
  3. Job gets submitted through njs_wrapper
  4. Resulting job id, method id, version, and cell id that started it (optional, mainly for input widgets) are used to compose a Job object.
  5. This gets submitted to the JobManager which uses the ipykernel.comm mechanism to push the right info forward to the UI jobs panel, and serialized in the Narrative metadata (that part's not done yet).
  6. That JobManager also has a Python `threading.Timer` loop that tells each Job to fetch their status every 10 seconds and use the same channel to push the updates forward.

On startup/page reload/kernel restart:
  1. Narrative front-end stack fires up, puts a listener on the `kernel_ready.Kernel` event.
  2. Whenever that fires (only when a *new* kernel is connected or restarted), the Jobs panel sets up the KBaseJobs comm channel, and pushes the Narrative's known job info back to the kernel side.
  3. This restarts the Timer loop and pushes updates forward.

In the end, we should be able to throw away a fairly massive amount of code and obnoxious logic from the Javascript Jobs panel. HOWEVER, this is only good for methods created with the SDK. Roman wrote a lot of management logic for the 4 or 5 different types of jobs, and in an effort to force the method cleanup issue that's been in discussion for the past 5 months (and was blessed on last week's leads call), this mechanism ignores multi-step apps, NJS scripts, and other non-SDK things. It wouldn't be too much effort to support those, but there's a lot of code that should be refactored in order to do so efficiently.

Tagging a few people who should probably see this, especially since there's been other changes with the SDK that might affect things - @rsutormin @fperez @msneddon @eapearson @scanon @cshenry @psdehal 

Still todo:
  * testing for jobmanager.py
  * improve testing for specmanager.py
  * testing for job.py
  * put some testing methods in the NarrativeTest SDK module
  * wire together the front end Jobs panel to deal with the back end updates.
  * plug together the front end input widget updates to invoke the back end changes.


briehl:
Starting a new job now pushes that job's information to the front, where it gets serialized in the narrative. This includes:
  * job id
  * method id
  * method version hash
  * tag (release, dev, beta)
  * cell id that launched it (if available)
  * set of inputs passed into the job (not including system variables like workspace name, auth token)

There's a little goofiness still with the kernel's side of the comm channel getting out of synch when the page reloads, but that should get fixed by tomorrow.

I also put together an SDK module called NarrativeTest in CI. The `test_input_params` method mainly has a bunch of input options right now, but it does do a round trip job (just makes a copy of a genome). It's a decent way to make sure that the input validator works, and the input UI works... though it may not be complete enough for that yet.

briehl:
Okay, more changes. This should be ready to integrate with @eapearson now.  
  * jobmanager lookup loop thread should behave better.
  * job status and info are properly passed through comm channel to the front end.
  * jobmanager, specmanager, and methodmanager are now proper singletons and can be loaded with:  

```Python
jm = biokbase.narrative.jobs.jobmanager.JobManager()
my_job = jm.running_jobs[job_id]

mm = biokbase.narrative.jobs.methodmanager.MethodManager()
mm.list_available_methods()
mm.method_usage(method_id)
mm.run_method(method_id, ...)

sm = biokbase.narrative.jobs.specmanager.SpecManager()
sm.list_available_methods()
sm.method_spec[tag][method_id]
...etc
```
  * added logging as well. logs track errors as well as method id and params sent to each method (along with username)

briehl:
And now the method panel generates code cells with metadata that gets read by the nbextension. Looks like this:
```
{
 "kbase": {
   "type": "method",
   "method": {
     "version": "release",
     "name": "run_megahit",
     "module": "MegaHit"
   }
 }
}

briehl:
Merging now, so @eapearson can have a crack at it.

PR #611, opened 2016-05-07
integrate input cell refactor

eapearson:
After attempting a merge, I just manually added the new code and the few changes necessary to support it. Since it is an nbextension, with most new code or changes being local to the extension, the changes have a small footprint. Yes, after a couple hour struggling with bad merges, it only took 15 minutes to manually merge the code. Sigh.
The examples I was working with locally (method cells in a narrative) still work, and now I'll proceed this weekend to wire it up to the method manager and complete the cell lifecycle support (adding, deleting, updating). At present it is only works when a notebook is loaded with the properly configured code cells.
I do want it to also continue working with a local fresh jupyter 4.2 install, so will push back some changes that were made to support the narrative environment to be more generic. E.g. there is a seminal runtime object which provides hooks for auth and config, which can be mocked in a non-narrative environment and would be useful for unit testing.
So the extension lives in /nbextensions/methodCell, and is installed at the end of narrative_install.sh. Currently it is installed as a link to the source, to make development easier.
One caveat is that the JS CDN must be running. I didn't include instructions on that, bad me, but I'll do that this weekend, although it would be the same routine as for the first sprint. The CDN is located at https://github.com/eapearson/kbase-cdn-js and should be available to your browser at http://cdn.kbase.us/cdn (see narrative_paths.js if yo want to change this.) I do plan on more work to rationalize the CDN usage although the way it is being used now works and is simple ... we do in general have a bit of work in the narrative to ensure that dependencies are compatible ...

### DIFF
--- a/kbase-extension/static/custom/custom.css
+++ b/kbase-extension/static/custom/custom.css
@@ -173,8 +173,13 @@ span#notebook_name:hover {
     background-color: #FFF;
 }
 */
+
 .notebook_app .celltoolbar > .button_container {
-    width: 100%;
+    flex: 0 auto;
+}
+
+.notebook_app .celltoolbar > .button_container:nth-child(1) {
+    flex: auto;
 }
 
 .notebook_app .prompt {
--- a/kbase-extension/static/custom/custom.js
+++ b/kbase-extension/static/custom/custom.js
@@ -156,9 +156,14 @@ define(['jquery',
         cellToolbar.CellToolbar.prototype.renderToggleState = function () {
             var toggleState = this.cell.getCellState('toggleState', 'unknown');
             // Test to see if the kbaseNarrativeCellMenu is attached to this toolbar.
-            var elemData = $(this.inner_element).find('.button_container').data();
-            if (elemData && elemData['kbaseNarrativeCellMenu'])
-               $(this.inner_element).find('.button_container').kbaseNarrativeCellMenu('toggleState', toggleState);
+            // 
+            //var elemData = $(this.inner_element).find('.button_container').data();
+            //if (elemData && elemData['kbaseNarrativeCellMenu'])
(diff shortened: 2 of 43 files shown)


eapearson:
fixed cell menu breakage after the merge.
also fixed layout within the cell toolbar with multiple cell button areas
still some work to do to re-route the toolbar updating -- there is still some code attempting to locate the cell toolbar jQuery plugin and update it directly.

eapearson:
Just pushed up some more changes. Highlights?
- break parameters into groups which match the spec better - input, output, parameters, advanced parameters
- advanced parameters is now again a collapse
- use bs panels for now just to make them clear. We can probably stick with bs panels, but make them more compact.
- dim (opacity < 1) unselected method cells
- fix method cell metadata - the method cell creation should have used the method id and module. Also instead of version for tag, using tag, version, and gitCommitHash. This should better support persisting the exact method used, but allow the user to become aware of method updates, and either opting in or auto-updated, or whatever.
- added new event updated.Cell so that a cell can take action when something is changed in the cell from outside. The problem is that create.Cell (which I'd say should be created.Cell, since it occurs after cell creation) happens internally in Jupyter after it thinks it is done, but this only covers the Jupyter scope for setting up a cell by way of the built-in cell types. We don't have a way of injecting our own code or data into that process. After the cell is created, we add the metadata for the method cell, but that is too late for the cell.Create event. So after setting the metadata, we emit an updated.Cell event. For the method cell this simply triggers the same "setupCell" method that is invoked when the notebook is loaded. It ignores non-method cells.

At present -- adding a new method cell works, the parameter setting works, and running "works" as before. Next is to wire up the running to create python compatible with the method manager, and to invoke the actual method!

briehl:
Great! Merging this now so I can tinker with the changes, too.

PR #612, opened 2016-05-08
Fixed bug in method code cell setup.

briehl:
The `name` field in the metadata needs to be just the method name, not the id (which is module_name/method_name).

Also tweaked the install script to put the final output after installing the extension.

I also installed and updated the CDN on narrative-dev and pointed the narrative_paths.js file there for now.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeWorkspace.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeWorkspace.js
@@ -320,43 +320,38 @@ define (
         },
 
         buildMethodCodeCell: function(spec, tag) {
-            // TODO: should probably throw an error here
-            // should not even get here!
             var methodName = "Unknown method";
-            // if (spec && spec.behavior && spec.behavior.kb_service_method) {
             if (!spec || !spec.info) {
                 console.error('ERROR build method code cell: ', spec, tag);
                 alert('Sorry, could not find this method');
                 return;
             }
-            // methodName = spec.behavior.kb_service_method;
-            // var moduleName = "Unknown module";
-            //if (spec && spec.behavior && spec.behavior.kb_service_name) {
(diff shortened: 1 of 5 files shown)


briehl:
@eapearson - I also modified methodCell/main.js to create the right Python code to run cells. Kept the old version in `updatePython_old` and created a new `updatePython` method.

But here's a question (I have to run, otherwise I'd just try to tackle it now, maybe later this evening) - how hard would it be to generate the code whenever the metadata is updated, then have the Run button just execute it? The use case here is to present the code to the user while it's being edited, and not just when the cell is run.

eapearson:
That would not be hard to do, but we start to get into something different, but I know this is one of the things you want to see.
There are a couple of things to think about though:
- the code should not be inserted until the parameters are complete and valid (the UI should enforce this, e.g. by not even allowing code view until this is true)
- in fact, what is missing from the run code right now is a whole-input validation (on my todo list)
- changes the user makes to the code cell should lock the inputs, but provide a control to allow the user to unlock them (which would result in their code changes being zapped)

Truth be gold, I've been focusing more on getting the cell/notebook mechanics just plain working, before getting to the more interesting use cases ...

eapearson:
On the other hand, it doesn't make much sense to have a code view as it is if it doesn't reflect the state of what has been built so far. It is honestly at this point just a control to show and hide the code view area. It should be smarter, of course, only becoming enabled when there is something to show, etc.

briehl:
I see what you mean. There are enough validation casese that we should handle, still.
In the meantime (at least for the demo on Tuesday) can we include another button that generates the code, but doesn't run it?

eapearson:
Yes, yes, I have this stuff open now so I'll just do it.
As an alternative to an "advanced" collapsible, I'm adding a "show/hide" advanced. For one, there are cases of advanced inputs, and since inputs should be with inputs, and parameters with parameters, it simplifies the interface to just inputs, outputs, and parameters, with a switch for showing advanced elements. It is all just rearranging deckchairs...

PR #620, opened 2016-05-20
Catching up with input cell refactor [wip]

eapearson:
much work on the method cell input widget and friends, mostly focused on integration of job status events. 
refactoring code as the design dust settles in
most communication between the layers -- app, method cell controller, cell widget, input widgets -- is by message bus. the message bus is a lightweight implementation, with no dependencies.
changes to the job management infrastructure (python, js) to support events out of band from job executing and monitoring. although functional, these events are not fully integrated into the method cell ui.
integration of job state and information into the method cell. This is still being worked out, as new job states are discovered.
job and other state are reflected in the ui in various ways, and this is in active development. 
work in progress -- it works, and i got it back to a working point last night after a bunch of refactoring to move code around, eliminate dead code, simplify through DRY, so seems like a good time to check it in.


briehl:
Great. Let's get this merged in, and I'll merge in my own changes, so we're working from the same base.

PR #622, opened 2016-05-24
Naming refactor, adjusted timestamps in jobs panel

briehl:
Some pretty destructive changes here ( @eapearson @cshenry should probably be aware... )  
  1. After chatting with the SDK team, we settled on calling all KBase methods 'Apps' - naming of the Jobs module stuff has been updated to reflect that for consistency.
    * `methodmanager` --> `appmanager`
    * `MethodManager.run_method()` --> `AppManager.run_app()`
    * `method_spec` and `method_id` references internally are now `app_spec` and `app_id`, etc.
    * Still uses the Narrative Method Store and those kinds of refs are the same, but this is a step toward being consistent.
  2. Converted from using `NJS_wrapper.run_app` to `NJS_wrapper.run_job`. This has several side-effects.
    * Uses `check_job` instead of `check_app_state` to look up Job info (should read the method spec for details).
    * Provides the `get_job_params` for easier lookup of the inputs that were sent to the job.
    * Requires that app ids are sent in as `Module.method_name` instead of `Module/method_name` (this is all handled for the user right now - it does the / --> . substitution before starting the job, and substitutes back when handling specs and serializing job info.
    * Will no longer work with any non-SDK things at all. The support was kinda shaky before, now it's not there.
    * Updated the code that the nbextension widget emits to reflect this change (made it a little more brief, too).
  3. Updated the Narrative's Jobs panel to deal with states being forwarded in a slightly different format. Timestamps should now work again.
  4. Tweaked around with how messages get sent back and forth over the `KBaseJobs` channel.
  5. Added backend tests for the different module pieces.

Still to do:
  1. Mocks for tests that run jobs in different states.
  2. Get the NarrativeTest module working again, add apps that fail in interesting but predictable ways.


eapearson:
Cool. I've been sick the last few days, but did some work over the weekend and this week when I should have been resting, just can't help myself. I'll merge this work in today and make the necessary adjustments before pushing my changes up.

briehl:
Great. Should I just go ahead and merge this into the repo now?

eapearson:
Yes, damn the torpedoes, full steam ahead!

briehl:
Lay on, Admiral!

PR #623, opened 2016-05-26
[WIP] Job _repr_ working

briehl:
Got the Job _repr_javascript_ method working. Now, starting a new job should (properly) create and return a Job object. Viewing that Job object (e.g. just doing something like: 
```
my_job = apps.run_app('blahblah', foo='bar')
my_job
```
in a code cell) should put the job status widget in place, as it defaults to _repr_javascript_
`str(my_job)` works, too, and just coughs up the job id. For more, theres `job.info()` for a text view of the job with its inputs or `job.state()` for a more programmatic view of the state, or `job.parameters()` for the set of params used to start the job in the first place.

This hooks a listener onto the bus that now broadcasts from the Jobs Panel. Should behave a little better and just receive updates whenever the Jobs panel does.

Next up, make it behave as the older style did - switch between running and error states, and put the console log in place. On instantiation it sets up a Comm to the Kernel, specifically to its sister Job object in the backend (uses the name "KBaseJob-{{job_id}}" as the channel). This can be used to push log info forward as required.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeJobStatus.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeJobStatus.js
@@ -1,27 +1,37 @@
 define([
     'jquery',
     'kbwidget',
+    'bluebird',
     'bootstrap',
     'narrativeConfig',
     'base/js/namespace',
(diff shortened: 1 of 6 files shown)


briehl:
Gonna go ahead and merge this.

PR #624, opened 2016-05-27
[WIP] Jobs panel should no longer refresh finished jobs.

briehl:
* Add a 'refresh' flag for jobs known by the kernel-side job manager. If True, will refresh their state and send to the front end on the update cycle.
* On the front-end Job panel update cycle, if a job comes through that has a 'finished' state, it'll send a message back to the kernel not to update that one anymore.
* Only Job widget things in the Job Panel that get updated get re-rendered now.
* Removed (commented out) a pile of deprecated code from the Jobs Panel.

Todo before merge:

- [x] Sort jobs in panel by last updated date (still running ones should be at the top, in order)
- [x] Wire job status widget to work better - they initialize in a kinda broken way and wait on the Job panel to update them. If it's a finished job, no update will be coming, so they should be inited with their finished state.
- [x] Rewire the job deletion button in the jobs panel properly.
- [x] Handle the case where the job was just started by a user executing `AppManager.run_app` - shouldn't have a "scroll to" button, should be aware it was started programmatically (e.g., it should be okay for a job to no longer have an associated cell - we aren't pulling info from the cell anymore anyway)


briehl:
Adding some notes here from conversations with @eapearson:
The jobs panel mainly communicates to the input widgets (and job status widget) over the Runtime bus. There's a few messages it emits:
`job-status` = a running job's status, updated every 10ish seconds. If a job goes into a terminal state (completed, error, etc.), it no longer updates.
`job-error` = emitted during an error to fetch job info due to timeout, etc.
`job-deleted` = emitted when a job gets deleted by a user action.
`job-delete-error` = emitted when failing to delete a job.

It also responds to a couple of messages:
`request-job-deletion` - asks the job panel to delete a job,
`request-job-status` - asks the job panel to fetch a particular job's status.

Those two will (eventually...) have some wrapped up Promised response. For now, they'll just emit the expected event above once done. (`request-job-status` will fire off a `job-status` event, etc.)

One of the reasons for the spammed events is because it's now possible for multiple cells and widgets to be listening for job information updates, and they'll all have to get updated. Rather than try to track the location and state of each of those widgets, they just need to subscribe to the bus and respond to the messages sent across it.

briehl:
Still missing the last two ticks. But got most everything else wired up together.

briehl:
Thought things were wired up, but was wrong about some things.
Latest changes:  
* Fix issue with comm channels in kernel overwriting each other
* Got first pass at log working - now gets cached in the kernel, and colorizes errors, too. Next up - request only one smallish window at a time.
* Added the `source_ws_objects` field to `run_job` - those should now get fed provenance where appropriate.
* Did lots of fighting with method spec default values in validation. Should work from the back end now.

briehl:
Okay, checked off all boxes. Things should run a little more smoothly now.

briehl:
Also merged down the recent changes to the develop branch from last week, re: copying.

eapearson:
Should we merge this in? I would like to merge upstream, deal with any consequences, before issuing a PR for my changes.

briehl:
It should be safe to merge. Let's go ahead.

PR #625, opened 2016-05-27
Updates to method cell extension, method cell widget, input widget, job status widget

eapearson:
... and their many friends.
Highlights include: 
- separation of code from the main method cell widget into individual to level widgets (input, input view, job status)
- new top level parameter view widget to show a read-only view of parameters/inputs (not just disabled controls, but think about the best way to represent the data without input considerations.)
- integration of pre-job execution events (emitted from kernel)
- conversion of main method cell widget to a half-way FSM device (controls the ui)
- unit testing via karma in place, set up for FSM.
- integration of changes to job state format 
- more full implementation of the method cell controls - rerun, cancel, remove (delete).
- improvements to input widgets (int widgets now show the min and max)
- many bug fixes, DRY changes, code removal.


eapearson:
So I think I'll just check this in now as a sign of life, before starting another round of work, since it sort of works as long as you try it on a nice method, like megahit.

briehl:
Great! Thanks. I'm working now on changes to the Job Panel to deal more with bus-based messages (like the other method widgets). If there are things you want it to respond to, let me know. The most obvious ones have to do with job deletion and state updating, but others might be necessary, too.

eapearson:
> On May 27, 2016, at 12:45 PM, Bill Riehl <notifications@github.com> wrote:
> 
> Great! Thanks. I'm working now on changes to the Job Panel to deal more with bus-based messages (like the other method widgets). If there are things you want it to respond to, let me know. The most obvious ones have to do with job deletion and state updating, but others might be necessary, too.
> 
I just sent you a note on job_err — I think we need to represent that in the job panel and have it sent as an event  as well. At the moment it is just eaten up, disappears from the job panel, and the method widget just spins waiting for some update on the job. 

I suppose a sane thing for the widget to do is that if no updates are received for an incomplete job after X amount of time to show a warning. The ability to inquire about a job directly would be useful in this case as a fallback as well, because it could then just resolve itself.

briehl:
Yup, good idea. I'll have it send a message over the bus on a job_err, to start with.

PR #626, opened 2016-06-02
Addressing Narrative Copy problems

briehl:
See tickets: 
* https://atlassian.kbase.us/browse/KBASE-4159
* https://atlassian.kbase.us/browse/KBASE-4154
* https://atlassian.kbase.us/browse/KBASE-2034
* https://atlassian.kbase.us/browse/NAR-849
* https://atlassian.kbase.us/browse/KBASE-4140
* https://atlassian.kbase.us/browse/NAR-850

These are all related to the same two issues:
1. The Narrative copy process is internally kinda complicated and kinda fragile. If anything breaks, you could wind up with a broken copy.
2. Copying a Narrative shouldn't also copy jobs.

Tasks to deal with them:

- [x] Refactor copying code - divorce business logic from view
- [x] Refactor view logic in the Narratives panel 
- [x] Refactor view logic in the Copy button in read-only mode (should use same widget/flow)
- [x] Make failures fail hard - delete the half-complete workspace if a clone was made.
- [x] Delete jobs from newly copied Narrative

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,15 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.1.0 (more notes will follow).
 
+### Version 2.0.8
+__Changes__
+- Numerous small fixes to text and layout of various widgets.
+- Genome view deals with plants and eukaryota better.
+- Proteome comparison widget uses SVG now.
+- Tree browser widget is properly clickable again.
+- Ontologies, Assemblies, and GenomeAnnotations are uploadable.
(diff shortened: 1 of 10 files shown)


briehl:
Job info should now get deleted from a copied narrative. The cell states might still be out of whack, though. I think the right thing to do here is to reset them to edit mode.

briehl:
Oops, meant to do this in another branch, oh well...
Snuck in another change that removes a duplicate auth cookie (the kb_narr_session cookie that isn't needed anymore).

briehl:
Now it copies the entire workspace, but NOT the narrative. The narrative gets saved in separately (after modifications), so it starts at version 1. Previous versions aren't copied, since they can cause problems if the user tries to roll back.

briehl:
This all seems to work now. Any objections to merging it?

briehl:
Well, nobody seems to mind (or noticed), so I'm going to bump the version and merge this.

PR #627, opened 2016-06-14
[WIP] Narrative updater

briehl:
### Don't merge! Not complete!

The goal here is to have an updater that will browse through cells in an old narrative and update them to the new style, if necessary.

Mainly it searches through the metadata fields to make sure it needs to update. All KBase cells have a `'kb-cell'` metadata field with info on what type of cell it is, and these are only in markdown cells. All others will be left alone.

The metadata currently looks something like this:
```
{
    kb-cell: { 
        type: "function_input", 
        method: { method information, id, params, etc }, 
        widget_state: [{
            params: {
                param1: blah,
                param2: blah,
            },
            time: ms_since_epoch
       }],
    },
}
```
and will soon look like this:
```
{
    kbase: {
        type: 'method',
        app_id: 'Module/app',
        state: {
              params, etc.
        }
}
```

The focus here is on methods, other things will be adjusted if time permits.
Apps are being deprecated, so they'll be replaced with an obsolescence warning.

- [x] Update method cells
- [x] Update app cells (just flag them so the front end can put an "obsolete!" label and block them)
- [ ] Output cells
  - [ ] Parse code and reproduce as code cells
  - [ ] Auto-run the code cell and produce output in the output area


briehl:
Got method cells updated, got older app cells flagged. Now app cells come with a metadata like this:
```
{
    'kbase': {
        'old_app': true
        'info': { old info structure }
    }
}
```

One problem, though - we never updated method cells to know what tag their spec came from. We just dropped the whole spec and serialized it in the notebook (which was a silly decision). It assumes master, but the method widget should probably do something special if it can't find it. Maybe look first in 'beta' then in 'dev'? Or we just show that there's a problem?

Either way, tested this in my extra large narrative with 50+ cells and it works. (also the pure Python tests pass, though they might not be thorough enough).

@scanon @eapearson

briehl:
Merging this now since it fixes job output display.

PR #628, opened 2016-06-14
Catching up with Erik's method cell changes and friends

eapearson:
Post merge from upstream got the method cell just functional enough again to start bashing away again. Changes are mostly functional, but beginning to work on layout and fine tuning behavior, layout, and style and it shakes out. Nearly generic multi-item input widgets (they wrap the single -instance versions.) Disabled and need to re-integrate the output widget into method cell (as well as fork off a special method cell for direct invocation of run_app from a code cell). Note that these "special" method, er we should start calling them "app", cells will be placeholders for now, we won't focus on making them fully functional; their function now is to allow the method cell to focus on the primary use case.
Some high (or low) lights
- after merging with Bill's upstream changes, adjusted most message bus calls
- fixed run-status messages
- subdata widget ugly but mostly (over 50%) functional
- begun to fork off method cell use cases off of the main line - i.e. viewers.
- message bus efficiency work (just enough to avoid glaring problems), and also error detection, usage warnings, request/response.


eapearson:
This also requires installing the latest CDN.

briehl:
Is that still in https://github.com/eapearson/kbase-cdn-js/ ?

briehl:
I'll go ahead and update the CDN on narrative-dev

eapearson:
Yes, still in eapearson.

briehl:
Okay, it's up on -dev. Merging. Note that we should redirect that back to -dev in narrative_paths before redeploying there (I'll try to do that this afternoon, but maybe tomorrow).

PR #635, opened 2016-06-19
app cell inserts output cell and other changes

eapearson:
The primary change is to insert the output cell after successful app execution. This is handled by the app cell widget (currently named methodCellWidget) upon the first time it is notified that the job has reached a completed state. This may be while after it has started running and has received the job state message, or after it has re-loaded and receives the job state message.
Another more radical change is to keep references to outputs within the app cell. This is exposed in a new output widget within the app cell. 
So the app cell can show a "history" of executions. There is provided for proof of concept -- so it is not fully formed. It is already useful, though, because it allows handling outputs (deleting) directly from the app cell rather than hunting them down in the narrative. 
I added this because it was frustrating to run an app cell multiple times and not know what outputs were still lying around, and also because the app cell does need to maintain some reference to the output cell so that it know that it has been created and avoid doing so again.
A natural extension would be to also show references to objects which were output, but this is delegated to the output cell itself.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeJobsPanel.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeJobsPanel.js
@@ -215,6 +215,11 @@ define([
                 this.deleteJob(message.jobId);
             }.bind(this));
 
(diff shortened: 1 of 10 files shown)


briehl:
Great, thanks!

PR #636, opened 2016-06-20
method -> app renaming refactor, improved fsm messages

eapearson:
It was finally time to bite the bullet and ensure that confusing references to "method" are replaced with "app".  Many many files were changed, sorry about that.
In addition, in order to handle the job state transition from something -> completed, the fsm message mechanism was adjusted to emit just before and after a state change, as well as when the state is initially set. This makes state change events much easier to configure and reason about.


eapearson:
please don't accept this yet, output saving (the list of outputs stored in the app cell) is broken atm.

eapearson:
okay, that is fixed now.

briehl:
ready to merge?

eapearson:
yes

PR #641, opened 2016-06-21
Updates to utilize changes to UJS and NJSW

briehl:
This has some pretty dramatic changes.  
1. Job info is no longer stored in the Narrative object - it is maintained ONLY in UJS.
2. Newly started jobs are associated with a workspace id (and given a cell_id and run_id where appropriate).
3. Restarting the kernel or reloading the page will fetch the set of running jobs from UJS using the 'kbaseworkspace' auth strategy and the workspace id. This should solve the sharing problem and make NJP obsolete.

Note this ONLY WORKS for newly created jobs, and furthermore only for SDK-based jobs. All others are effectively obsoleted by these changes.


eapearson:
Should we merge this? I've been building off of it already anyway :)

briehl:
Yeah, let's merge.

PR #645, opened 2016-06-23
Fixed sort icons in datatables.

briehl:
There were overlapping sort icons in datatables. That was obnoxious, yet nobody reported it or even brought it up...


briehl:
I'll go ahead and merge this if there are no complaints.

PR #646, opened 2016-06-23
Catching up with develop

briehl:


briehl:
Also contains some updates to the Job Panel rendering.

PR #648, opened 2016-06-30
Apps now return code for their viewer

briehl:
Well, mostly, they now return a structure like this under widget_info (gets renamed to widgetParameters):
```
{
    "name": widget_name,
    "tag": [release | beta | dev],
    "params": {
        "param1": "value1",
        "param2": "value2",
        ...
    }
}
```

This gets parsed by the client into a block of code like:
```Python
WidgetManager().show_output_widget(
    "widget_name",
    tag="release",
    param1="value1",
    param2="value2"
)
```
and injected in the new output cell. It had the same effect as before, but now that output cell should be easy to re-run, and doesn't rely on the continued existence of an app id. (Should be more clear what it's displaying, too).

This PR also creates an "obsolete" cell in the updater - if a method can't be mapped to an SDK module (or it's an old multi-step app), one of those gets produced.

Currently, view methods (those methods that don't have a service calling behavior mapped to them) can't be updated. That really shouldn't matter, since they're expected to just do a loop-back with the widget-mapped results anyway. But ideally, those should live in an SDK module. I'll fix that in the next patch so they'll be functional either way.

Finally, there's also a few fixes:
* adjustments to job status widget
* fixed race condition that can lead to errors while registering a new job (the random error messages that would pop out in code cells complaining about modifying a dictionary while iterating)
* NaN in job panel status


eapearson:
Cool, let's merge this here, then I'll merge it locally, adjust to it, get my local changes in shape, and then turn it around.

briehl:
ok. There's a few nagging problems where app cells seem to lose state or bus connections or something (there's a lot on ongoing errors in the JS console), and the view jobs that start their life finished might be causing issues. I just tweaked some ordering in the job messages - I think the FSM might have just been freaking out a bit.

PR #654, opened 2016-07-02
Better app startup handling

briehl:
This introduces a few things.  
* `AppManager.run_local_app` - this is for running synchronous, fast-running apps. It'll still register a log call for the app run, but it'll return the Widget with the result immediately. It's mainly meant for viewers, but can be adapted for other locally running code eventually.
* Fixed issues with positional parameters from method specs.
* Fixed issue with mapping fetched parameters that are stored in job info, back onto the output widgets.
* Fixed issues with parameter mapping and transforming (making explicit ws refs from input names, etc)
* Fixed issues with error handling on run_app - shouldn't throw a stacktrace, just print an error message.
* Removed the ViewerJob idiom and branching... because that was silly. The viewerjob file will get removed next.

### DIFF
--- a/src/biokbase/narrative/jobs/app_util.py
+++ b/src/biokbase/narrative/jobs/app_util.py
@@ -1,7 +1,7 @@
 """
 Some utility functions for running KBase Apps or Methods or whatever they are this week.
 """
-__author__ = "Bill Riehl <wjriehl@lbl.gov>"
+__author__ = "Bill Riehl <wjriehl@lbl.gov>, Roman Sutormin <rsutormin@lbl.gov>"
 
 import os
 import re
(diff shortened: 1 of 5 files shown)


briehl:
@eapearson - run_local_app doesn't actually launch an app, so it probably doesn't make sense to send the `launching_job` and `launched_job` messages... but I imagine a quick `completed_job` or `completed_app` message might help for the viewer widget?

eapearson:
@briehl you were reading my mind, I was thinking of that on the walk downstairs from the office. I think we need just two messages -- a success and error, with error info (type, message, details). For errors in the output widget itself, well, it would be nice to trap what we can and route back to the viewer cell, but for some classes of error this will be internal the vis widget.
I'm basically ready to start integrating this, so either you can set up those messages and then merge it, or merge this and I can add the messages and I'm integrating it into the view cell . I'll be working on this for the better part of a day over the weekend to make up for the time I had to take off this week.

briehl:
Okay, I gave run_local_app the same treatment as run_app - it should cough up an error message through the comm channel, and a run-status with event='success' right before it returns the widget. Should I go ahead and merge?

(heading home now, so i'll merge later tonight if you want, or you can - you should have permissions set now. I'm silly and thought I couldn't set that up.)

PR #659, opened 2016-07-08
custom app widgets, improved icon display in prompt area [wip]

eapearson:
- the custom app widget functionality is currently hard-coded for the media editor, this will be fixed. The custom app widget works via a new small nbextension, which specializes a code cell to run AppManager().run_widget_app, which generates the javascript wrapper code. The mechanism is essentially to feed workspace name + app info to a widget, and let it just do what it wants. This is in contrast to apps and app viewers which use the input machinery, and apps which use the job machinery too. Some minor modification to the widget necessary.
- the icon work is mostly about getting a handle on inserting a kbase-specific icon into the cell layout and at the same time surpassing the python "prompt" which appear in code cells.

NOTE this is a work in progress PR, which should not be merged quite yet. A merge would not be harmful, but I do have a round of cleanup and refactoring esp. some final prompt tasks.


eapearson:
Code cleanup, tweaks, etc. in the last set of commits make this somewhat more palatable.

eapearson:
additional changes to support conditional building of the ui, to support toggleable advanced and developer features

eapearson:
latest commits add a primitive Narrative settings feature -- an icon in the global toolbar, associated dialog, two settings (enable/disable advanced and developer features), and storage of this in the narrative metadata. This overrides properties in the main config.json 'features' property, but a real design discussion should follow to figure out a sensible path forward. 
In any case, this was necessary now because with these features in place, a user needs to be able to control them outside of config.json!!
Note also that changes to the settings are not realized until the Narrative is reloaded (the features are used to build parts of the ui.)

briehl:
Is this safe enough to get merged in?

eapearson:
as far as I'm concerned, yes.
do you typically try out PRs, at least bigger ones, locally before merging?

briehl:
Usually with the bigger ones, but there's some big changes here, and it might be worth getting our working branch up to date and just ironing out issues there.

PR #664, opened 2016-07-14
Narrative Cell Update / Narrative v3.0.0-alpha-1

briehl:
This does a ton of things. The biggest of which is replacing the Frankenstein'd markdown cells with actual code cells that make use of the Jupyter nbextension mechanism.

It also makes non-SDK methods obsolete. The updater does its best to auto-update cells and make suggestions when it can't auto-update, but those fixes will likely be ongoing with work on CI.

For other things, check out the release notes.


briehl:
Tests are passing again!

eapearson:
Yipeeeee :)

PR #667, opened 2016-07-14
More UI updates

eapearson:
- horizontal layout of input rows
- rearrangement of stylesheets
- conditional render of cell options in toolbar (only for kbase cells)
- advanced input notice and toggle button are in the relevant cella areas (input and parameter)


briehl:
Once the tests for going from vis-cell-refactor -> develop pass, I think we should just ditch this branch and do all other PRs straight there. So maybe hold off on this one for a bit?

eapearson:
sure.

PR #674, opened 2016-07-15
Fixed typo in outputCell extension

briehl:
Co-opting this for fixing Job running things that @MrCreosote exposed

- [x] text inputs in uploader panel should shut off the Jupyter keyboard manager on focus
- [x] completed jobs should only map their outputs together from inputs that exist, and not throw Exceptions
- [x] uploader panel modules need to add all of their parameters to the run function

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,12 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.2.1 (more notes will follow).
 
+### Version 3.0.0-alpha-5
+__Changes__
+- Fixed issue when starting SDK jobs from the upload panel with numeric parameters.
+- Fixed crash bug when trying to unpack a finished job that has incomplete inputs.
+- Shut off Jupyter command-mode quick keys when a text parameter input is focused.
+
 ### Version 3.0.0-alpha-4
 __Changes__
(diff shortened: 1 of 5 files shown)


briehl:
The problem with the uploader panel wasn't that all parameters were being added, but some of them weren't being reported properly. Running `getParameters` on the text input widget with something that should be validated as an int or float was only returning a string. So sending those to the run_app function was causing problems.

briehl:
Last one - mapping from submitted parameters back to inputs. This would throw an exception if a parameter was expected as a property, but no property was found. That is, if the spec looks like:

```
 input_mapping: [{
        'target_position': 0,
        'target_property': 'foo',
        'input_parameter': 'foo_param'
 }, {
        'target_position': 0,
        'target_property': 'bar',
        'input_parameter': 'bar_param'
 }]
```
but the actual parameters were:
```
[{'bar': 123}]
```
Then trying to fetch the 'foo' value would crash. Now it just makes it None.

MrCreosote:
Did the latter problem occur with the paired end reads uploader? When there wasn't any data in the numerical entries it worked, so I assume not.

briehl:
No, I don't think so. I still think it's weird that the job ran at all - it should have failed to start instead of starting with missing parameters.

Also, a null entry or empty string is still a valid entry, so it should be fine.

briehl:
gonna go ahead and merge so it can hit CI for the night owls.

PR #675, opened 2016-07-16
Basic support for SDK functions in download panels.

rsutormin:


rsutormin:
@briehl: Hey Bill, I think it's ready to be merged. Can you please review it after travis tests are done?

rsutormin:
@msneddon: Mike, can you give it a try after Bill merges it?

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDownloadPanel.js
```diff
@@ -260,10 +325,68 @@ define (
 			self.timer = setInterval(timeLst, 5000);
 			timeLst();
         },
-        
-        downloadUJSResults: function(ujsResults, wsObjectName, unzip) {
+
+        waitForSdkJob: function(jobId, wsObjectName) {
```
briehl: If this gets started using the usual SDK job starting (the AppManager.run_app function), then we can put a listener from here onto the job itself. The Jobs Panel then communicates to all listeners as the progress goes.

Unless you need job logs, then that's a touch more complicated.

Either way, I think this is fine for now so you don't stop progress, but we can refactor to make use of the message bus.

PR #676, opened 2016-07-18
kernel-side app validation

briehl:
The App param validation is maybe a little too strict in how it adheres to the app specs. This has some updates to loosen things up a bit.

- [x] Empty strings (for string inputs and dropdowns) should become `null` before being sent to jobs. These should also be a valid entry (except for required parameters)
- [x] Empty strings for checkboxes should be treated as 'falsy' and get interpreted as `False`, with the appropriate value being used for the job
- [x] Treat 'no-display' as a valid output widget name, and ignore the display if that's the case.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,17 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.2.1 (more notes will follow).
 
+### Version 3.0.0-alpha-6
+__Changes__
+- App parameter validation updates:
+  - Empty strings in either text fields or dropdowns get transformed to null before starting the app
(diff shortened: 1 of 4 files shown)


briehl:
@eapearson - an output widget named 'no-display' means that we shouldn't make an output widget. I patched the appCellWidget to detect that and skip the rendering part... but I'm not sure if that'll break the expected state. Mind taking a look?

briehl:
Now any checkbox parameter will accept `True`, and 1 as being checked, and `False`, 0, `None`, or "" (empty string) as being unchecked.

This mostly obeys the common "Truthy" / "Falsy" rules, but is more strict in that any other values for True are rejected. So any strings, list, dicts, etc., throw an error.

briehl:
This all worked before the text edits (release notes, version), so I'm gonna go ahead and merge.

PR #678, opened 2016-07-19
Adding parameters to OntologyTranslation upload

samseaver:


briehl:
This isn't valid JSON. There's extra open brackets on lines 565 and 571.

PR #679, opened 2016-07-20
Update run_app and show_output_widget signatures

briehl:
Using keyword args for run_app (and others) doesn't work. Parameter names are defined by method spec, and can (and should!) be arbitrary strings. The problem here, is that there are other keyword args in those methods that should be retained, since it's useful to provide defaults values, or just make them optional without requiring a user to put in "None".

The solution is to change that list of keyword args for param ids into a single, positional dictionary. So we turn an app call like this:
```Python
AppManager().run_app(
    "MyModule/my_app",
    tag="dev",
    version="xyz",
    cell_id="foo",
    param1="a",
    param2="b",
    param3="c"
)
```
into this:
```Python
AppManager().run_app(
    "MyModule/my_app",
    {
        "param1": "a",
        "param2": "b",
        "param3": "c"
    },
    tag="dev",
    version="xyz",
    cell_id="foo",
)
```
or, more generally:
```Python
AppManager.run_app(app_id, tag="release", version=None, cell_id=None, run_id=None, **kwargs)
```
into:
```Python
AppManager.run_app(app_id, params, tag="release", version=None, cell_id=None, run_id=None)
```

The same applies to running an output widget viewer.

Checklist of things to update
- [x] AppManager (run_app, run_local_app, _run_app_internal, _run_local_app_internal)
- [x] AppManager tests
- [x] Job (show_output_widget)
- [x] Job tests
- [x] WidgetManager (show_output_widget)
- [x] WidgetManager tests
- [x] Data panel invoked widgets (clicking/drag & drop)
- [x] Output widgets from the App Cell
- [x] Output widgets from the Viewer Cell
- [x] Output widgets from the Widget Cell (kind of a debug/test thing, still needs updating)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,12 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.2.1 (more notes will follow).
 
+### Version 3.0.0-alpha-7
+__Changes__
+- Updated invocation signatures for AppManage.run_app, .run_local_app, WidgetManager.show_output_widget -- inputs to apps (and widgets) must now be a map where the keys are the input ids and the values are the inputs themselves.
+- Newly generated output cells auto-hide their input areas (still not ideal, since it's the generated widget, but... it's a start).
+- Fixed a couple UI typos
+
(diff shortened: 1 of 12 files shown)


briehl:
Okay, I think this is ready to go.

PR #680, opened 2016-07-20
exec widget updates

eapearson:
- tabbed areas, rather than toggle able
- update tests for python generation
- add catalog link to cell toolbar
- fix launch time bug
- don't delete job when re-running cell
- add icon to tab in tabset


eapearson:
i think this is good enough to play with, although there are some things I'm still working on. The main ones being
- a mini exe-status summary widget to appear in the collapsed cell toolbar and in the expanded cell
- style and icon for Results tab to reflect success/failure/incomlete
- results tab does not currently update when open

briehl:
OK, merging. I'll add my own PR in a bit with updates to the subdata widget.

PR #681, opened 2016-07-20
Updates to subdata selection

briehl:
There's a bunch of issues with subdata selection, this fixes all but a few. The remaining ones are:
* Searching doesn't do a good typeahead - it re-renders the search box when the set of selectable items changes
* There's currently no mechanism for typing in (and selecting) an arbitrary input value (say, 'bio1' for the biomass reaction for FBA models)
* Reverting to defaults doesn't restore the pre-selected default value

This ALSO utilizes a new backend mechanism for canceling jobs - still needs some testing and revision, most likely.


briehl:
Folks are waiting on this to get merged, so I'll pull the trigger in a moment.

PR #685, opened 2016-07-22
[WIP] UI fixes, docs

briehl:
Starting a PR here with a few planned UI fixes from the maintenance week, and a bit of long-overdue documentation. Also linking together the (semi-new) differences between canceling and deleting jobs.

- [x] [KBASE-1283](https://atlassian.kbase.us/browse/KBASE-1283) Red (+) button not going away when data panel gets emptied


briehl:
Updated main comment after meeting with Paramvir today. We decided to hide all of the Job Cancellation vs. Deletion permission stuff. A canceled job just won't appear in the Job Panel, and gets treated as deleted. This'll be fixed in an upcoming push.

briehl:
@eapearson - just a note, I had to modify the NMS image lookup config to help out with Roman's problem. Now there are two distinct urls. Instead of
```Javascript
runtime.config('services.narrative_method_store.image_url');
```
it's
```Javascript
runtime.config('services.narrative_method_store_image.url');
```
Kind of a subtle different, but needed.

I also tried to adjust how to handle jobs to keep up with the UJS changes... There's no modifying jobs after they're in a completed state, so there's no way to hide them without deleting them, for users who can't delete jobs. Instead, two things happen.
1. Users should see who created a job if it's someone else besides them. That is, a 'Created By' field shows up in the Job Info if it's made by someone else:

screen shot 2016-07-26 at 8 07 17 pm

And those have a no-op delete dialog that just says the owner can only delete them. Not much else we can do, I don't think.

2. Canceling a job from an App Cell tries to delete the job, and cancels it along the way. Also, canceled jobs are treated as deleted by the whole mechanism - they're not rendered, and a 'canceled' event just gets recognized as 'deleted' by the app cell, and it moves to a (more or less) reset state. I removed the explicit 'canceled' state from the FSM, since that was causing problems.

Anyway, I'd appreciate if you could take a look before merging.

eapearson:
just curious why the need to change the NMS image url config. I'm sure there is a good reason :) 
It just kinda breaks the pattern of services.$service.$property

MrCreosote:
> And those have a no-op delete dialog that just says the owner can only delete them.
Workspace admins can also delete them now.

briehl:
@eapearson: Since it takes from config.json, there aren't discrete blocks for each service. They're still just key-value pairs, but they get mapped together to services.$service.$url so runtime.config works the same.

services.NMS.image_url was the only outlier, so it seemed easier to just give that its own key. For now, anyway. I'm all for refactoring that config file all together.

eapearson:
ok, thanks. i'm doing a fresh local install on this PR to poke around.

briehl:
@MrCreosote yes, but we decided in an earlier meeting to leave it as simple and flattened as possible. @psdehal 's suggestion was to not even allow any deleting, but just to flag all jobs as hidden or canceled or something. That's not possible, so the simplest thing to do is try to cancel/delete those that can be, and just disallow any non-job-creator from deleting jobs they created.

Actually, is it possible to modify, in any way, a job that's reached a final state? Anyone with write access to a narrative can cancel any *running* job (which hides it), but can't *delete* a job, which is the confusing part.

MrCreosote:
> Actually, is it possible to modify, in any way, a job that's reached a final state?

No, and the reason why is answered in the question: it's in a final state.

MrCreosote:
> yes, but we decided in an earlier meeting to leave it as simple and flattened as possible. 

I don't see how allowing admins to delete jobs makes it any more complex. Instead of showing a delete button to the job owner only, you show it to the owner and admins.

eapearson:
Job cancellation does work, but it is cranky.

It looks like canceling the job while it is queued does not work. There are also conditions in which the job acts like it has been deleted but the comes back to live. This annoys the app cell because it loses the execution state, and then starts receiving job status messages so goes back into running mode with incomplete job startup info (like job-id, or launch time).
It is a little hard to diagnose from the ui due the lag.

MrCreosote:
> It looks like canceling the job while it is queued does not work.

It should. What's the error message?

> There are also conditions in which the job acts like it has been deleted but the comes back to live.

I'm assuming this is UI side? If the job deletes successfully on the UJS side, it's removed from the DB, so there's no way it can return unless there's a really bizarre bug.

briehl:
The deleted one coming back to life seems to be related to the JobManager not cleaning up properly for some reason. Refreshing the page seems to fix that.

briehl:
Ok, here's a problem: canceling someone else's job throws an error with NJSW.cancel_job (also might have to do with it already being in a terminal state? I think I'd expect that to just silently fail).

screen shot 2016-07-26 at 9 34 24 pm

MrCreosote:
Arg, I see why that is. I'll need to discuss a fix with Roman tomorrow.

MrCreosote:
And no, currently cancelling a finalized job will throw an error.

MrCreosote:
Actually, sorry, the UJS will throw an error but the NJSW checks to see if a job is already completed and if so doesn't try to cancel it.

MrCreosote:
> canceling someone else's job throws an error with NJSW.cancel_job 

This should be fixed in CI (or at least once a build is done it should be fixed.

briehl:
I also have a commit almost finished that will mirror the NJSW behavior if an attempt is made to cancel a job that's already finalized. That should help.

briehl:
Ok, I think this is ready to merge. Also contains a few other non-job-cancellation fixes.
- Red (+) not going away from data panel
- Service config not updating properly (esp. for service_wizard)
- Dialogs should all accept an Enter keystroke for "yes" or "ok" or "accept".

PR #686, opened 2016-07-25
Job related changes and other fixes

eapearson:
- job cancellation invokes request-job-cancellation (which does nothing yet)
- back end job status events split into job_status and job_status_all
  - this allows the "detection of deleted job" in the job panel to work since the logic requires that all known jobs be included in the message (which is what job_status_all is for)
  - this is a [wip] but fixes a problem with new jobs issuing a job-deleted message
- fixes to properly remove event handlers

### DIFF
--- a/kbase-extension/static/custom/custom.css
+++ b/kbase-extension/static/custom/custom.css
@@ -299,6 +299,7 @@ span#notebook_name:hover {
 }
 
 .kb-btn-icon {
+    display: inline-block;
     padding: 4px;
 }
 .kb-btn-icon .fa {
--- a/kbase-extension/static/kbase/css/appCell.css
+++ b/kbase-extension/static/kbase/css/appCell.css
@@ -58,7 +58,7 @@
 }
 .kb-app-cell .kb-app-parameter-input {
     vertical-align:middle;
-    white-space: nowrap;
+    xwhite-space: nowrap;
     xpadding-left:10px;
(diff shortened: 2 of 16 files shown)


eapearson:
Oh, and a placeholder "summary widget" for app cell job status -- this should become the same one displayed in the app cell and in the toolbar when the app cell is collapsed. This is basically the job status in a single line box  for the nth time.

eapearson:
Just noticed my console is flooding. Working on it...

eapearson:
This PR now includes updates to the subdata widget. This should be ready to merge now.
Still to do includes the incorporation of default(s) into the available items. Will work on that next.

briehl:
I'm not sure why the tests are still failing. It's apparently not providing a few dependencies correctly.

briehl:
Eh. All better. I blame cosmic rays.

eapearson:
Thanks, I was waiting for the new build to finish -- it may have been the ever-troublesome phantomjs. It could not find some js libs which really should have been there. There is one quirk of phantomjs which reared its head in the past, I see that the package.json requires ^1.9.17, and I know that the problems (js bugs) were fixed late in the 1.9 series and in 2+. Since this version spec is not deterministic, we don't really know which phantomjs may be installed (+ the upstream repo has problems.)

briehl:
Since that's only used for testing, we should just bump it to the 2+ series, or whatever version is most stable.

PR #687, opened 2016-07-26
use phantomjs ^2.1.0

eapearson:
hopefully this will reduce the chance of false negative test failures in travis.


briehl:
Ha! Guess that didn't work. :)

There's a few lines like this near the top of the testing portion (~line 1645)
```
26 07 2016 20:54:58.740:WARN [proxy]: failed to proxy /narrative/static/ext_components/requirejs-plugins/lib/text.js (connect ECONNREFUSED)
```

Somehow the connection's getting refused. Maybe the Jupyter server isn't starting up? Or hasn't started up yet?

eapearson:
The phantomjs stuff is a mess. I see (well, now I don't, it is building again!) 2.1.7, 1.9.20, and 1.9.8 seems to be the one actually running. Travis must have phantomjs built in.
But yes, it looks like phantomjs is not the culprit, I think you may be right, a race condition.

eapearson:
No, that is phantomjs, and it should be using node as the server, it should not involve the narrative, right? It simply isn't finding the files.

briehl:
No, it needs the Narrative server started up, since that combines the couple of `static` directories. Karma proxies to those.

eapearson:
oh, ok, right, it needs that directory magic. does it retry now? one test run fails with "PhantomJS 1.9.8", the other succeeds with "JS 1.9.8". I must admit I've never watched the travis build in detail.

eapearson:
not that it matters for this issue, apparently, but it looks like there will be further magic incantations required to either force karma to use the npm-installed binary or to install it directly via .travis.yml.

briehl:
I think I hit the retry button. It should keep the previous run, though!
RE: .travis.yml It's a little tricky, since this is a mixed Python and Javascript project. It's set to use their Python environment now, and just install everything it needs via bower/npm. Things might be getting overridden by the Jupyter Notebook installer, though. Probably worth a look.

eapearson:
The last set of commits include:
- exec status result tab updated when results are in
- workaround for invalid checkbox default value
- update node dependencies to see if it uses the most recent phantomjs -- it does!

A note on the last issue. PhantomJS is a "headless" web-kit based browser used for testing javascript in a browser context. Travis includes an old (1.9.8) version of phantomjs by default, but the current version is 2.1.1. Other than bugs in the old version (but I think that 1.9.8 fixes some of the most egregious ones), a more recent phantomjs will track better with real browsers. The older version of the phantomjs test integration library seems to have preferred using the system version of phantomjs (old) rather than the newer one we install. Updating this integration library seems to have fixed the problem because it looks first for a locally installed phantomjs before the system one.

briehl:
```
27 07 2016 20:23:31.339:INFO [PhantomJS 2.1.1 (Linux 0.0.0)]: Connected on socket /#bDFW7OOMLdg2ZcWiAAAA with id 92363213
...
```
it seems to be loading the right one now!
Should we merge this now?

eapearson:
Yes, I think it is good to go.
Roman's PR has js test errors which are not going away with refreshing travis, if he merges this in and pushes up those may go away.

PR #688, opened 2016-07-27
Fix for bug in Generic Client related to wrong promise in URL lookup.

rsutormin:
Also switching KBaseFeatureValues widgets back to dynamic service.


rsutormin:
Here is an example how to set dynamic service which is going to be used by Narrative for sub-data selection: 
https://github.com/kbase/feature_values/blob/dev-sdk/ui/narrative/methods/view_expression_gene_table_heatmap/spec.json#L31-L32

And here is corresponding service function:
https://github.com/rsutormin/SubdataSelectionTester/blob/master/lib/SubdataSelectionTester/SubdataSelectionTesterImpl.py#L49-L50

PR #689, opened 2016-07-27
Fix Narrative metadata

briehl:
Saving should now properly make metadata for the dashboard.


briehl:
disable codemirror editing, and hide cell control button in readonly mode
Hiding ALL the buttons might be overkill, but it is a read only mode - you shouldn't be able to interact with the structure of things very much.
Also, it's easier than cherry-picking...

briehl:
App cells should now respond properly when in read only mode.
Mostly. For most cases. For most of its buttons.
I think the glaring one that doesn't work right is the button that lets you delete an output area from the output widget.
I'm starting to agree with @eapearson that a parallel read-only app cell widget might work better? But that still screams of code duplication to me.

briehl:
Ok, I think this is probably good enough to test with some feedback.

PR #690, opened 2016-07-28
UI fixes and improvements to the app cell

eapearson:
- improved error handling for input widgets. Essentially, most errors in building an input widget should result in an error alert in the place of the widget, leaving the help and label intact. This is targeted to catch invalid input parameter specs, as well as logic errors in input widgets.
- spiffed up the run, stop, re-run buttons with icons
- add intro text to app cell settings, and removed the unnecessary advanced mode toggle. I played around with tooltips for the settings checkboxes, but it was unwieldy. I think we need more full-featured info/help pop up widget and other support (e.g. storage of help topic text outside of the code it is invoked in.)

NOTE: the icons need font awesome 4.5, which is available in the can, but the current deployment of the CDN is broken for fa 4.4 and 4.5 due to a change in the bower configuration therein. I've pushed up a fix to the CDN develop branch. Does that redeploy automatically now?


eapearson:
- also added a "job does not exist" message to handle cases in which a request is made for a job which has already been deleted. While not beautiful, it does work, allowing an app cell to reset the job state if it receives this message. This helps us deal for now with the app cell getting out of sync with the actual state of jobs. The handling of this condition previously resulted in another message in one context, the main handler for job info requests, and in exceptions in the code which handled the requests.

briehl:
I think the CDN should redeploy automatically now. Might be a question for @scanon .

briehl:
Hm. It does redeploy kbase-cdn-js, but it looks like your custom thing isn't quite there yet - there's no css subdirectory under font-awesome/4.5.0 - just less and scss. I'll check into it.

briehl:
I see. It's in the code repo, but I don't think it got built on CI. It even has the correct git hash in the Jenkins console output, but I don't think it got built?

PR #691, opened 2016-07-28
Lazy rendering

dangunter:
Lazy-rendering of output cells, i.e. do not really render until they scroll into view.

Tasks:
- [x] Add code to detect visibility and wire this to the container scroll
- [x] Get basic lazy rendering working
- [x] Manual tests (with some large pages)
- [ ] Unit tests


dangunter:
Unit tests are too hard right now, and this seems to work. I think we should merge it.

PR #693, opened 2016-07-28
Added dynamic service clients for data_api

msneddon:
Includes both JS and python clients, with a commented out example JS call in the genome viewer widget.


kbase-extension/static/kbase/js/widgets/function_output/kbaseGenomeAnnotation.js
```diff
@@ -141,6 +149,13 @@ define (
             var self = this;
             var pref = StringUtil.uuid();
 
+            // Example of calling the new API
+            //var api = new GenomeAnnotationAPI(Config.url('service_wizard'),{'token':self.token});
```
msneddon: Here is an example of using the JS client.

briehl:
great!

PR #694, opened 2016-07-29
Fixes to job panel scrolling

briehl:
Also fixes some issues with the JobManager - refreshing the page can trigger a nearly constant refresh of the jobs panel (and accompanying kernel spam), and the list_jobs function should now render the right data.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,14 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.2.1 (more notes will follow).
 
+### Version 3.0.0-alpha-12
+__Changes__
+- Fixed JobManager.list_jobs (again)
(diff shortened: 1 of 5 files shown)


briehl:
Also adding a fix in here for Jobs that fail to set up their output, then disappear. This can happen with a bad app spec (should be the usual case), then the job should report an error.

- [x] Fix problem when mapping job outputs where the result type is a string, not a list
- [ ] Fix issue with vanishing jobs when there's an error while looking them up

briehl:
The job lookup process should now be a bit more error tolerant. Also, any errors in job lookup now get added to the job's status. The only real fatal error that could occur was due to trying to map outputs together based on the App spec. If that failed, it would crash the whole process, and make that job look like it disappeared. Now, it puts that job in an error state with a reasonable message.

briehl:
Should be ready to go.

briehl:
Er, wait, I take it back, not yet.

briehl:
Okay, NOW it's ready.

PR #695, opened 2016-07-29
Fixes some front end unit test quirks

eapearson:
- adds cdn modules into test
- works against ci CDN - this should be done better, by installing a local CDN and firing it it up during testing
- adds new tests (under tests2) for some modules used by the app cell and friends
- updates makefile so tests should run on travis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ build-travis-narrative:
 	npm install && \
 	bash $(INSTALLER) --no-venv
 
-test: test-backend test-frontend-unit test-frontend-e2e
+test: test-backend test-frontend-unit test-frontend-unit-standalone test-frontend-e2e
 	@echo "done running backend and frontend test scripts"
 
 # test-backend should use nose, or the like, to test our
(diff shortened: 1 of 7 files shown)


eapearson:
observing that for a locally deployed (non-dockerized) narrative there can be a period of a few seconds after startup when it is unresponsive (e.g. returns a 503, service unavailable), why wouldn't that be true for the front end test as well? A sleep of 15 seconds after the narrative is started and before the tests are run solves the problem of failed travis tests, but it is essentially non-deterministic, probably having to do with initial contact with kbase, so this time was just by trial and error.
Maybe 30 is more reliable or 5 adequate most of the time? Maybe a loop to detect when the server is ready (e.g. try to load a javascript file, which is what the narrative server is required for in this case) would be worth the effort?
The value of a shorter sleep is shorter travis thumbs up/down, since that is a bit of a hiccup in the workflow of validating a pull request (or your own commits if you have travis turned on for your own account), but frankly since that process takes several minutes as it is this added time is not too significant (but we also don't want to add to it if we can help it.)

briehl:
> Maybe a loop to detect when the server is ready (e.g. try to load a javascript file, which is what the narrative server is required for in this case) would be worth the effort?

That makes sense. The front end tests get fired up by the `test/unit/run_tests.py` script (shamelessly lifted from an older Jupyter test runner). It's pretty out of date, but it is intended to fire up the narrative, wait until it's really started, then issue the `grunt test` command. That could be modified to change its test-for-startup to not just pay attention to the command line, but actually make a request against the server.

PR #699, opened 2016-08-03
Narrative viewer for the new GenomeAnnotation.Assembly type

msneddon:
Basics are functional, but still needs some styling improvements.  Also need to fix an intermittent Assembly server error, but that should be independent of this widget.  We will probably add more to this PR today.


msneddon:
I think this is functional enough to go out to CI now.  @eapearson: can you please take a look?  thanks.

PR #702, opened 2016-08-04
New icons for new data types

dangunter:
In the process, needed to re-work the icon colors to have an explicit mapping, analogous to their shapes.


dangunter:
note that there is some lazy-load crap in here too. sigh. well, that stuff works too afaik.

dangunter:
@eapearson You should be able to merge this, especially after merging the other PR with the lazy-load of output cells.

eapearson:
we'll find out :)

PR #707, opened 2016-08-08
Updated styles for the new GenomeAnnotation and Assembly widgets

msneddon:


msneddon:
Also updates the GenomeAnnotation clients and widget to use the new version of the service.

PR #708, opened 2016-08-08
Minor typo fix for GenomeAnnotation widget

msneddon:


jmchandonia:
There's another minor bug in this widget:  on the Overview tab, the name of the Domain Model Set that was used is not displayed.  I can fix this on Wed if it's not trivial for you to find.

msneddon:
Hmm- @jmchandonia  is that field available through the new GenomeAnnotation data api?

I'm merging this in now to make sure it gets into the next CI build, but we can add more fixes anytime.

PR #713, opened 2016-08-17
Hide the Jobs Panel

briehl:
We had a pretty detailed discussion about removing the Jobs Panel (@eapearson, @scanon , @psdehal ) for a few reasons.

1. That information is easily available in each App Cell.
2. As that panel grows, and gets clogged with completed jobs, it's very difficult to navigate.
3. Even putting toggles to only show Jobs in a particular running state wouldn't solve most UX problems.
4. We'll be migrating all of that Job viewer functionality outside of the Narrative eventually.
5. There's no real way to actually manage running jobs yet, or any system for dealing with Job management (other than starting and canceling).
6. We want to remove the ability for a user to delete the Job record. There's enough rich information there that's useful for other things that probably shouldn't be deleted.
7. It helps clean up and simplify the overall interface.

I think there could be issues with Navigating to App cells with a running job, especially if you have several that are active. A separate navigation might be implemented elsewhere in the page. Also, since all job information is stored outside the Narrative in UJS now, that should be reasonably easy to put into a page on the Dashboard.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/kbaseExpressionAnalysis.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/kbaseExpressionAnalysis.js
@@ -23,8 +23,8 @@
 		bootstrap,
 		$,
 		Plotly,
-		kbaseAuthenticatedWidget,
-		KBModeling
+		kbaseAuthenticatedWidget
+		// KBModeling
 	) {
 
     return KBWidget({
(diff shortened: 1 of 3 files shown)


eapearson:
Look at that, no job panel!

briehl:
To be fair, all I did was comment out all of the rendering code. It's trivial to turn back on. But there's enough needed logic in that module for communicating with the kernel that it seemed easier to just leave it there.

And, if there's a huge uproar, we can just flip the switch back in the other direction. :)

PR #716, opened 2016-08-18
html for cell update

dangunter:


eapearson:
Did you mean to include an entire bootstrap package? Will it conflict with the existing bootstrap? It should already be available as a module in the Narrative. Although we load bootstrap via modules, BS loads globally in the browser (styles and javascript), so the most recently loaded one will overwrite whatever was there before.

briehl:
That's all just the standalone html mockups of the app cell changes we discussed yesterday. Should be safe. Maybe we shouldn't put it here, but in a separate page somewhere? Maybe narrative-dev?

eapearson:
Yes, it isn't clear to me why we would incorporate mockups into the narrative...

briehl:
It wouldn't be IN the narrative, just part of the repo for storage and a single point of communication and documentation. I suggested it to Dan, but now I'm rethinking that!

scanon:
Did this get resolved?

PR #717, opened 2016-08-18
improvements to custom widget and custom subdata widgets.

eapearson:
The custom input widget is now mostly functional. The remaining item is to synchronize the cell parameters with the media editor (mostly relevant at cell startup time.) 

However, we may not support the custom input widget going forward, so this may not see the light of day. There are two instances of custom widget, the media editor and fba model editor, and they are in the process of being updated to use server side apps for the object updating. This change would allow the editor widget to function as a parameter widget rather than an entire input area widget.

On a related note, custom subdata widgets had gone wonky with an ill-advised attempt at a 
reset method, and this is fixed in this PR. 

Curing custom widget work, the keyboard manager was causing problems. Efforts to disable the keyboard manager were being undone while operating the custom input widgets. It ends up that Jupyter now detects when new dom nodes are inserted, and will hook the keyboard manager events into any new inputs that are inserted. Since blurring on an input will cause the keyboard manager to become enabled, it just takes a short interaction with a newly inserted form control to re-enable the keyboard manager and render the narrative unusable.

The keyboard manager was ultimately disabled by overriding the default dom event hooks which tend to revive it even after it has been disabled. This was done in custom.js, in which the other jupyter monkeypatching takes place.

Also, for many widgets I'm beginning to wrap any jupyter direct calls into a jupyter proxy object to avoid sprinkling these dependencies around. Maybe this code should live in an existing module, but for the moment I'm just trying to avoid that hardcoded jupyter dependency and put it into one module (common/jupyter).

### DIFF
--- a/kbase-extension/static/custom/custom.js
+++ b/kbase-extension/static/custom/custom.js
@@ -104,6 +104,7 @@ define([
     'notebook/js/celltoolbar',
     'base/js/dialog',
     'base/js/keyboard',
+    'notebook/js/keyboardmanager',
     'notebook/js/cell',
     'common/utils',
     'kb_common/html',
@@ -124,6 +125,7 @@ define([
     cellToolbar,
     dialog,
     keyboard,
+    keyboardManager,
     cell,
     utils,
     html,
@@ -152,6 +154,7 @@ define([
 
             Jupyter.narrative = new Narrative();
             Jupyter.narrative.init();
+            Jupyter.narrative.disableKeyboardManager();
 
             /*
              * Override the move-cursor-down-or-next-cell and
@@ -769,6 +772,29 @@ define([
             //     "header will appear as the cell title." +
             //     "-->"
     };
+    
+    // KEYBOARD MANAGER
+    
+    /*
+     * Ensure that the keyboard manager does not reactivate during interaction
+     * with the Narrative.
+     * 
+     * Although we disable the keyboard manager at the outset, Jupyter will 
+     * hook into the blur event for inputs  within an inserted dom node.
+     * This causes havoc when kbase widgets manipulate the dom by inserting
+     * form controls.
+     * 
+     * So ... we just disable this behavior by overriding the register_events
+     * method.
+     * 
+     */
+    
+    (function () {
+        keyboardManager.KeyboardManager.prototype.register_events = function (e) {
+            // NOOP
+            return;
+        };
+    }());
 
 
 
--- a/kbase-extension/static/kbase/js/common/jupyter.js
+++ b/kbase-extension/static/kbase/js/common/jupyter.js
@@ -2,9 +2,11 @@
 /*jslint white:true,browser:true*/
 
 define([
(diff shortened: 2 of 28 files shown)


briehl:
Great!

I also think we should just deprecate the ability to choose a completely custom input widget that overrides the rest of the process we've put in place over the past few months. All the App cells should behave the same - they're intended to be (essentially) a fancy code generator. A user tinkers with inputs, and once those inputs are valid, code gets made.

Overriding the way that the input area works would probably cause a lot of conflict and UX confusion. Based on conversations over the past few days (and my own crazy thoughts), I think it would be better to bring in the following:

  1. A richer set of parameter input widgets. Those that we have are functional, but, in a way, kind of minimal. Though they work well enough for about 95% of cases! It would be nice to have a set of related parameters across a single row, rather than using a whole row and filling a vertical space. One example that comes to mind is setting some range limits - say flux boundaries for some reaction, or a specific range of interest for some feature-value set.
  2. Tighter coupling between parameters. Many Apps require sets of parameters that put limits on each other that should be validated before launch. For example, the RNA-seq sample set creation utility requires that each selected Reads object gets associated with a treatment label, and, optionally, a technical source (e.g. Illumina). There's currently no way to ensure that those are all valid before running the app.
  3. A more flexible approach to selecting subdata (e.g. in the FBA running App, you can select a reaction to optimize from the chosen model - that process fetches the model's "subdata" as a list of reactions to choose from). As we move to the data API, the subdata searches will need to make those API calls. I think this is being worked on now, but it would be nice to have a single generalized solution.
  4. A way to arbitrarily search for one or more objects to act as inputs. This is essentially the core of what the media and model editors do. In terms above, the user selects a workspace object as input, the App input area then fetches that object's subdata, but it ALSO provides a way to search for other inputs. Now we have two possible inputs to the app - a set of components to delete (from the object's subdata), and a set of components to add (from the search results). It would be great to link all of these together in a rigorous and general way. This could be applied to modifying genome sets, feature sets, trees, etc.

This is all kinda pie-in-the-sky right now, but I think it's worth thinking about. Tagging @eapearson @msneddon @rsutormin since we've been talking about this stuff anyway, and I think an equal amount of work will need to be done to modify the App Spec capabilities.

eapearson:
Also as we've seen there is a need for more providing input for parameters with more complex structure. This type information is available in the app service spec, but it does not make its way up the stack to the ui. There simply isn't a type system for the ui spec other than string, int, float and whatever we can glean from the spec (multiple selection implies a list of values, workspace object name implies a set of constraints on the string, etc.).  On the other hand, useful information like constraints (numeric ranges, a set of valid values for a string) are not expressed at the service spec level, just the ui.

The type information would let us more easily implement features on types and automate testing, much like implementing methods on types in <pick your favorite language>. E.g. validation, input widget, unit and generative testing, recursive input controls.

briehl:
Gonna merge this now, since it's needed for the push to next.

PR #723, opened 2016-08-25
Fixing App version confusion

briehl:
Currently just a checklist of things to address re: app versions (@msneddon @rsutormin @eapearson)

- [x] Sync App Panel refresh button with kernel-side cache.
- [x] Fetch module version info for Released apps, and use that to overwrite "version" in app specs
- [x] Don't use app spec semantic version as a testing point in App Cells
- [x] Display (somewhere) git commit hash for dev/beta apps
- [x] Adjust Catalog link to go to proper tag in slideout (optionally go to commit hash instead?)

### DIFF
--- a/kbase-extension/static/kbase/css/kbaseNarrative.css
+++ b/kbase-extension/static/kbase/css/kbaseNarrative.css
@@ -695,7 +695,7 @@ ul.pane-list li {
     text-align: justify;
 }
 
-.kb-method-list-more-div > div:nth-child(2) {
+.kb-method-list-more-div > div:last-child {
 	text-align: right;
 }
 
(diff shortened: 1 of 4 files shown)


briehl:
@msneddon - just noting that there's no way to get the semantic version for all released apps at once. Right now I use `list_basic_module_info` then use `get_module_version` on one at a time. It's speedy enough since it only looks up 5 of them in CI (only the released ones, only non-dynamic services), but we'll need a batch version eventually. I'll make a JIRA ticket.

briehl:
@rsutormin - the "..." button in method panel will show the git commit hash now in beta and dev modes. Hopefully that helps with debugging.

briehl:
Seems to be okay now. @eapearson - mind taking a look? I think I fixed the front end test issue, too? Or maybe I'm just lucky on this run...

briehl:
I'll go ahead and merge this to get other things worked on.

PR #731, opened 2016-09-02
Switching to genome annotation API in proteome comparison viewer.

rsutormin:
Also there is some clean up in mouse/tooltip interactions in two widgets related to recent change in output cell UI components (in particular it's "position: relative" style added in some top level output cell panel).


briehl:
Great!

PR #733, opened 2016-09-06
Updating Phytozome Genomes in data_source_config

samseaver:


samseaver:
A decision was made to revert back to the use of the KBaseGenomes.Genome object, and in order for them to show up in the public data drop-down, I submitted this change.

briehl:
Does this need to be applied to the other GenomeAnnotation objects referenced in that file?

samseaver:
From what I understand, we'd be making a similar change for the RefSeq genomes and removing the Ensembl genomes entirely, but given that making the change effectively removes them from the drop-down, I'd like to wait on those.

I'm actively working on the Phytozome ones now, so need the change implemented soon for others to test. Thanks.

briehl:
ok. merging now.
The failed test is a false-negative, due to something being out of sync on Travis.

samseaver:
@briehl This change needs to be applied in app-dev, what branch should I submit the PR to?

PR #738, opened 2016-09-09
"tab" buttons now utilize text not icon

eapearson:
refactored (minor) action button configuration, simplifies fsm config

### DIFF
--- a/kbase-extension/static/kbase/js/common/ui.js
+++ b/kbase-extension/static/kbase/js/common/ui.js
@@ -391,13 +391,16 @@ define([
             if (arg.classes) {
                 buttonClasses = buttonClasses.concat(arg.classes);
             }
+            if (!arg.event) {
+                arg.event = {};
+            }
 
             return button({
                 type: 'button',
                 class: buttonClasses.join(' '),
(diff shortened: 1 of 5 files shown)


eapearson:
Add more changes this morning. 
The "status box" (100x100 left most box) - rearranged the icon and status lines so that the icon is on top. Now it reads better. The status label is above the measurement, so e.g. "queued" is above the queue elapsed time, "running" is above the running elapsed time, and "completed" is above the final run time.

Also, to try it out, the buttons are bottom aligned to the display pane, and the border of the display pane matches the selected button color. Yes, a tab set if you squint. A bit more would be done to rearrange other elements to allow it to breathe on its own, though. 

One  thought is to take the status area, rearrange it horizontally, layer it on top, make the icon smaller. 

This would allow the status area to be shown alone when the cell is collapsed.

Also the status text could then be expanded, the status text in the tab area could be removed, the action button could be a bit smaller, and perhaps the action button could have the associated text label. E.g. status text could be "queued for 5min 4sec" or "app completed successfully in 1min 12sec at 5:03pm on Sep 3, 2016."

PR #739, opened 2016-09-09
Prevent the narrative from saving every time the page loads.

briehl:
The narrative currently saves whenever the app cell initializes. This is real bad when you have, say 10 or more apps, and you just refresh the page.

### DIFF
--- a/nbextensions/appCell/widgets/appCellWidget.js
+++ b/nbextensions/appCell/widgets/appCellWidget.js
@@ -473,7 +473,7 @@ define([
                 }
             };
         }
-        
+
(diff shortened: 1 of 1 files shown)


eapearson:
Sorry I re enabled autosave planning to test the denouncing, which I thought was in place for 5 seconds but I guess it wasn't. I'll make sure the dirty flag is set instead.

briehl:
No problem. I just got concerned when I was doing something in a different window, and saw my test Narrative twitching, then discovered I had over 300 saves in a Narrative I opened an hour ago. :)

PR #740, opened 2016-09-09
Genome Widget update

msneddon:
Not fully complete, but opening the PR so people can start testing in CI.  It adds a new tab to the genome widget (all the old behavior is still functional).


briehl:
Tests are passing now! Merging, and then I'll call for a CI deploy.

PR #742, opened 2016-09-12
Extract widgets from app cell

eapearson:
One round of extracting widgets from app cell -- most of the tabs now live in separate widgets in the tabs directory.
Move some common code into common/ui.
Also merged the updated results widget jnto the results tab widget, resolved conflicts.
Issues --
- showing the report in the results implies that the raw results should not be shown (they just specific the report widget), so the raw results are not shown in this case
- they also obviate the need for the report widget to be displayed in the output, but no changes there.


briehl:
Regarding the issues:
1. Yes, the raw results (just the formatted JSON) don't need to be displayed. It should focus on the report, since, eventually, all apps will come with some form of report.
2. Also yes, but since many apps are configured to have the report view widget as their output viewer (likely because they don't have a better viewer), that's fine to leave as it is. The UI shouldn't try to think around individual App design.

eapearson:
re 2: but isn't it a bit circular? I'm not sure I understand the distinction -- since we are using the app spec to show the same output in two different ways. I don't think the app spec allows the author to specify that the output should be shown as a standalone output cell -- until now it has been the only mechanism for displaying the output widget. Rather the author is specifying how to display the output of their app, wherever it is rendered.

I feel like this leads to a discussion of the relationship between output cells and app cells, how to represent multiple app cell runs, how to represent the link between app cell run/output and the specific versions of objects (esp. when rerunning and overwriting, and then using specific versions of a output object for input to another app...)

briehl:
Right, which makes it a little bit circular. Really, the output viewer has always been intended as a way to visualize the generated output objects from an application, NOT just the report. If we're covering the report view under the results tab, then designing the app to view the report is redundant.

One thing that @psdehal brought up (and is in production in some cases now) is having a list of generated or modified data objects as part of the Results tab. Clicking one of those would then produce an 'output' cell (really just a data viewer) to vis that data. It should also be a little more rich than they are now - list what version of the object it is, and when it was created. Maybe a toggle for HOW it was created, too.

PR #744, opened 2016-09-13
Hide prompt area, move icon to toolbar

eapearson:
The prompt area is hidden with a single simple css directive.
This required moving the icon into the toolbar. This was not too strenuous, requiring an additional area in the toolbar on the left. Since the toolbar is rendered to the beat of Jupyter, it requires a different method of rendering the icon. Specifically, it must request it from the host cell via our cell extension method "getIcon". This replaces "renderIcon" which was called at cell creation time, not toolbar rendering time. It is possible that there will be flickering at times, since the cell toolbar may be rendered at arbitrary times. E.g. it renders before the cell is even ready, so there is an initial state of no icon, it renders after the cell extensions are loaded, and it can also render when the cell metadata is updated as a cell is being used.
The Jupyter machinery actually wipes out the cell toolbar when it thinks it needs re-rendering, so there does not appear to be much we can do about the flickering, and we will not be able to install long-lived widgets up there (but if they are fast enough it may not matter.)
Also removed the "more..." line from the toolbar since there will be another mechanism for getting app info.
If it looks like this will work, I'll proceed to remove all of the old code for icons to reduce clutter.

### DIFF
--- a/kbase-extension/static/custom/custom.css
+++ b/kbase-extension/static/custom/custom.css
@@ -318,3 +318,7 @@ span#notebook_name:hover {
 p {
     margin-bottom: 9px;
 }
+
+.prompt {
+    display: none !important;
+}
--- a/kbase-extension/static/custom/custom.js
+++ b/kbase-extension/static/custom/custom.js
@@ -597,6 +597,18 @@ define([
 //                icon
 //            ]);
         };
+        
+        p.getIcon = function () {
+            return span();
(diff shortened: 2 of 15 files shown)


briehl:
> Also removed the "more..." line from the toolbar since there will be another mechanism for getting app info.

My next PR will be adding that. It's linked to the 'more...' right now, until we have the (i) icon for it.

PR #748, opened 2016-09-14
Ontology widgets update

thomasoniii:
Just a new chaplin pointer, which contains newer ontology dictionary and translation widgets to use tabs when in the narrative interface, as opposed to boxes as on the landing pages.


briehl:
👍

PR #749, opened 2016-09-14
Make breadth of each cell menu item clickable.

briehl:


briehl:
Gonna merge this if there are no complaints. It's really quite minor - change the `<button>` elements in the menu into `<a>` elements.

PR #750, opened 2016-09-14
Provide full object info as input for viewer widgets

msneddon:
@briehl @eapearson - What do you think about this?

Right now, we capture all the object info and pass it along the long chain to actually render a widget, then we throw it away and only keep parameters from the method spec at the very last step.  So in most cases, we lose version information, and the viewers only show the latest version.  So why don't we just always pass this full object info to all of our viewer widgets (in addition to the other parameters of course, so nothing breaks)?

This would let me make sure the Genome and Assembly viewers respect versions.  And actually, In most cases, this object info is all the viewers need, so we wouldn't even need some complicated parameter mapping for most viewers.

A possible downside is that if these widgets are created in other widgets, or outside of the viewer context, we might not have or need the full object information to populate.  It also looks like you aren't just passing around the WS info array directly, but some structure with that information processed, which means we would probably always have to process things this way in other places.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/narrativeViewers.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/narrativeViewers.js
@@ -150,7 +150,7 @@ define([
             }
             var spec = viewerInfo.specs[methodId];
             var inputParamId = spec['parameters'][0]['id'];
-            var output = {};
+            var output = {'_obj_info':o};
             _.each(spec.behavior.output_mapping, function (mapping) {
                 // Get parameter value
                 var param = getParamValue(o, mapping);


briehl:
I see. With this change, every viewer would get an '_obj_info' property with the Workspace object info? The tuple that gets returned from get_object_info_new?

I think it's a good idea. Should be a lightweight way to start supporting object versions across viewers.

msneddon:
It unfortunately would get the obj_info that the widget framework passes around, which is not the same as the workspace object info tuple.  I guess we can process it to match the WS structure again before we pass it in.

But yes, the idea is then at least each widget invoked as a viewer would get the obj_info.

briehl:
That's fine, as long as it's something consistent. In The Future (which is further out now...), we'll be moving away from direct WS calls anyway, so that specific structure shouldn't matter.

msneddon:
Cool- should we merge this then?

briehl:
Yup. 👍

eapearson:
A bit late to the party -- but wouldn't passing the object info be a bit more than is necessary? If the object reference is not adequate, why not ObjectIdentity? Although object info is well defined, it is not very easy on the eyes -- object ref is simpler and ObjectIdentity has named properties.
How is this different than supplying objects as inputs generally?

msneddon:
What's exactly in ObjectIdentity?

One of the reasons for passing everything is that we already pull the metadata for the side tab, and we can use that metadata in the widget without pulling in any additional info.  So I actually do want nearly everything returned by the WS object info (even things like saved_by, last_modified, md5, etc can be useful).  For many widgets, the first thing they do is pull the object_info from the ws based on the object...

PR #751, opened 2016-09-14
Updated tour to (mostly) match Ben's script.

briehl:
Still a few minor things to finish before merging. But mostly there!


briehl:
I think this is done enough now to get up on CI so @psdehal @thebenallen and @nlharris can have a look at it. The button is still in the hamburger menu right now, though maybe it's worth exploring a separate Help menu.

One caveat - there seems to be a flaw with the BootstrapTour library we're using where having multiple orphan steps (steps not linked to a particular element so they float in the middle of the page) causes the whole thing to fall down. I'm looking into trying this with a different version, but for now, the extra step saying "This will introduce you to some advanced features..." is being skipped.

Honestly, going through it, I don't think it's necessary.

PR #755, opened 2016-09-15
Volcano plot widget for appdev

briehl:
@scanon - I'm assuming you want to keep appdev on master, otherwise, we can merge this elsewhere.

This should bring the kbaseExpressionVolcanoPlot widget from the develop branch (with a slightly adjusted widget style) to the master branch, with no other changes. It also updates the Bower file to include that widget's external dependencies - jquery.tipsy and bootstrap-slider.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,6 +16,9 @@
     "underscore": "~1.8.3",
     "plotly.js": "git://github.com/plotly/plotly.js/#1.5.1",
     "bluebird": "3.0.6",
-    "handlebars": "^4.0.5"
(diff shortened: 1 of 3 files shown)


briehl:
Tested locally, in the appdev-ish environment. This should be ready enough for deploy.

one note - it uses the CDN on CI to get a small dependency, since that's not deployed anywhere else (to my knowledge). Shouldn't be harmful, and will get reverted once the rest of the Narrative changes make their way upstream to appdev.

scanon:
I've fine with this.

PR #763, opened 2016-09-16
Support for version tags in import tab.

rsutormin:
In addition: import tab is now cleared into original state after every upload.


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSideImportTab.js
```diff
@@ -182,76 +188,104 @@ define (
...
                 }, this)
             );
             return this;
         },
-
-        showWidget: function(type) {
+        
+        getMethodSpecs: function(callback, errorCallback) {
+            var self = this;
+            var tag = Jupyter.narrative.sidePanel.$methodsWidget.currentTag;
```
briehl: Might want to make the tag default to 'release' if that's not there. I think there's a small chance it might not initialize before this panel needs it.

rsutormin: Ok, let me add this check...

PR #766, opened 2016-09-19
Another pass at taming the Jupyter keyboard manager

eapearson:
After finding the cause of the "i" key become unresponsive in app cells, I realized that there be more dragons. The gist is that we are un-mapping some especially dangerous Jupyter keyboard captures, such as "d" to delete a cell, but we must leave some, such as ctrl-enter to run a code cell. These are the dragons that remain. In this example, it was possible to start any kbase code cell executing with ctrl-enter, or shift-enter which also selects the next cell, etc. This causes some mayhem in the cell and would certainly be confusing to the user who is not expecting it. Aside from specific examples, Jupyter may at any point introduce additional key captures.

So, in addition to this keyboard capture patches to remove the most dangerous ones, the keyboard manager is disabled upon entering kbase cells. This is achieved not by trapping focus on individual elements, but on the kbase cell itself. This has the effect of disabling the keyboard capturing while interacting the the kbase part of the cell, but allowing it to operate within jupyter controlled elements, to wit, the code input area.

Of note is that Jupyter enables the keyboard when focus enters certain cell areas, so this partly explains why the behavior could appear erratic and unpredictable. The approach taken here is to disable the Jupyter keyboard manager upon entry to kbase areas, but not to bother with enabling it.

We should keep an eye on this, because there may be other situations in which Jupyter enables the keyboard.

Ultimately, we should probably work no a Jupyter solution to this. Then again, if Jupyter provides a cell-extension mechanism, this may be part of that feature. 

Also included in this PR, some code rearrangement.


eapearson:
Also, the command palette was removed -- this actually re-exposed the removed keyboard mappings. 

It also raises the issue -- pathways to executing these dangerous notebook and cell actions still exist, and could potentially be exposed. It would be nice if there were simply a way to disable specific Jupyter features.

briehl:
This is kinda maddening. The fix that you updated (yes, my variable naming was snarky and fueled by frustration) should remove those commands from the KeyboardManager all together. Somehow, they're creeping back in.

Anyway, merged, and let's see if this fixes things. For what it's worth, here's a short path to reproducing:

Open a Markdown cell.
Enter text.
Shift-enter to render it.
Open an App cell.
Select a text field input and try typing.
Wail loudly and smash your keyboard.

eapearson:
Yes, the markdown cell was guilty of switching the keyboard manager back on and leaving it that way. Thus the focus hook on kbase cells to turn it off. Also, the "i" mapping should have been "i:i". But my worry became that such focussed fixes may fail eventually anyway, since Jupyter may add additional key mappings, and may even change the entire api for mapping keys, since it is internal!

briehl:
Maybe we could try to force the state back into Edit Mode, since Command Mode is the one that keeps causing problems? I remember tinkering with that at some point, but didn't get very far.

PR #770, opened 2016-09-21
Cleaning up Narrative code base from legacy NJS app/method support.

rsutormin:


briehl:
Just making a note here that this nukes some very VERY old code that hasn't been used in an awfully long time. Remember those "services" that were run in the Kernel, and haven't been used in almost 2 years? Those.

It also officially makes obsolete the "old" way of executing NJS jobs. There's no turning back after this merge...

briehl:
I think, from cleaning up tests with the coverage chart yesterday, something is still referencing kbjob_manager. Let me see if that's real or a false-positive before merging this.

briehl:
It's maybe a little scary, but this looks good.

PR #772, opened 2016-09-28
Implement hierarchical data view in data panel

dangunter:
I think this is ready to merge, once the kbase-client-api.js has the SetAPI in it.

https://atlassian.kbase.us/browse/KBASE-4566


briehl:
What's the 'cell-redesign' directory? Is that the old HTML view we had before invision?

dangunter:
yeah, I think so. sure, it can be deleted at some point..

PR #773, opened 2016-09-28
First draft of reads viewer by Matt and Jason

jkbaumohl:


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeOutputCell.js
```diff
@@ -32,7 +32,9 @@ define([
             this.data = this.options.data;
             this.options.type = this.options.type.toLowerCase();
             this.cell = Jupyter.narrative.getCellByKbaseId(this.options.cellId);
-            this.cell.element.trigger('hideCodeArea.cell');
+            if (this.cell) {
```
msneddon: So was this case happening when you were testing (where the cell is not found)?  @eapearson @briehl - could this happen and if so, should we be catching it or setting the cellId somewhere else?

msneddon:
Great.  I had one comment which should still be addressed by Bill or Erik, but I'm pulling this in so we can put it up on CI.

PR #775, opened 2016-09-28
First pass at a semi-generic "Editor" class of Apps

briehl:
Notes for whoever takes a look at this later this week:

This is really just a first crack at making an Editor Cell. The goal here is to make the entire input area of the App Cell able to be overridden with a widget representing a new, very specific editor. This trades the flexibility of the spec-driven app cell for something with a specific task that's not likely to be included in other apps - in this case, creating or editing a Reads Set object.

This PR has the skeleton code for firing up an editor cell and instantiating the editor widget. It does this in two steps:

1. The kbaseNarrativeWorkspace (around line 478) is currently HARD-CODED to only represent the NarrativeTest/test_editor app as an editor (it sets kbase.type = 'editor' in the metadata... we need a better way to represent that in the app specs - @msneddon ? This instantiates the editorCell notebook extension.
2. The widgets.input field of the spec is set to 'reads_set_editor'. The Configure tab has changed so that instead of creating an appParamsWidget, it creates a editorConfigurationWidget, which has a mapping to its known editors. All this does is plop the editor widget in place and wire it to listen to messages from the controlling cell.

That's pretty much it. It's extremely unfinished, but encapsulated enough to start a PR. Here's a checklist of things that need to get done.
- [ ] Find a way to update method specs to label them as editors.
- [ ] Pop the editor type -> editor path mapping into a configuration (in `nbextension/editorCell/widgets/tabs/editorConfigurationWidget.js`)
- [ ] Finish the editor widget itself. I started more or less from scratch with it, but I think (with a little glue), it can make use of a few of the more standard input types. Right now, it just renders a template and binds some events to the input boxes. The widget is in `kbase-extension/static/kbase/js/appWidgets/editors/readsSetWidget.js`
- [ ] Fix results of execution. Right now the code does execute and run as expected, but it breaks the cell in fun and exciting ways. I suspect that's because the bus messaging isn't entirely glued together right. There's so many messages!

Again, right now, it uses the same apparatus and header-area look and feel of the App cell. We can explore changing some behavior, though. For example, turn the 'run' button into a 'save' button? Unwire some of the expected long running behavior? Have it call a different backend function that's just for talking to the SetAPI? Ignore the back end all together and invoke the SetAPI from the front? I think I'd recommend against the latter, in favor of consistency with the other cell types.


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSideImportTab.js
```diff
@@ -285,13 +284,13 @@ define (
...
             var numberOfTabs = importMethodIds.length;
             if (numberOfTabs > 1) {
-                var $header = $('<div>');
-                var $body = $('<div>');
+                $header = $('<div>');
+                $body = $('<div>');
                 this.widgetPanelCard2.append($header).append($body);
             }
 
-            for (var methodPos in importMethodIds) {
```
rsutormin: May I ask why do you remove local scope here?

briehl: That's mainly a style thing imposed by following eslint. Since methodPos was already declared in the loop on line 279, it doesn't need to be declared again.

I agree it should still be forced to keep to local scope, but the linter was pretty angry about it, and I'm exploring using those, based on @sychan 's suggestion.

briehl:
Any objections to merging this now so @eapearson can work on it from the develop branch of the repo? It should be non-destructive to other apps and such.

eapearson:
Please go ahead!

PR #777, opened 2016-09-29
ReadsSet viewer based on mocks

mlhenderson:


msneddon:
Looks like pulling in the other PR generated a conflict here.  Looks like it is probably trivial though.

PR #782, opened 2016-10-03
Improved error handling, do fully qualified reference when _obj_info available

jkbaumohl:


msneddon:
Thanks.  This looks good, but I have yet more minor comments :)

While the widget is loading, we should add a loading spinner icon to the panel- @eapearson and @briehl may have some input on style guidelines for loading gifs, but I've done it this way for the genome viewer: https://github.com/kbase/narrative/blob/develop/kbase-extension/static/kbase/js/widgets/function_output/kbaseGenomeAnnotation.js#L1085

Second, more style guidelines would be appreciated here too, but if we do catch an error we should decorate the container to look like an error.  In the genome module I just wrap the error message in a div with some bootstrap error looking css classes: https://github.com/kbase/narrative/blob/develop/kbase-extension/static/kbase/js/widgets/function_output/kbaseGenomeAnnotation.js#L1122

PR #785, opened 2016-10-04
updated to use SetAPI for fetching top level Set object

mlhenderson:


kbase-extension/static/kbase/js/widgets/function_output/kbaseReadsSetView.js
```diff
@@ -50,20 +50,26 @@ define (
...
-                $self.ws = new Workspace(Config.url('workspace'),
-                                         {'token':$self.authToken()});
-                
-                return Promise.resolve($self.ws.get_objects([{'ref': this.obj_ref}]))
-                    .then(function (data) {
-                        var info = data[0]["info"],
+                this.setAPI = new SetAPI(Config.url('service_wizard'),
+                                         {'token': this.authToken()},
+                                         "dev");
+                this.wsClient = new Workspace(Config.url('workspace'),
```
msneddon: We should switch away from using the WS to get item information- the SetAPI can give you ws metadata for each item.  If there is something we want to show in the set viewer for each item, we should add it to the automatic metadata so we don't pull all the reads ws data into the client (looks like you aren't using subdata either, but just pulling in all the reads data- they are small objects but there could be a lot of them in a set...).

msneddon:
This shouldn't necessarily hold up this PR, but same comments here as for for the ReadsViewer:

1) need to handle and show errors
2) show a loading icon
3) use a specific reads set version when the _obj_info is provided

mlhenderson:
1) errors are handled for network calls, but currently only print to the
console.  Is there a standard error display for widgets?
2) loadingImage is specified, but must not be invoked.  Does this need to
be manually invoked inside the widget code?
3) should be already addressed, see lines 39-46

msneddon:
I don't think there are clear standards for most of these things (@eapearson or @briehl may have thoughts and some common libraries they've written- maybe it is time to just pick some conventions/functions and have everyone use them?  maybe we can discuss briefly after the standup tomorrow?), but there are general patterns that tend to get used- like using bootstrap alert alert-danger classes for errors.  You can see examples in the new Assembly/Genome widgets and other places in the narrative.  Same for the loading icon- you can look at those widgets for examples.  I don't think the loading widget ever happened automatically (there was and maybe still is a showLoading() method that was sometimes used).

And great on (3)- I missed that while skimming the commit diff...

PR #786, opened 2016-10-04
Initial commit for new backend editor service call

briehl:
This creates a new call in the AppManager that makes an editor request against the service wizard.

Still todo:
 - [x] Generalize parameter lookup and validation (hook into other AppManager machinery).
 - [x] Trap errors and optionally return them over the message bus (for Editor Cells).
 - [x] Make example App (over in briehl/narrative-test module... maybe migrate that to kbaseapps / kbaseIncubator?)
 - [x] tests!


briehl:
Minor change -
This now is an extension to AppManager.run_local_app (maybe should change that name...).
It examines the spec - if that spec is just a viewer, it returns the viewer. If it's an editor, it runs the Generic Client.sync_call and returns that.

But this has the advantage of making use of pre-existing machinery for handling message passing.

briehl:
Made an example app in https://github.com/briehl/NarrativeTest called test_create_set. I guess you can set Apps to call out to other modules and functions now. Awesome!

Last step = wire together parameter mapping based on @rsutormin's changes to method specs.

briehl:
Final step (probably not tonight) = rewire AppManager._map_inputs a little bit. It currently does the mapping and transformation (e.g. object name -> reference) in a loop over all parameters. It should call a single function to transform a single parameter and do that one at a time. That way we can apply the same logic to parameters inside a group, THEN transform the keys of the group if necessary.

Anyway, a little work left, hopefully tomorrow morning.

briehl:
The NarrativeTest/test_create_set now works, and uses parameter groups. It uses the same run_local_app function in the AppManager, but the spec is wired to use SetAPI.save_reads_set_v1, which gets invoked through the service wizard client. Here's a snippet of how it looks:

screen shot 2016-10-07 at 1 37 12 pm

The above parameters dictionary gets mapped (based on the spec) to:
```
[{
    'output_object_name': 'MyReadsSet',
    'data': {
        'items': [{
            'label': 'reads file 1',
            'metadata': {'key1': 'value1'},
            'ref': '11635/rhodobacterium.art.q20.int.PE.reads'
        }, {
            'label': 'reads file 2',
            'metadata': {'key2': 'value2'},
            'ref': '11635/rhodobacterium.art.q10.PE.reads'
        }],
        'description': 'New Reads Set'
    },
    u'workspace': None
}]
```
which is a valid input to the final generic client call.

Things to note here:
After a conversation with @rsutormin - there's no clear way to declaratively state that parts of a group should be transformed into references, not without some redesign of the method spec architecture itself. So any inputs that are intended to be objects are expected to be references.

Also, this only works right now for data groups sent in as a list of groups, as in the example.

briehl:
Merging now so @rsutormin can put in more changes.

PR #788, opened 2016-10-06
First draft of editor cell

eapearson:
Note that it has not quite converged to the mockup.
The main divergence is that the reads set selection and creation are handled in a header area. I'll continue to work on merging that functionality directly into the editor. 
All editor controls, other than the header, are app input controls.
The spec driving it is a hand-crafted copy of the reads set editor app.
This work was based on the viewer cell, bringing over some of Bill's initial work. Mostly because the viewer cell was closer to the mockup in layout, and simpler in design. More can be brought over later.
This involved changes to the object input controls to allow object ref as well as object name, the generic client (the current one was not changed, but a new one was created based on the sdk clients library due to some issues).
Known issues:
- no contraint on the reads set name (just be careful for the moment); will fix soon to honor workspace names;
- workaround for bug in setApi when requesting a listing in workspace in which there are no reads sets.
- not in compliance with the mockup
- does not distinguish between reads object refs in a different ws vs the existing one.
- some cases of workspace modification issues not handled (e.g. deletion of a reads object when in a set.)
- i'm sure lots of other issues...


briehl:
Probably good enough for now. Let's merge and move forward from there!

PR #790, opened 2016-10-06
don't add fake reads object when creating new

eapearson:
This was a remnant to work around the empty-set bug (add one item when creating a new set.) Not necessary now.


briehl:
👍

PR #793, opened 2016-10-07
Addressing test issues

briehl:
Appmanager tests were failing occasionally since one test in that file wasn't properly cleaning up after itself.


src/biokbase/narrative/jobs/appmanager.py
```diff
@@ -226,7 +202,7 @@ def _run_app_internal(self, app_id, params, tag, version, cell_id, run_id, **kwa
         self.spec_manager.check_app(app_id, tag, raise_exception=True)
 
         if version is not None and tag != "release":
-            if re.match(version, '\d+\.\d+\.\d+') is not None:
+            if re.match(r'\d+\.\d+\.\d+', version) is not None:
                 raise ValueError("Semantic versions only apply to released app modules. You can use a Git commit hash instead to specify a version.")
```
rsutormin: Is it allowed to have such long line in python code style in kbase?

src/biokbase/narrative/jobs/appmanager.py
```diff
@@ -780,7 +756,12 @@ def _generate_input(self, generator):
         """
         symbols = 8
         if 'symbols' in generator:
-            symbols = int(generator['symbols'])
+            try:
+                symbols = int(generator['symbols'])
```
rsutormin: You don't check that symbols is actually > 0.

src/biokbase/narrative/jobs/appmanager.py
briehl: I'm not sure... but that's all over the place in appmanager.py (and in KBase in general). I think @sychan wants to adopt the < 80 char limit.

src/biokbase/narrative/jobs/appmanager.py
briehl: I figure there are two types of things to check here. I let the int() function test if that's actually an integer string (and raise a ValueError if the int() function fails). Then, once it succeeds to make an int, line 763 tests that it meets the minimum value.

briehl: Though the exception should read "integer > 0". I'll fix that.

src/biokbase/narrative/jobs/appmanager.py
rsutormin: I feel like 100 chars in 2010s is short enough to fit in screen width. So it would look like:
```
                raise ValueError("Semantic versions only apply to released app modules. You " +
                                 "can use a Git commit hash instead to specify a version.")

```

src/biokbase/narrative/jobs/appmanager.py
rsutormin: Oh, yeah, that's checked 3 lines below!

src/biokbase/narrative/jobs/appmanager.py
rsutormin: I switched to 100 char lines in Java code recently and maintain this rule for all new code. For old code it's a matter of time of course.

briehl: I like 100 much better than 80. I'll see if the linting scripts can be modified to do that.

PR #796, opened 2016-10-10
Changes in default behavior of transformations in method-spec mappings.

rsutormin:


briehl:
This converts all workspace_name/object_name references to ws_id/obj_id/version references before submitting the jobs, right? I'm guessing this will break some apps in ways we won't expect. I'd like to see the change put in, because we should only be dealing with references anyway, but I'm unsure about doing this right now while trying to push out things to production.

@msneddon? Thoughts?

rsutormin:
@briehl: I rolled back changes related to transformation of typed non-grouped parameter refs. I decided to keep new modes by new transformation types (just to move them back as main ones later). How does it look to you?

briehl:
Looks good. Thanks! 👍

PR #797, opened 2016-10-11
Drops ui-common submodule [THOROUGHLY TEST]

thomasoniii:
This isn't critical to merge soon, but it would be nice. It should be thoroughly inspected. It cleans up technical debt and doesn't add new functionality.

This pull drops one of the ui-common submodules, the one currently pointing to the #chaplin branch, and integrates everything in #chaplin which is referenced in the narrative_paths.js file to new locations within the narrative repo. Locations were mostly guessed, but hopefully make sense.

Notes and errata -
* I did some minimal basic testing - I can see that the rna seq widgets seem to work, and the ontology widgets do. Those are the things that I've worked on recently. It needs more eyeballs on it, just to ensure things are behaving properly.
* Nothing blew up, though, which is good since this moves kbwidget around and could break everything.
* There's additional clean up required of some things - the geometry helpers aren't wrapped up as AMD modules, and there are still some RNASeq modules which are duped as RNASeqFoo and RNASeqFooNew. I'm holding this for the ominously named "phase 2", instead opting to just get things working as is and then go back and fix up the cruft later once it's functional in place.


briehl:
I'll try testing through this and hopefully merge it in today.
@scanon - you might be interested, too.

briehl:
@thomasoniii - the kbaseRNASeqPieNew widget isn't working in the update. Getting this error:
```
Cannot read property 'Blues' of undefined at http://localhost:8889/narrative/static/kbase/js/widgets/function_output/rna-seq/kbaseRNASeqPieNew.js:70:55
```

briehl:
But everything else I tried seems okay.

thomasoniii:
"Blues" is part of color brewer, which had previously been installed as an
external module inside of ui-common.

I'd added it to the bower.json file (but didn't bother to tell anyone ;-)
), so try a "bower install" in your narrative repo and hopefully that'll
resolve it.

If it shouldn't be in the bower file (maybe it should be in package.json as
an npm module instead?) then please move around the dependency as necessary.

But once it's installed and found, it should be okay.

briehl:
I installed a shiny fresh copy of the narrative from the branch with the changes, which should have done the bower install. I'll see where it got put, though.

briehl:
Okay, it's not that simple. Colorbrewer gets installed fine - the problem is it's not an AMD, so including it as one doesn't work. It's just a big file that says `var colorbrewer = { big_old_object };`

The minified one in ui-common was wrapped in a define() block that exports it. I'll see if it can't be shimmed in the narrative_paths file, but we might need to keep the same local copy.

briehl:
Just added ui-common/ext/colorbrewer.min.js and that fixes it, since it's wrapped in a define statement. Also, added that to the repo and removed the dependency on colorbrewer through bower.

briehl:
Got it added - @thomasoniii , mind giving that a test, too? I think that's the only one causing hairy issues.

thomasoniii:
Looks correct to me, so hopefully we're good to go.

briehl:
This works against all widgets I've tried. Let's go ahead and merge it now.

briehl:
Okay, I made test specs that should touch all widgets affected by this change. There's still some big gaps (like the function_input directory... though I think most of those are deprecated), but when these tests pass, that means those modules are all able to load, along with their dependencies.

Eventually, we should fill in actual unit tests for those modules, but these stubs are a start. Maybe some maintenance week tasks?

briehl:
Ok, after much wailing and gnashing of teeth, I feel this is safe enough to merge. @thomasoniii @scanon do you mind taking a look first?

PR #798, opened 2016-10-11
Kill revised widgets [NEEDS TESTING, do after #797]

thomasoniii:
Building upon pull #797, which drops the ui-common submodule dependency, this one drops the revised-widgets submodule dependency.

I haven't tested this one at all, since I don't know what to look for. It just copies the modeling directory out of the revised-widgets submodule into widgets/function_output and updates narrative_paths.js.

So someone more knowledgeable than I do needs to look at this. It will theoretically work, but I'm just being optimistic.

Due to how github sets up pulls, this also includes everything in #797, since I'd branched this one off of the kill-ui-common repo. I'd recommend doing #797, then at some point in the future, come back and do this one. Maybe it'll need a new pull against the kill-revised-widgets branch at that point? I dunnno, just raising it as a concern.


briehl:
Yes, it should just be the modeling widgets.
FBAModel, Media, FBA, etc.

briehl:
Gonna spruce this up a little bit before merging - generate tests for those widgets, test the bejeezus out of it, etc. But at a first glance, it seems to just work!

briehl:
This seems okay to me now, too. Got a basic test specs up for those modules.

briehl:
Everything passes. I vote for tearing off this last band-aid and seeing if we can test it on CI.

PR #800, opened 2016-10-13
Add support to specify narrative image in the config block.

scanon:
This allows the image to be set in the nginx init.  This helps support legacy narratives.

@briehl and @sychan  Can you review?


briehl:
Looks okay to me. So you can set the image now through the `access_by_lua` call in nginx?

PR #804, opened 2016-10-18
Editor cell saves via code cell exec & backend dynamic service call

eapearson:
This set of changes is mostly in support of the editor cell saving through generated python submitted to the narrative jupyter kernel. An overlapping set of changes to improve the editor tags along. These second changes are caught in-flight, so there are some intermediary changes in the editor, although it does still work as before.
To look for:
- need to use NarrativeTest/test_create_set/dev for the editor (it has the mapping required by the kernel methods.)
- Generates appropriate code in the code area
- Generates status messages in a "for now" status area at the bottom of the editor
- The "New" button is to test out using a create and update editor in tandem. It works, but is not fully wired together (i.e. after saving it should switch to the update editor.)
- Cloned and modified the AppManager run_local_app to just run dynamic service calls (below)
- Kernel method "AppManager().run_dynamic_service" supports both editor cell (via messages) and code cell (via python return value mechanism).


briehl:
Tests are failing because the Travis installer hasn't caught up to deal with the missing submodules yet. I'll fix that in another PR after the standup.

PR #806, opened 2016-10-19
Some post-submodule cleanup

briehl:
1. Removes `git submodule init/update` commands during install.
2. No longer installs data api local scripts (those are now captured in dynamic service calls).
3. Little bit of Travis directive adjustment so it'll hopefully cache JS objects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,13 @@ python:
   - "2.7"
 cache:
   directories:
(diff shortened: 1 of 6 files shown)


briehl:
This is gonna fail tests until CI workspace comes back to life.

briehl:
There. All better.

PR #813, opened 2016-10-22
Another pass at the hierarchical data view

briehl:
Builds on work that @dangunter did, but also has a pretty hefty refactor of the data list to do the following:
* Rewire data and set fetching to go to the NarrativeService.
* Restructure internal data model to separate between data and view models (based on conversations with @dangunter and @msneddon).
* Added a new internal ViewModel (more or less) - a list, in order, of how the data should be viewed, with a flag for whether to render that item or not.
* Rewrote the search function for some simplicity - it just updates the view model now and triggers the renderer.
* This might be a bit less performant, but now on anything that'll re-render, it detaches everything and rebuilds all items. Since we're following the scrollahead idiom, this shouldn't be a big deal. But it does make a number of things easier - searching, sorting, filtering, set toggling are now just functions of the view model.
* Fixed problems with set toggling. Now, clicking on a set will just inject the set members beneath the set object.
* Applied search/filter/etc to set members... that's gonna be kind of a weird experience, I think, but it works.
* Made a separate icon model - it was previously wired to be triggered by some event and callback mechanism that was causing icons to just pop in whenever. Now they get created on the spot.

Todo:
* Visually show somewhere how many members are in each set object... maybe in the icon?
* Don't render all set objects at once in the set toggle - needs to be reworked to fit with scrollahead, in case we have 10,000 set items.
* Maybe add some css/jquery-ui prettiness like sliding on toggle?

Todo maybe later:
* Flesh out template structure for data object view. I'd like to remove all that ugly DOM building stuff and put it in a template or two or three. There's an initial one there, but it's not used.


briehl:
Okay, tests are all better again. Any objections to me merging?

PR #814, opened 2016-10-24
Fix ci problems

briehl:
See tickets:
https://atlassian.kbase.us/browse/TASK-138 (running app problem with validator) and
https://atlassian.kbase.us/browse/TASK-139 (inserting app cell from the catalog slideout)

TASK-138 still needs some stronger tests written. I'll work on those now, but this should be ready to merge.


briehl:
Better tests are there now, with some better coverage. Things should work again. Still todo = generalize tests (they require the CI workspace with a particular public workspace - should create a new WS and upload its needed data).

But this is good to merge now.

PR #817, opened 2016-10-26
Fix Add Data button

briehl:
The red Add Data button got itself broken. Neat.


briehl:
Took this opportunity to bump the version to 3.1.0-alpha-1. Whatever's getting prepped for production release should get changed to just 3.0.0...

PR #822, opened 2016-10-26
Added view directives for things from data palettes

briehl:
Just a first pass demonstration. Should be cleaned up otherwise.

Data palette stuff has (from palette) labeled by it, with a briefcase icon (couldn't find a palette in our fonts... maybe add one?).


briehl:
Skipping tests to get out on CI. Just changed the release notes anyway.

PR #824, opened 2016-10-26
Added the reads stats to the viewer in a new tab

jkbaumohl:


kbase-extension/static/kbase/js/widgets/function_output/kbaseReadsViewer.js
```diff
@@ -302,6 +391,13 @@ define (
 		}
 */
 
+        get_uuid4: function() {
```
briehl: There's a utility module that has a UUID function in it, so you don't need to make one in every widget.
Here's how to use it:
https://github.com/kbase/narrative/blob/develop/docs/developer/widget_helper_modules.md#utilstring

briehl:
Merging this now so I can make some changes in place to support fetching by reference.

PR #828, opened 2016-10-27
[WIP] Reads view update

briehl:
Putting in some suggested changes, and getting it up to eslint snuff, and fixing errors.


briehl:
Still needs better errors - should use the createErrorDiv function from util/display. Will do that soon.

briehl:
Okay, I think this is ready. I also made a more, ahem, professional change to the palette icon (less rocketships, more chain link).

I altered the shape of the data icon for palette data, too. now it's a rounded square (same as app icons) instead of circle. Sound okay?
@psdehal @msneddon

msneddon:
I think that looks good.

briehl:
Ok. I'll merge it in, then. Gonna bump the version first.

PR #842, opened 2016-11-04
Remove last vestige of kbase_narr_session cookie

briehl:
There was a remaining reference in a test nodejs server.

This also adds some JS lint (ES lint) stuff and missing docs images.


briehl:
I know, those docs images are severely out of date already, but they at least clog the holes that are there.

This doesn't affect any running code. Any objections to merging this?

PR #843, opened 2016-11-08
Rerouted New Narrative button, changed UX for it.

briehl:
Now, when clicking the 'New Narrative' button in the Narratives panel, it does this flow:
1. Disable the button.
2. Call the NarrativeService create narrative function.
3. Build a link to the new narrative and put it in place under the button. This is the biggest shift - just popping open a window via JavaScript will get stopped by most browsers' pop-up blockers. And I think we still want it to come up in a new tab/window. So that's the feel for now.
4. Re-enable the New Narrative button when the link is clicked, and dismiss the link, while we're at it.
5. On error - show the error in a slide down alert (in the same spot) and continue to disable the create button while the error is being shown.

This also does a little bit of cleanup.

### DIFF
--- a/kbase-extension/static/kbase/js/util/timeFormat.js
+++ b/kbase-extension/static/kbase/js/util/timeFormat.js
@@ -34,7 +34,7 @@ define([], function() {
 
     /**
      * VERY simple date parser.
-     * Returns a valid Date object if that time stamp's real. 
+     * Returns a valid Date object if that time stamp's real.
      * Returns null otherwise.
      * @param {String} time - the timestamp to convert to a Date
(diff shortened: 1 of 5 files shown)


briehl:
Note, this will also include updated dropdowns to fetch data directly from the data panel (including data palette, sets, etc.)

briehl:
Object metadata should properly appear in the data panel again.

briehl:
Any objections to merging this now? It doesn't have the updated design for dropdowns, but that'll be a larger fix I'd like to do in a different PR, but based on this one.

briehl:
I talked to Erik, we agreed to merge it in. Next up is the new version of Select2...

PR #846, opened 2016-11-11
Reads set editor datapanel

eapearson:
- switch from workspace query to data panel for the object selector input widget
- this allows the editor to show reads objects from the data palette.
- disable object selector reset when object disappears from narrative workspace. an unresolved race condition exists between the control and the data panel ... the control may be rendered before the data panel is populated, and the data panel query returns no available objects. The logic of the control was that if it points to an object which does not exist, it will reset itself. This has been disabled. The effect for the user is that the control is first empty for a moment, and then the value appears. This is because there are no items for the dropdown initially.
This will become moot when this control is updated to support richer vis and query.
- some minor code cleanup


briehl:
This is putting back the submodules that should be gone - can you remove those from your branch?

PR #847, opened 2016-11-11
Grouped parameters

eapearson:
Initial support for grouped parameters, functional enough for testing, in that it can be used to populate grouped parameters for input to apps, even if all of the features are not yet implemented.
Supports common controls (text, dropdown, integer, float, texture) as grouped parameters. More coming soon.
Supports standalone groups as well as repeating (list) groups.
A single group provided to the app as a json object; a repeating group is an array of objects. 
Much of the work was simply to extend the existing control architecture to allow for nested controls. The parameter "group" is implemented as a struct control, which nests one or more parameters and their associated controls.
Within the narrative codebase, the main widgets and app cell itself were forked to allow this change to proceed without interrupting normal app cells. The changes will of course be folded back in and the forks will collapse.
The appCell2 is invoked based on the presence of one or more parameter groups.
Currently the grouped parameters do not support fancy constraint logic, and some constraints may have broken. The plan is for a required group to always enforce the "required" constraint on sub-parameters, and for optional ones to enforce it if any input contained is not empty (that is, if the user fills out a field inside of a group, it activates the group, and it will then ensure that any required field be flagged and validated as such.)
I've been using a test module (eapearson_GroupedParametersTest) for testing specs as well as the receipt input parameters by method functions. It is a java module, since the json->java layer is quite tight and validates basic types and structures automatically.


briehl:
Okay, just noting a reminder that we need to fold all the duplications and forks back together eventually.

PR #850, opened 2016-11-15
Fix problem with multiple text fields

briehl:


briehl:
@eapearson - am I doing anything crazy here? A little context - Janaka's taxonomy_service has a function 'create private taxon', and one of its advanced parameters gets parsed as type `[]text`, not `[]string`. But they should be parameterized the same way, I think, so I'm just doing that.

Any thoughts?

PR #852, opened 2016-11-15
Update to new Select2 version

briehl:
This will mostly affect the uploaders, but hopefully not too much.

It also rewires the app cell dropdowns to use select2 in (hopefully) all cases.


kbase-extension/static/kbase/js/widgets/function_input/parameter_input/kbaseNarrativeParameterTextInput.js
```diff
@@ -348,134 +348,170 @@ define([
...
-                                                    mm: key + ' - ' + d.info[10][key], info: d.info});
-                                                // allow us to show metadata match!
-                                            }
-                                        }
-                                    }
-                                }
+                tags = true;
+            }
+            console.log($input);
+            // $.fn.select2.amd.require([
```
msneddon: We probably just want to drop all this code if it's not being used?

kbase-extension/static/kbase/js/widgets/function_input/parameter_input/kbaseNarrativeParameterTextInput.js
```diff
@@ -348,134 +348,170 @@ define([
...
-                    var pageSize = self.options.wsObjSelectPageSize;
-                    query.callback({results: data.results.slice((query.page - 1) * pageSize, query.page * pageSize),
-                        more: data.results.length >= query.page * pageSize});
                 },
-                formatSelection: function (object, container) {
+                tags: tags,
+                placeholder: placeholder,
+                allowClear: true,
+                selectOnBlur: true,
+                // query: function (query) {
```
msneddon: and drop this code too?

kbase-extension/static/kbase/js/widgets/function_input/parameter_input/kbaseNarrativeParameterTextInput.js
briehl: Yes. I missed a spot.

kbase-extension/static/kbase/js/widgets/function_input/parameter_input/kbaseNarrativeParameterTextInput.js
briehl: I missed a couple of spots. But I think that needs to get ported over to the new version of Select2. I'll do that.

briehl:
Did some cleanup.
1. Dropdown contents should be sorted by last updated again.
2. Fixed import panel. Due to how the new version of Select2 works, it might behave unexpectedly (like, resetting inputs) if the data panel updates while a user is inputting something. This will only affect the import panel, though, so I think it should be okay. It would take some work to fix it, and since we're planning to migrate to the app cell style anyway...
3. Cleaned out old commented out code.

briehl:
Still to do before merging tonight

- [ ] Fix share panel with new select2 dropdown.
- [ ] Fix rendering of multiple object dropdown input.

briehl:
This should be good enough for now. There's one flaw in that for the multiple object selection widget, multiple copies of the same object are currently allowed. It (can) pass around the disallowed objects by name (or whatever the widget is configured to return), and we can have multiple things with the same name, so that causes problems.

Eventually, when input widgets emit only references, this'll go away. Until then, some more tooling is probably needed to improve the experience. But this works for now.

msneddon:
Looks good.  Things seem to work, so let's get this in so everyone can see it on CI.

One minor thing I noticed, which I think is probably very related to the flaw you pointed out, is that if I have an object in my data palette with the same name as an object in my workspace, which ever one I select, even if it isn't a multiple item selection, will still resolve to the object name in my WS.  So I can never select the one in the data palette.

msneddon:
There also seems to be something broken in the Set Editor too.

msneddon:
@briehl hmm, despite the travis build working fine here, on the develop branch the builds are failing on not being able to find a select2 dependency.

PR #854, opened 2016-11-16
Reads set editor datapanel

briehl:
@eapearson 's PR #846 re-introduced the removed submodules, which were breaking tests. This is an exact copy of that PR, but without the renewed module info.


briehl:
@msneddon - can you double check this for me?

PR #855, opened 2016-11-16
mostly working POC of autocomplete.

thomasoniii:
Doesn't actually send the param through to anything, but works well enough for demo purposes tomorrow.


briehl:
Seems to work for me, I think. Good enough to merge for demo tomorrow, at least.

It seems to have some styling problems - the autocomplete-container element should have 100% width on it. But that might just be exposing something wacky elsewhere in the widget stack.

PR #860, opened 2016-11-17
Trap errors caused by null parameter properties

briehl:
It's possible for a text or string parameter (or any param, really) to be missing its text_options (or dropdown_options) field. This change quietly just makes the parameter spec constraints objects undefined, as expected by the rest of the input widgets.


briehl:
@eapearson - I put these wrappers in since there were a few apps that were failing pretty hard when trying to load. I'm not sure this is the best way to do it, though. However, the options seem specific enough to each parameter type that this might be the most robust way to handle missing ones. Thoughts?

briehl:
This was fixed more properly elsewhere. Closing.

PR #861, opened 2016-11-19
removed cdn deps, replacing with bower-installed

eapearson:
google-code-prettify copied from cdn

- Various adjustments to the bower and narrative-paths files to load the previously cdn-loaded dependencies.
- google-code-prettify as provided by the CDN was patched to work as an AMD modules; the version available via bower is older and cannot load; so the CDN version was simply copied into ext_packages where we keep copied as opposed to bower-installed external deps.
- one fix for select2-bootstrap-theme -- the location of the files is different than in the narrative template. Perhaps the library (which is in beta btw) fixed that location.
- a 5 minute tour through the narrative, and it looks like things are still working, but perhaps we should have a quick look-see in CI before porting these changes to staging?


sychan:
@eapearson It blew up in Travis for submodules. Can you look into it?

eapearson:
yes.

eapearson:
Hmm, I though the ui-common submodules had been merged into the Narrative codebase.  Is there a need for any submodules?

eapearson:
Sorry, I must have added the submodules back in (partially, by the looks of it, but I'm not a git submodule user.) Maybe the submodules were re-added when I cloned from my fork but before I synced up with kbase upstream? In any case, they seemed to be only partially set up, so I had to remove them with git rm, clean up the .git/modules dir, and clean them out of .git/config.

sychan:
Hi  Erik,
   Thanks for doing this, I'll give @briehl a chance to look it over before merging, at the latest I'll see him tomorrow at the airport/plane/hotel, and hopefully this will be merged by Monday.

sychan:
ps.
   It would be good to do the same set of commits against the staging branch, since that is what is on track to be released.

briehl:
Hey guys, just getting caught up.
Why are we moving away from using the CDN? I know we haven't had time to work on that, but I thought the idea was to move toward using the CDN for more things site-wide, not fewer.

For select2-bootstrap-theme - I'm not against dropping it all together. I put that in because of the change to select2. Updating to v4.0.3 prevents using the other bootstrap patch (select2-bootstrap-css), and the same team is building up the new one. I think we can live without it if there's beta-fear (and it hasn't been worked on for a couple of months). Also, even with screwing up the new file location, it looks and feels okay. There are a few minor annoyances that we should deal with anyway.

eapearson:
I'm noncommittal. 
In this case it was because the narrative was not running in some environment (appdev?) in which the can was at that time not available. Steve asked if we really needed dependency up on the CDN, and I thought it would be relatively easy. It was (the only glitch being google-pretty-print).
In general I think the problem in the Narrative is bigger -- we have dependencies coming in from multiple directions -- external installed by kbase via bower, external hand-installed by kbase, kbase internal, jupyter external, jupyter internal. As far as I know, most of these are not evaluated dependency-wise, other than the runtime AMD module dependencies, which do not account for actual library/package compatibility (i.e. versions) and of course are only activated when the AMD module is actually invoked.

Some thoughts ... I think the use cases for the CDN are:

- always available dependencies (less reliance on fetching things from npm or bower / github during deployment)
- we can distribute patched libraries without
- includes only files needed (including raw packages with our web apps often exposes many more files than we intend.)
- more efficient browser caching when using iframes
- development and testing against known library distributions could make module-based widget development easier
- if we use any CDN other CDN's will not host all of the files we need, especially our own internal or patched packages.
- we can produce validated sets of packages/libraries. I.e. we build a dummy project using packages frozen at some specific version, and publish this as an entire, validated set of dependencies. Without that, one does not know if package versions are compatible with each other.
- we can produce a complete ams-compatible path configuration for such a validated set of packages -- this assists in asserting standard names for amd modules
- each new release of the CDN is by its nature completely backwards compatible -- no old dependencies are ever removed
- by analyzing CDN usage metrics, we can determine if some package is being mis-used, under-used, not used at all.

Downsides:
- requires deployment in each environment
- requires a process for redeploying in each environment when we need a new or updated package
- requires time to evaluate existing packages for recommended or required updates, new packages for inclusion, assembling sets of packages, etc.
- may require local deployment when developing against (but it is trivially easy!)
- it goes around proper dependency management, in which the project's entire dependency graph can be evaluated and validated (but if we use dependency sets this can mostly be ameliorated.)
- by the time we sort this out we will have moved into a completely new javascript technology and it will all be moot!

briehl:
I see. Now that I've been awake for a while, I remember chatting with Steve at the airport yesterday, and the missing deployments in multiple environments was causing problems with his testing environments.

That's fine. We can use bower for now so progress isn't blocked. I also think we need to rethink how we'll get the most use from a CDN going forward, especially if we ever get closer to doing sdk-ui things...

PR #863, opened 2016-11-21
Copy of code from develop that handles headless narrative runs as wel…

sychan:
…l as changing the backend services based on the ENVIRON environment variable


scripts/install_narrative_docker.sh
```diff
@@ -82,6 +100,13 @@ jupyter nbextension enable widgetCell/main --sys-prefix
 jupyter nbextension install $(pwd)/dataCell --symlink --sys-prefix
 jupyter nbextension enable dataCell/main --sys-prefix
 
+jupyter nbextension install $(pwd)/editorCell --symlink --sys-prefix
```
briehl: These extra `jupyter nbextension install` blocks reference cell types that don't exist in staging - editorCell and appCell2 should only be in develop right now.

briehl:
@sychan - the appCell2 install block also needs to be removed.

sychan:
Removed appcell2 as well.

PR #864, opened 2016-11-22
better styling on autocomplete

thomasoniii:
Better styling on the autocomplete drop down.
Also now searches on both public and private data, broken apart.


briehl:
The test build passed, but still reported an error. I just restarted it - hopefully it's just a Travis-CI glitch.

PR #866, opened 2016-11-28
More finishing work on grouped parameters

eapearson:
- added enable/disable group (experimental)
- merge rest of controls
- merge select2-based object selection control
- various fixes

Still todo: remove the duplicate parameter control code

Known issues:
- select2 object input does not restore value up on narrative loading
- advanced parameters within a group are a strange experience, especially combined with the "enable/disable group" experiment. This can be seen in the dev trimmomatic app
- this also leads to loss of (advanced grouped) parameters when reloading a narrative
- whole-app validation is temporarily disabled (so you can attempt to run an app without everything filled in.)


eapearson:
Isn't passing travis due to test configuration (karma tests in test2)

briehl:
What's the 1.7MB 'tags' file in the root folder?

eapearson:
for now just removed the failing test (it is just due to a Jupyter dependency not loading ...)

eapearson:
Damn, must be a VSC artifact.

briehl:
Yeah, that was gonna be my guess, too. From ctags.sourceforge.net (referenced at the top of that file):
```
Ctags generates an index (or tag) file of language objects found in source files that allows these items to be quickly and easily located by a text editor or other utility. 
```

VSC likes to put in random stuff. It should keep to its own directory! :)

eapearson:
Building from this PR leads to weird behavior (breakage in weird ways) upon inserting an app cell. I suspect a difference between my accreted build of the narrative and a fresh local build. Investigating...

eapearson:
It is the registration of nbextensions. Need to sort that out, so please hold off on this.

eapearson:
okay, after the latest changes I pulled down the PR, built it, and the nbextension ghosting is gone. it looks good to go.

PR #868, opened 2016-11-29
Update string dropdown input to use select2

briehl:
Saving and reloading was a little tricky, but works also.


briehl:
I fixed the conflict, but there were other functional changes that needed to be merged. Can you test again?

PR #875, opened 2016-12-03
Change data panel hierarchical view

briehl:
Adjustments to the hierarchical view.
1. Dan was right - we should have a toggle button to show set members, not just click on the icon. So it does that now, and clicking the icon always puts the object in the page.
2. Shrunk the stacked icons for sets so they take up the same width as the others.
3. Added a reads set icon - just the same as the reads icon.


briehl:
Merging this now so Paramvir can effectively review.

PR #886, opened 2016-12-08
Grouped parameters finishing work

eapearson:
validators broken into modules by type (were bundled into a single module before.)
validators invoked dynamically
validation now required to be promise
whole-param-set validation
recursive validation, required for validating lists and groups
(most of the changes above implemented to enable the latter)
fix app cell run handling to fix an annoying hang after hitting the run button
first pass at a hook in the data list module to solve the race between widgets and the data panel

after this change I'll remove large swathes of code related to the old widgets, including the obsolete "widget cell".

working on:
finishing "view only" param fields (like inputs, but read-only and with most of the machinery removed.)
custom subdata - a hairy widget that needs updating
reads set editor, which broke with this change


eapearson:
i see the travis issue, i'll attend to it.

eapearson:
Partially solved, but will need to finish up later.
The current problem is that the typeahead dependencies seem to have disappeared out of bower.json, and after adding them back in, the currently available versions don't seem compatible with the autocomplete param control.

PR #887, opened 2016-12-08
Remove reference to ci

kkellerlbl:
Remove the hardcoded reference to ci.  I'm adding a CDN to the other environments so we should no longer need this.


kkellerlbl:
Once merged I will be sure to test this in next.  We may even want to wait to push up to appdev and production till we know the CDNs actually work in those environments.

PR #889, opened 2016-12-09
Reads set editor repair

eapearson:
The reads set editor had broken after the last set of changes to app cell parameter machinery.
At the same time, there were some reads set editor changes in flight that needed to the tied up.
The reads set editor changes are mostly internal -- rearranging code so that new editors can be more readily added. This work is not yet finished. 
Although it does not conform with Paramvir's changes, the ui is tightened up a bit. There are still logical and api issues to clean up, but it does allow one to create and edit reads sets


msneddon:
Thanks @eapearson!  It's probably something minor, but when I try to run this locally, when a Narrative first starts up, I get a "Error loading editor cell extension" error popup, and I can't add editor cells to the narrative (existing ones just show code cells).

eapearson:
It looks like that was problem with previously-created editor cells. Fixed in the latest commit, and testing now.

eapearson:
And ... fixed a select 2 bug.

eapearson:
Tested and works, so maybe someone can give a quick look, and merge/rebuild CI if it looks ok?

eapearson:
@briehl

PR #891, opened 2016-12-12
Updating so that Phytozome genomes appear in other environments

samseaver:
This is a simple change, but needs to be done before PAG (mid-January)

### DIFF
--- a/kbase-extension/static/kbase/config/data_source_config.json
+++ b/kbase-extension/static/kbase/config/data_source_config.json
@@ -31,12 +31,6 @@
                 "ws": "KBaseMedia",
                 "search": false
             },
-            "plant_gnms": {
-                "name": "Plant Genomes",
-                "type": "KBaseGenomes.Genome",
-                "ws": "PlantCSGenomes",
-                "search": false
-            },
             "se_reads": {
                 "name": "RNA-seq Reads (single-end)",
                 "type": "KBaseAssembly.SingleEndLibrary",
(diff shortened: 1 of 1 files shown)


briehl:
Just verifying that there's no more "Plant Genomes" now, just "Phytozome Genomes" now? 

Also tagging @kkellerlbl - that file should just live in webroot (e.g. narrative.kbase.us/data_source_config.json), and I'm not sure how it's deployed now with the new new changes.

samseaver:
Given that the RefSeq genomes and (at a later date) the Ensembl genomes will contain plant genomes, we're better off naming them by their sources.

kkellerlbl:
Right now the file is deployed out of a config directory which is not updated by a deploy.  So we'll need to put this file in place by hand.  I'll make a note of it for Thursday.

kkellerlbl:
Will it break anything to put this file into production today?

samseaver:
No. At worst, a user who may have been expecting to see "Plant Genomes" in the drop-down box, from past experience, would now see "Phytozome Genomes" instead.

kkellerlbl:
I've put this on ci.  @samseaver can you check it out and make sure it works as you expect?  If it looks fine I'll propagate it to the other environments.

kkellerlbl:
You may need to restart your narrative container.

samseaver:
So, it's "working" in CI, but it had already been that way for CI (the changes were for next and production/appdev) so for CI, I don't see any change in behavior.

kkellerlbl:
Okay, I've just updated next.  You may need to restart your container there too.

samseaver:
Ya that worked

PR #892, opened 2016-12-13
Address 'generated_value' bug

briehl:
This fixes a couple problems with how app parameters get validated, especially those that make use of randomized input strings.

Also added more tests to that module.

Note that some of this will need to be tweaked when merging develop up to staging, but that'll be handled elsewhere.


msneddon:
Looks good.  It seems to run locally.  I never reproduced the bug, but running annotate genome registered in CI (which uses the generate output name behavior) works for me.

PR #896, opened 2016-12-14
Limit Markdown cell titles to not include newlines or paragraph symbols

briehl:
See:
https://kbase-jira.atlassian.net/browse/KBASE-4752

### DIFF
--- a/kbase-extension/static/custom/custom.js
+++ b/kbase-extension/static/custom/custom.js
@@ -596,7 +596,7 @@ define([
 //                icon
 //            ]);
         };
-        
+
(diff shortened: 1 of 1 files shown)


briehl:
Also snuck in with a fix for the /notebooks/ path being added to the URL.
see:
https://kbase-jira.atlassian.net/browse/KBASE-4013

eapearson:
looks great, running quick local test...

eapearson:
It works for me. I can't replicate the paragraph tag, but the lines and path /notebooks/ renaming, yes, and fixed.

PR #898, opened 2016-12-15
Staging tab update and disable

briehl:
1. Added a tour (mostly empty) for staging area.
2. Added yet another config file for the front end - feature-config.json - with booleans for features we want to activate. For now, includes the staging area.
3. Disabled the staging area all together.
4. Bumped version to 3.1 official.


briehl:
It will still need a redeploy for now. Once we figure out a good solution for injecting other configs in at runtime, we can lump this config in with them.

PR #902, opened 2016-12-19
Changed CI search to point to CI search API

danielolson5:


briehl:
That endpoint seems to work again. I think it was down for some time. Merging when tests complete.

PR #904, opened 2016-12-21
Taxonomy ref control

eapearson:
Changes to autocomplete input to make it work
But then abandoned that in order to use select2, since that is the current use case
That control is taxonomyRefInput
There still is a use-case for autocomplete for scientific name, so it is still in the code-base, but there isn't anything to use it now
More work is needed in the taxonomy_service module, e.g. the field_type should not be autocomplete, but taxonomyRef (at least that is my suggestion).
This should be considered unstable, experimental work, in that it currently works with the state of the taxonomy_service, which defines the only apps using it, but both taxonomy_service and the control implementation will need to change.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,7 @@
         "jquery.tipsy": "*",
         "kbase-common-js": "1.9.0",
         "kbase-service-clients-js": "3.0.0",
-        "kbase-sdk-clients-js": "0.4.1",
+        "kbase-sdk-clients-js": "0.5.0",
         "numeral": "~1.5.0",
         "plotly.js": "git://github.com/plotly/plotly.js/#1.5.1",
         "pure-uuid": "1.4.2",
--- a/kbase-extension/static/kbase/js/widgets/appWidgets2/input/autocompleteInput.js
+++ b/kbase-extension/static/kbase/js/widgets/appWidgets2/input/autocompleteInput.js
@@ -5,6 +5,8 @@ define([
     'bluebird',
     'bloodhound',
     'kb_common/html',
+    'kb_service/client/workspace',
+    'kb_sdk_clients/genericClient',
     'Taxonomy-client-api',
     'kbase-generic-client-api',
     'narrativeConfig',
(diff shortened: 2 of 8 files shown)


briehl:
If I'm getting this right, this makes a specific "taxonomyRef" input field, that now needs a "taxonomyRef" type from the app spec?

Do the apps (or, just the singular app, I think) need to be updated to support this? Or is this ready to merge and test end-to-end as is?

eapearson:
This is a little weird.

It is invoked still from the "autocomplete" field_type. However, since it is no longer autocomplete, but rather a dropdown for selecting a taxon object, the field_type should be named something like "taxonomyRef" or "taxonomy_ref". I've communicated that to Jim, and presumably the taxonomy service team.

The only apps that reference this are the taxonomy_service module. There are two apps, "Change or Assign New Taxon to a Genome" and "taxa browser" which take the taxonomy ref and thus need this new control. Those should have the scientific_name parameter switched from autocomplete to a new field_type like "taxonomyRef" or whatever.
A third app, "create private taxon", actually needs the scientific name parameter to be a text or autocomplete field rather than this new one, although it does need a new parameter (it is actually missing in the spec) for the parent taxon object, which itself should therefore be a taxonomy ref.

What happened is that the first two apps which now take a ref originally took just the scientific name (or at least that is what was implemented), and then the request to Jim was to have it pass the ref (but of course still show the scientific name), at which point it became clear that this goes beyond what an autocomplete-assisted text input does.
At the same time, there was a request to have the "create private taxon" app take a structure with both the taxon ref and the parent ref together. But when looking  at the implementation of the app, it is clear that it really needs the scientific name (text) perhaps with autocomplete assistance and separately the parent taxon.

eapearson:
I'm find letting this PR stew for a bit, or withdrawing it, until those changes are made to the taxonomy service and then updating it to accommodate those changes. But we could also merge it and get it up on CI for folks to look at.

briehl:
From @janakagithub 's comments on Slack, let's merge this in and see if the apps break, then just fix those apps on CI. Nothing here is public facing anyway.

PR #907, opened 2017-01-05
Add metadata support for editor and view cells

eapearson:
These cell types were not being added to the notebook metadata stats upon narrative save, this fix simply replicates the code to do this using the same "method." property prefix. This treats all app cell types as the same, which reflects the way they are treated in the narrative

### DIFF
--- a/src/biokbase/narrative/contents/narrativeio.py
+++ b/src/biokbase/narrative/contents/narrativeio.py
@@ -339,6 +339,18 @@ class KBaseWSManagerMixin(object):
                     commit_hash = app.get('gitCommitHash', 'unknown')
                     method_info[u'method.' + id + '/' + commit_hash] += 1
                     num_methods += 1
+                elif kbase_type == 'editor':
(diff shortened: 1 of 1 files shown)


eapearson:
I think we should just move forward with this, leaving "method" in place, and file a JIRA ticket to update "method" to "app", maybe part of a broader effort to transform method -> app within the codebases? It solves the problem of getting all app cells to show up in the dashboard narrative lists (at least after they are saved.)

It will take a coordinated change in Narrative and kbase-ui and I'm not sure I have the patience for that atm. As to whether it should reflect the different app cell types (app, editor, viewer) ... I don't think we want to reflect this in the dashboard yet, since users are probably not aware of the distinction (we don't expose this distinction, do we?). Or at least, I know Paramvir would like a refactor of the dashboard and perhaps that would be an opportunity to address that part of the issue.

briehl:
Agreed. Those are all good points. Let's make a ticket and get this merged as-is.

PR #909, opened 2017-01-06
Fix problem with certain input fields in importers

briehl:
A "select" tag should have been an "input" tag for those that don't use select2.


scanon:
Looks straight forward enough and this is very broken at the moment.

PR #910, opened 2017-01-09
Return a default type of "app" for missing app_type fields.

briehl:
App specs should now have a field called "info.app_type", as maintained in the NMS. Those that don't should try to infer what the app type is based on the presence of a few keys.

**Long version** = there are 3 possible app types - editor, viewer, and app. "Viewers" don't have their execution behavior defined in the spec (that's all managed by their widget) and are easy to infer. "Editors" are expected to be explicitly called out by the app_type field, since the cell change is mainly one of usability and how the code is executed. "Apps" are, well, all the rest of the apps.

Currently, in production, there are no Editors (e.g. the Reads Set Editor in CI), so defaulting to App won't break anything. This just means that before registering Editor apps in prod/appdev, we need to update NMS and Narrative. Once NMS is updated to generate that key on app registration, it will all work itself out.

**Short version** = If no app_type key is found, and the spec defines how code is run, it's treated as an app cell.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,7 @@ __Changes__
 - Release of support for data palettes (currently disabled in the service)
 - Data now gets fetched from a Narrative Service to support data palettes.
 - Support for an FTP-based data file staging area (currently disabled until import apps catch up).
+- Fixed issue where an undefined app_type field would cause a crash while instantiating an app cell.
 
 ### Version 3.1.0-alpha-4
 __Changes__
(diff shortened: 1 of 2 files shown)


eapearson:
Looks good. That throw looks like my handiwork. One day It may be just a switch on app_type -- period, no grokking the rest of the spec -- with a default that throws, and assume the NMS always populates that property, and has validated the rest of the spec vis-a-vis that app_type.

briehl:
That's pretty much what it was. It did already default to "view" if there was no execution behavior, but it expected app_type to be there if there was a behavior.

Now it just defaults to app and does a warn instead of a full throw.

briehl:
Also, FWIW, this is the only place in the current code that depends on a value for app_type being present.

rsutormin:
Looks good.

PR #913, opened 2017-01-10
Job polling fixes

briehl:
The way the Narrative polls for jobs is a little unoptimized, and this will contain a few fixes for that. I just want to write down some notes on the problem and some solutions.

**Problem**:  
Job lookup polling is spamming NJS and possibly causing slowdowns. Even if not, it's spammy and shouldn't be. There's also a bug that jobs that have entered a terminal state (completed, errored, cancelled) are still being looked up, which isn't helping.

The flow generally goes like this, split between the browser and kernel.
1. User clicks "run" to start a job.
2. Kernel starts the job, keeps job info and puts it on a list of running jobs.
3. Kernel has a thread that, every 10 seconds, looks up status of all jobs. That info is pushed to the browser (kbaseNarrativeJobsPanel widget), which doles it out to each app cell.
4. When a cell perceives that a job has terminated, it should tell the jobs widget not to look up that info anymore.
5. Jobs widget passes that message back to the kernel.

However, if it's long, the user can shutdown, restart, or refresh the page. When that happens, things reload like this:
1. Kernel asks UJS for a list of jobs for that Narrative.
2. Kernel starts the job lookup loop, fetching status for all jobs, just like before.

Because of the inherent asynchrony here, there are some problems.
1. The app cells don't always get the status at an expected time. Also, when they start up already in a terminal state, they don't even listen for the status.
2. App cells and the browser's job manager don't have to start up in an expected order. A Narrative filled with app cells might have a few initialize first, then the app manager, then the rest. Which leads to...
3. A Narrative being restarted with a job that's now in a finished state needs to communicate that change to the app cell.

All of that is a long way of saying that the app cell should be the point that determines when to stop looking up job info. As far as I can tell, there are a couple of current bugs with that.
1. App cells don't communicate when they don't want to hear job messages properly.
2. Restarting a narrative with completed jobs will continually look up their info.
3. Job info doesn't get deleted, so we can maintain provenance. App cells CAN be deleted, however, so that place that says "Don't look up that job anymore!" should be assumed to go away.

I propose a couple solutions.
1. Fix the bug that makes app cells shut down lookup when a job finishes (**done**)
2. Create a check for an app already in a finished state at page startup, and prevent lookup of its job (might take some trickery to ensure that the jobs widget and completed app cell are both alive at the same time).
3. Add some heuristic where the job widget will stop looking up a job when it sees it in the same terminal status X many times (say, 3? Note, this will overlap with solution 2 above).
4. When the kernel gets a job status that's a terminal state, it's not going to change, so just cache it and always return it when asked.
5. Incorporate batch job lookup in the kernel.

These should reduce the redundant job lookup burden on NJS. I'm leaning toward doing these in this order: 1 -> 4 -> 5 -> 3 -> 2. Honestly, I think caching job status that's already in a terminal state will solve 99% of the practical problems, but the rest of the changes should be done anyway.


briehl:
Update:
* Added fix from app cells that should stop polling when it sees a job finish.
* Added kernel-side cache of status for jobs where "finished" = 1 in last known job state.
* Added change to kernel that will do a batch lookup of jobs with indeterminate status.
* Routed all kernel-side job status lookup through the cache.

One fun point of failure is that if a batch job lookup fails, that pass fails completely and throws an error. But at this stage, I think it's probably ready to merge.

To be clear, this is fixes 1, 4, and 5 from above. I'd rather not spend time implementing 2 and 3 unless really really necessary. And if the optimization goal here is to prevent spam to njs, the fixes in this PR should solve it.

briehl:
Any complaints against merging this and rebuilding CI?

PR #918, opened 2017-01-13
filters out empty sections

thomasoniii:
Only adds in the Summary or Report sections if there is actual content in them; changes the text of the placeholder in Links


briehl:
👍

PR #920, opened 2017-01-14
Removed more job poll spam, migrated some older tests

briehl:
This looks bigger than it is - that's mostly due to migrating some older tests from `biokbase/narrative/common/tests` to the more complete test dir in `biokbase/narrative/tests`

The main change here is to the jobmanager. Previously (thanks for the help @scanon) it was polling all jobs for their parameters, even when completed. Now, the list of parameters are fetched only once at kernel start, stored in each individual Job object, and returned from there on request. This should cut down the rest of the NJS spam.

In fact, if a Narrative is just full of completed (errored, suspended, cancelled, and otherwise) jobs, it should **never** talk to NJS for job info, except on restart. And even that should be a batch call.


kkellerlbl:
Should we deploy this to CI?

briehl:
Yes. I'm pretty sure I updated CI with this patch. If not, I'll do it in a sec.

PR #922, opened 2017-01-17
Develop -> Staging for release on Next

briehl:


kbase-extension/static/kbase/config/feature-config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "stagingDataViewer": false,
+    "stagingDataViewer": true,
```
kkellerlbl: Is this okay?  For next I'm guessing yes, but I want to be sure.  What about appdev and prod?

kkellerlbl: We should probably manage this file using a Docker volume mount in each environment anyway.  So I'm merging.

briehl: Oops. Thanks for catching that.
It should probably be turned off until we're ready to start supporting it more (hopefully later this week).

PR #923, opened 2017-01-18
View-only fixes

briehl:
This fixes a couple problems I found.
1. Using the copy button in read only narratives (e.g. public ones) worked, but created a bad URL in it's "Open the new narrative" button.
2. Disabled the execution controls for apps, editors, and view cells. Clicking on a data object should still create a (temporary) viewer, but a user can no longer run or modify any cell using the main action button.

@eapearson - I just added 'display:none' to the main cell areas when they get created. There's still plenty of internal logic that expects those buttons to be there for binding things, but just hiding their containers from the user seems to be sufficient.

see also https://kbase-jira.atlassian.net/browse/TASK-312

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,11 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.2.1 (more notes will follow).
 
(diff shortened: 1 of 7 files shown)


briehl:
Gonna just merge this to get it up on CI. We can fix things there.

PR #925, opened 2017-01-18
Fix subdata control, code org

eapearson:
- fix subdata control synchronization within grouped parameters (struct input) and multiple value (sequence input).
- remove a set of obsolete files (were causing false positives when looking for code to refactor)


briehl:
The zombie submodules that wouldn't die!
Someone call George Romero.

eapearson:
it was because I had built it before catching my develop branch up with upstream, and then rebuilt it afterwards, but some things didn't go away, like bower modules (which I noticed) and submodules (which I didn't)  :(

PR #926, opened 2017-01-20
fixed glitch in slider; selection button

thomasoniii:
The volcano plot had troubles.

* The slider values weren't pre-populating
* the show genes button didn't work for big data.

Should be all better now.


briehl:
👍 thanks!

PR #931, opened 2017-01-24
Report view fixes, job update fixes

eapearson:
- This includes the set of changes to the report view:
  - interaction with direct html report (resize, pass through clicks)
  - use direct html index and link for iframe rather than direct html string
  - omit html report section if no report
  - summary is not iframe, just a div
  - omit link sections if no links
- this change includes iframe communication modules ported from bulk-ui, although they are not used much in the end since they rely on javascript injected into the iframe and that is not always easy (or possible) without assistance from the report itself

- job update fixes:
- yesterdays job event flow fixes were flawed, I don't know how they tested out ok but in additional testing last night started to fail again. New fixes focus on ensuring that the job update loop closes when there are no jobs being monitored, and that when the app cell calls the correct method to start monitoring a job that the job monitoring loop on the back end starts up.


eapearson:
BTW there is a handy app for testing out report generation and display. eapearson_TestRichReports (dev in ci), lets you generate reports with most of the fields (direct html, summary, html links, html link report index, html height, summary height).

kbase-extension/static/kbase/js/widgets/function_output/kbaseReportView.js
```diff
@@ -75,48 +82,49 @@ define(
...
-          }
+        // this is an ugly hack. It'd be prettier to hand in just the shock node ID, but I don't have one of those yet.
+        // Also, this is embedding the token into the html and the URL, both of which are potentially security concerns.
+        // TODO: update to put the auth token into the url... should be working in CI already.
+        // TODO: NO HARDCODING OF URLS!!!
+        importExportLink: function (shock_url, name) {
+            var m = shock_url.match(/\/node\/(.+)$/);
+            if (m) {
+                var shock_id = m[1];
+                var url = 'https://ci.kbase.us/services/data_import_export/download?&id=' + shock_id + '&token=' + this.authToken() + '&wszip=0&name=' + name;
```
briehl: This should be a configure option in src/config.json. At least the URL - the form params should just stay.

kbase-extension/static/kbase/js/widgets/function_output/kbaseReportView.js
```diff
@@ -75,48 +82,49 @@ define(
...
 
 
-        properPreauthURL : function(url) {
-          var m;
-          if (m = url.match(new RegExp('^http://ci.kbase.us/preauth/(.+)$'))) {
-            url = 'https://ci.kbase.us/services/shock-api/preauth/' + m[1];
-          }
-          return url;
+        properPreauthURL: function (url) {
+            var m = url.match(new RegExp('^http://ci.kbase.us/preauth/(.+)$'));
```
briehl: Same with these URLs

kbase-extension/static/kbase/js/widgets/function_output/kbaseReportView.js
```diff
@@ -126,397 +134,796 @@ define(
...
+                iframeOrigin = document.location.origin,
+                iframeMessages = HostMessages.makeHost({
+                    root: window,
+                    name: 'panel'
+                });
+
+            // The iframe content needs requirejs amd.
+
+            var requireConfig = {
+                    baseUrl: 'http://localhost:8888/narrative/static/',
```
briehl: This should also be a configuration - it probably won't work anywhere but locally.

PR #932, opened 2017-01-24
Added an appdev stanza to data_source_config

briehl:
The Public and Example tabs weren't working.

I'm not exactly sure what's supposed to be there, so I just copied the sources from prod. This changes lets Public Phytozome Genomes and Media work (the rest don't work for some reason... not sure how they're supposed to be configured, or if there's a missing Search service in appdev). Also, in the Example workspace (KBaseExampleData), there's only a few data sets.

So that should probably be either reconfigured to point to valid workspaces, or have those workspaces filled up with appropriate testing data.

Either way, this is better than an error message.

### DIFF
--- a/kbase-extension/static/kbase/config/data_source_config.json
+++ b/kbase-extension/static/kbase/config/data_source_config.json
@@ -328,5 +328,107 @@
                 }
             ]
         }
+    },
+    "appdev": {
+        "publicData": {
+            "genomes": {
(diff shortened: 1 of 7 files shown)


briehl:
Gonna just merge this, since it fixes another issue where some old test print statements were creeping into code cell outputs. It won't have any other side effects outside of appdev.

PR #949, opened 2017-01-27
refactors buildObjectIdentity into super class

thomasoniii:
The sole purpose of this pull request is to cause merge pain for Bill. Well, okay, not sole purpose. Primary, probably.

Anyway, I noticed that we've had a proliferation of buildObjectIdentity methods in the system that are all identical. So this just refactors them and rips out the common method from the individual widgets and puts it up into the superclass in kbwidget. I know that we have other redundant methods, but this is the one that jumped out at me.

This has not been thoroughly tested - I made sure that a few narratives I had spun up, but I didn't actually have anything to test the actual affected widgets. See first comment about causing merge pain for Bill.

That said, it *shouldn't* cause any problems, since it's just refactoring a redundant method into a single definition in a common superclass. No functionality should change.

This is strictly a technical debt pull request, and can be integrated at leisure. No functionality changes (AFAIK), no new features are added. So I'll defer as to when is the most opportune time to include it, if at all.

### DIFF
--- a/kbase-extension/static/kbase/js/kbwidget.js
+++ b/kbase-extension/static/kbase/js/kbwidget.js
@@ -923,6 +923,28 @@ define (
                 return 'uuid-' + result;
             },
 
+            buildObjectIdentity: function(workspaceID, objectID, objectVer, wsRef) {
+                var obj = {};
+                if (wsRef) {
+                  obj['ref'] = wsRef;
+                } else {
+                  if (/^\d+$/.exec(workspaceID))
+                    obj['wsid'] = workspaceID;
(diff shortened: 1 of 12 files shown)


briehl:
I think this is a great idea. A lot of widgets out there are filled with duplicate code. There is a util directory full of things, too, (StringUtil has a uuid function, for instance) that should just be included as dependencies.

Putting this in the root kbwidget is fine as well, since all of our current viewers make use of that.

But I'm gonna hold off on merging for a while, since it touches a number of viewers that should all be pretty thoroughly tested.

PR #954, opened 2017-01-30
sample_ids can now be names or refs

thomasoniii:
Should work, but I couldn't test due to workspace permission issues.

Just looks at a sample id to see if it has a / in it. If so, now assumes it's a ref. If not, assumes it was a name, as before.


briehl:
Looks reasonable. I'll get this deployed in CI right away.

PR #960, opened 2017-02-01
enable output objects in report viewer

eapearson:
- the flag was not being set in the app cell
- the object list link was also a bar not a pointer
- ui needs work -- probably should be a button, not link, and scroll to the object inserted in the narrative.


eapearson:
Fixes https://kbase-jira.atlassian.net/browse/TASK-436

PR #965, opened 2017-02-01
lua changes from next

kkellerlbl:
Commit changes currently deployed on next.


briehl:
Is this the change that would fix all containers being lost on redeploy? Or is it just catching up to non version-tracked things?

scanon:
I think this is mainly the lost containers.

kkellerlbl:
It's actually both.  These files were on next03 but nowhere else, and the lost narrative connection fix is in there (the notelauncher.lua changes).

briehl:
Oh! I see what it's doing now. Yeah, that's great. Thanks for the fix.

PR #967, opened 2017-02-01
Fix advanced sub controls rendering as blank spaces

eapearson:
Don't propagate advanced flag to subcontrols because doing so makes then un-renderable, or rather, the class added to advanced subcontrols won't be removed.
Fixes https://kbase-jira.atlassian.net/browse/TASK-439


briehl:
This is fine. I think we're still ripe for a design review of what "advanced" parameters really means, especially with subcontrols. But let's just keep them working :)

PR #969, opened 2017-02-01
Bumped version to 3.1.8, added release notes.

briehl:


briehl:
skipping build, since there's no new code.

PR #978, opened 2017-02-03
parameter group field order same as first param (not just at end of parameters)

eapearson:
Previously, a parameter group is laid out in the parameter groups at the end of the parameters. This denies the app developer the ability to place the parameter group in a specific position amongst the parameters. Although this is a shortcoming of the app spec, we can accommodate this need with a simple hack -- the position of the first parameter within a param group determines the position of the param group.

- this change sets position of the parameter group as that of the first sub-parameter contained within.
- error dialog placed at the cell setup phase
- alert replaced with error dialog at cell insertion
- todo: info dialog containing error info should be replaced with a nicer error dialog


scanon:
Does this break anything currently?  If not, I would like to hold off on this until after the contractors meeting.

eapearson:
Yes, that is absolutely fine.

briehl:
I think we can get this on CI for testing now. @eapearson, mind fixing the conflict?

eapearson:
Fixed.

PR #982, opened 2017-02-03
Fixed broken minified import tab. Some related fixes, too.

briehl:
The parameter fields for the import tab were being set up in an unsupported way, and the minifier really exposed that.

There were other problems exposed by digging into this - a dropdown parameter was always treated as valid, even if it wasn't (e.g. the technology type dropdown for reads was allowed to be empty, even though it was marked required, which would cause the import to fail.)

There's still some interesting UX stuff going on if a text field is optional (it can be pre-filled, and still have a check mark when emptied), but those might be fixable with an improved docstring.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,13 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.2.1 (more notes will follow).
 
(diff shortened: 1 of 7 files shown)


briehl:
This also fixes a slew of console error messages that happen when the importers are first rendered. They seem benign, but were due to things not being constructed at the right time, so could have led to bad side effects.

PR #985, opened 2017-02-03
Minor style hack^H^H^H^H fix to data slideout in Safari

briehl:


briehl:
This might apply to some implementations of Firefox, too. It puts in a background in case there's some word wrapping in the menu/tabs bar of the data slideout.

PR #992, opened 2017-02-08
Improve app cell status line message and clock

eapearson:
The goal was to make it more compact  and succinct.
Highlights:
- don't show seconds after 1 hour
- show finished time as an elapsed-time clock if less than 1 day, otherwise show the timestamp. This makes the finished time, especially, more compact. Other than aesthetic, it enables the app cell to be more readable in a narrower browser window


eapearson:
Addresses https://kbase-jira.atlassian.net/browse/KBASE-4986

PR #994, opened 2017-02-09
change "copy" to "copy..." in view only header bar

briehl:


briehl:
Very low impact, so I'll just merge.

PR #996, opened 2017-02-10
Fix copying workflow in view only mode

briehl:
See https://kbase-jira.atlassian.net/browse/KBASE-5008

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeWorkspace.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeWorkspace.js
@@ -373,8 +373,8 @@ define([
                 body: $copyModalBody,
                 closeButton: true,
                 buttons: [
-                    $doCopyBtn,
-                    $cancelBtn
+                    $cancelBtn,
+                    $doCopyBtn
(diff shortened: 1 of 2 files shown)


briehl:
Any complaints if I merge this? @eapearson want to review?

eapearson:
I've noticed more Handlebars templates showing up. Is that the direction we're moving in for markup in code or just a personal preference? It is definitely an improvement over markup through jQuery, at some cost elsewhere.

briehl:
For the record, I know there are a ton of libraries out there we could use to handle templating. Nunjucks is/was in there for a while on the kbase-ui side. Handlebars was in there, as well, for other uses, so I've been sticking with it in the Narrative. I would like to move away from plopping HTML code directly in Javascript, though, however we do it. There will be an extra cost of event binding, etc., but I feel it's something we should try for going forward.

briehl:
Ok. I'll merge and rebuild on Jenkins.

I would also like to explore using React in the narrative, since it's another interesting solution to the problem, and it would also (probably) serve 3rd party devs who want to make widgets with something more modern than pure ES 5. Web components are a good idea, too.

Still, I think we should continue this discussion elsewhere.

PR #997, opened 2017-02-11
Bump to version 3.1.11, release notes.

briehl:


briehl:
This is just a teeny version bump. Merging now so I can redeploy CI.

PR #1000, opened 2017-02-14
Upgrade to Jupyter 4.4.1

briehl:
Might as well go for the most recent.

Also a note, Jupyter 5.0 is going into beta (today, I think), and a release won't be far out after that. I haven't looked into the laundry list of changes they've made, but I think this is the one that comes with a mostly new codebase, and is intended to be the launching point for Jupyterlab. So we have that work to plan around.


scanon:
Do you want to change the Dockerfile to use narrbase:4.6 at the same time?  Or do you want to do that separately and just use this to get narrbase 4.6 bootstrapped?

briehl:
Oops. I always forget something...

briehl:
I think we should merge this soon. Other dependencies (in the local installation) are bringing in the updated version, which is causing problems that this fixes.

kkellerlbl:
We're going to merge against the auth2 branch instead, to make sure that we don't do this till we're really ready.

PR #1007, opened 2017-02-28
Changes to support auth2.

briehl:
This is a first pass so far. Lots of small file changes, but here's the main gist.

* added an api/auth.js to talk to the auth server (returns Promises for its calls)
* narrativeLogin.js handles all login stuff, mostly as before
* kbaseLoginFuncSite is obsoleted - the previously used events (loggedInQuery.kbase, logout.kbase, etc) are still there, but set up through narrativeLogin. Logging in is done elsewhere, and there's no more cookie parsing, so that module is no longer needed. kbaseAuthenticatedWidget (and its children) all still work as before.
* built a new module to deal with the logout menu in the upper-right corner - userMenu.js. Should now look and feel like the dashboard.
* rewrote kernel-side support for authentication, removed nexus libraries.
* rewrote the server-side Lua narrative manager to deal with auth2. Also decomposed the auth management stuff into a separate module - kbase_auth.lua


briehl:
I think this is ready to go now.

briehl:
Fixed some sharing issues and caught up with develop.

PR #1023, opened 2017-03-16
Fix problem with building narrbase image

briehl:
Installing IPywidgets brings in the latest setuptools (or some recent version) which is incompatible with other things, and effectively breaks our Python install.

This sticks to the last good version of setuptools (33.1.1) and should build.

See also this issue:
https://github.com/pypa/setuptools/issues/942


briehl:
Wrong branch, dummy!

PR #1027, opened 2017-03-17
Getting develop up to date with changes to staging

briehl:
Also bumps version to 3.2.0


briehl:
The local installer was having version issues with Jupyter components, so I just brought in the other version changes that are in the auth2 branch.

PR #1033, opened 2017-03-22
Improve job log viewer behavior

eapearson:
- Was requiring user click the play button to get the log to appear, counterintuitive
- Also the "next" button was not working correctly
- Now initializes the job status and log, and starts polling if necessary
- Also refactor job log buffer code into separate internal module


eapearson:
Addresses https://kbase-jira.atlassian.net/browse/TASK-588

PR #1040, opened 2017-03-28
Add agent token and update tests

briehl:
This does two things - it adds the creation of an "agent token" (described below) and updates the kernel-side tests to deal with changes to auth2. It also has an update to the makefile so that `make test` should work end to end now.

### Test Updates
This has a scary number of files touched - most of them are parts of the test harness. The main changes there are to the configuration to make sure that different data file paths are correct and relative to narrative root. It also tweaks how tokens are dealt with - previously it required someone to provide a user name and password, now it can take a dev token provided to a couple of files (not included in this PR). The next thing to do will be to embed them in encrypted environment variables through Travis-CI.

### Agent Tokens
One problem we need to address is token lifetimes. If a user has a token with a remaining lifetime of ~10 minutes, and starts a job that will last an hour or more, that job will ultimately fail because it will lose authorization to act for the user.

One way around this is to create an "agent" token for each job that starts. This will act as the user's agent, and have a somewhat longer lifetime (1 week right now, can be configured in the Auth service). Here's the general flow of the changes.

1. User hits "Run" on an app.
2. AppManager in the kernel prepares the app to run, validating parameters, etc.
3. AppManager then uses the user's login token to request a new agent token (named for the app being run).
4. AppManager creates a new NJS client with the agent token and starts the job as usual. It adds the token's id to the job metadata.
5. The new Job object owned by the JobManager also contains the token id as well.
6. NJS runs the job as usual with a token that will hopefully last long enough.


eapearson:
And a user might simply log out, thereby revoking their token.
I've following this tangentially, but haven't followed the issue in detail. It is a bit worrying that we are considering cutting the agent token lifetime so close to long job queue + run times. I know the danger is that if a token escapes the clutches of kbase it provides authorization to do agenty things -- but it does seem that being internally produced and communicated, at least in terms of machines, and communicated across https, keeps them pretty well buttoned up. And the cost of a long-running job dying due to token expiration is very high!
So what about long lived (e.g. double the longest imaginable job duration) agent tokens in combination with conscientious agent-token management? E.g. agent tokens must be revoked as soon as the job is done, errored, or cancelled or used in testing, and wayward tokens simply tracked down and revoked as part of auth management.
Of course, that is out of scope for this update, and it is indeed configurable in the auth2 service -- I was just triggered by the "hopefully will last long enough." bit!

briehl:
Gavin and I talked a little bit about revoking tokens on job completion. It gets tricky and can have interesting effects on the auth server. The obvious way to go about it is this:

1. User runs a job.
2. Kernel gets agent token.
3. JobManager marks job as finished
4. JobManager tells auth service to revoke token.

The problem comes with asynchrony, and being sure that a token has been revoked. That same flow as above applies whether you shut down and restart your narrative between steps 2 and 3 - it would still act the same. But what happens next time the narrative starts? The job's already finished, but there's no way to know if the token's been revoked, so might as well make another call to the auth service. This can get really ugly if you have dozens or more jobs in a narrative. Also, this can be a problem with sharing - one user does not have permission to revoke a token made on behalf of another user. So now there's an extra introspection step for the tokens - gotta make sure that the current user has permission to revoke it.

This could be solved by tracking token ids elsewhere, either as part of the narrative object or ws metadata, then that can cause issues, too - how long do we track the ids, who can see them, what other data do we put with them, etc.

I guess the best way to deal with all of this asynchrony is to have NJS revoke the token - that's the only real authority that knows that a job is done as soon as it finishes. Maybe that's something we can discuss later?

I agree with what you said in there, though - since these tokens are only ever used internally, and only the token ids are exposed (and even then, only to a reasonably advanced user), there's not much danger in just letting them expire.

scanon:
I think for now we should just let them expire.

briehl:
Should we go ahead and merge this, and put it up on CI then?

eapearson:
yes, i'd say go for it, seems to be working fine for me.

PR #1043, opened 2017-03-30
add more staging uploaders

Tianhao-Gu:


Tianhao-Gu:
@briehl Could you review when you have a minute? Thanks!

PR #1046, opened 2017-04-11
Narrative loading changes [wip]

eapearson:
- do not merge as-is
- for bill to pickup or not, untested (since the original work months ago).
- the diff with develop is just about the same


briehl:
This seems safe enough, with a little extra polishing. I'll merge this now as-is, then make edits today before it gets deployed.

PR #1047, opened 2017-04-12
Update proxy_mgr for new interstitial loading page.

briehl:
Now it goes to load-narrative.html
Also, make sure that when there's an UNAUTHORIZED error, we return it as such and don't try to let users through anonymously.


briehl:
Note that this needs this PR in kbase-ui to work:
https://github.com/kbase/kbase-ui/pull/284

PR #1056, opened 2017-04-18
Small update to get pandas working again

sychan:
Pandas loads properly when I just tried with local docker against AppDev


sychan:
Going to re-submit to check github/jira integration

PR #1060, opened 2017-04-20
Make install logs appear in console, and a bit more verbose

briehl:


briehl:
There. Finally got the install tests fixed.

PR #1065, opened 2017-04-25
develop -> staging to release 3.2.5

briehl:
contains:
* fix for logs not able to scroll when there's an error
* adds binned contigs viewer and supporting modules 
* adds to and renames a few tests


briehl:
The single failure in a test is a red herring - there are apparently still issues with the webdriver. That doesn't actually test anything anyway...

PR #1079, opened 2017-05-08
RESKE search demo

psnovichkov:
widget for the demo app from this repository:
https://github.com/kbaseapps/RESKESearchDemo (available for the dev mode in CI)


kbase-extension/static/kbase/js/widgets/function_output/kbaseRESKESearchResultDemo.js
```diff
@@ -0,0 +1,535 @@
...
+                data_includes: [] 
+            }
+        },
+        searchObjectsOutput: {
+            pagination :{
+                start: 21,
+                count: 40        
+            },
+            sorting_rules: [],
+            objects: [
```
briehl: A lot of this stuff seems hard-coded. Is this just for the demo for now? Should it move to a config file?

kbase-extension/static/kbase/js/widgets/function_output/kbaseRESKESearchResultDemo.js
```diff
@@ -0,0 +1,535 @@
...
+            this.$messagePane.hide();
+            this.$messagePane.empty();
+        },
+
+        clientError: function(error){
+            this.loading(false);
+            this.showMessage(error.error.error);
+        },        
+
+        escapeHtml:  function(string) {
```
briehl: This seems useful enough to move to an API call, probably in util/string.js. But we can do that later.

kbase-extension/static/kbase/js/widgets/function_output/kbaseRESKESearchResultDemo.js
```diff
@@ -0,0 +1,535 @@
+ define (
+	[
+		'kbwidget',
+		'bootstrap',
+		'jquery',
+		'plotly',
```
briehl: It doesn't look like Plotly is being used in this widget. Should be removed as a dependency.

briehl: Same with datatables and kbaseTabs

briehl:
The failed tests is due to a dependency mismatch that isn't a problem in deployments, only on Travis for some reason.

PR #1078, opened 2017-05-05
Catch up with auth2 branch, update installer tests

briehl:
Add the minify step to the build-travis-narrative makefile rule.

PR #1085, opened 2017-05-09
Fix kbase common dependency

eapearson:
Root of the problem was a transitive dependency -- kbase-ui-widget was a dep. of kbase-service-clients-js but was not being used. It had an updated dependency on kbase-common-js ^2.5.0, so this was causing a conflict with 1.9.0 in the narrative bower.json. The widget library should have been bumped to 2.x when this change occurred. In any case, the fix was to move to kbase-common-js ^ 2.5.0 (which currently resolves to 2.7.1), and at the same time fix the md5 library change (it is another transitive dependency, and the upstream source changed the installation directory). The kbase upstream libs were updated to remove dependencies, since they didn't really use them.

Tested locally, but CI is experienced problems with auth2 atm, so it could use more eyes. Basically, if it builds it should work. The only breaking changes for kbase-common-js was the base directory and the usage of a single requirejs root directory, which had already been in effect in the Narrative.


briehl:
Thanks for fixing this. I wasn't sure of the chain of dependencies that led to kbase-ui-widget and its components, or even if it was in use.

I had disabled the Narrative backend (Python biokbase.narrative) testing for just that reason, especially on the develop branch. I'm not sure why it's trying to run again here, or maybe that's just on the other PR #1078 . Let's merge this, then I'll work through other parts of the test harness.

eapearson:
In post-mortem, it could have been solved in two ways:
- banish all transitive dependencies (I don't think there is a bower option for this, but one can inspect the bower_components directory to see what creeps in, and of course the requirejs path to module mapping is the entry point for all modules)
- put in an exception to force a version for a misbehaving dependency, but this may break other things.

PR #1092, opened 2017-05-31
clusters in tabs, not bounching through methods [TASK-430]

thomasoniii:
This alters the 3 View methods in the "Explore Cluster" drop down to pop up in tabs.

"Build Feature Set" is a different type of method that needs to be run and produce output, so it's just hidden for now until there's a good method to invoke it.


briehl:
Looks good!

PR #1100, opened 2017-06-13
Updating prod section for example Assemblies

samseaver:


kkellerlbl:
Should we merge despite the test failure?  I think @briehl mentioned you were fighting the test suite still.

briehl:
Yes, this is failing for the same reasons I've (more or less) fixed in my PR. Let's go ahead and merge it all.

PR #1101, opened 2017-06-13
develop -> auth2 for CI build

briehl:
CI still builds off of auth2, so these changes need to be merged there.


kkellerlbl:
Should I just put CI back on develop for narrative?

briehl:
Eventually, yeah. auth2 needs to get merged -> develop first, since it has newer changes (like, all the auth2 support). I want to get that sorted out after this meeting today.

PR #1108, opened 2017-06-16
TASK-588 - issues with job status widget

briehl:
Address the following issues with the job status widget (that's the cell that appears for upload jobs)

The real fix is putting in an actual app cell, whose machinery works well. This is just a stand in for a real app cell - it's the Javascript representation of the Job object returned by the biokbase.narrative.jobs.jobmanager.JobManager.

Anyway.

This fixes two things.
1. The constant chatter and spam to the kernel. Now, it should only try to update its status every 5 seconds.
2. The job logs not auto starting as expected. Still untested with jobs that fail fast, but jobs that proceed normally should be fine.


briehl:
Just tested with failing jobs. It seems to behave. I'll put it on CI for external testing.

PR #1112, opened 2017-06-21
add dist-upgrade to narrbase image build (TASK-861)

kkellerlbl:
Adding specifically for the stackguard bug (TASK-858) but should be helpful for keeping the image OS up to date in general.


briehl:
Hm. Should this be in narrbase or narrprereq? narrprereq is the one closest to the metal, so to speak.

kkellerlbl:
Ha, an upgrade was already in narrprereq!  I changed it to be a dist-upgrade, which has some minor differences but should be no problem.

briehl:
great. thanks!

PR #1115, opened 2017-06-22
Create dockerhub_builds.md

briehl:
Documentation on using Dockerhub to build KBase images.

### DIFF
--- a/docs/install/dockerhub_builds.md
+++ b/docs/install/dockerhub_builds.md
@@ -1,0 +1,34 @@
+## Building Narrative Images with Dockerhub
+(Updated 6/22/2017)
+
+We currently make use of Jenkins and Dockerhub (https://hub.docker.com) to manage Narrative production builds. There are three images here that stack on each other.
+
+### List of images
+1. **narrprereq**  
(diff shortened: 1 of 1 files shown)


docs/install/dockerhub_builds.md
```diff
@@ -0,0 +1,34 @@
...
+git tag narrbase1.2
+git push --tags
+```
+2. With the new tag in place, head to Dockerhub. In the KBase organization, click the `kbase/narrbase` block. Navigate to the `Build Settings` tab to see the list of builds available. The top one should have a green '+' sign - click that to add a new row.
+3. In the new Build Settings row, select type=tag in the dropdown, and set the name to the name of the git tag you created (above would be narrbase1.2). Set the location to the directory where the Dockerfile is (relative to the root of the repo), and add a semantic version for the Docker image.
+4. Once that's done, hit "Save Settings." Then you can trigger the build with the Trigger button.
+
+Note that the Dockerfile of narrbase and narrative depend on specific versions of previous images being present, so watch your build order.
+
+With the two base images built, you can now use the internal Jenkins instance to build and deploy the main image, as it pulls the base images from Dockerhub. Instructions for that are elsewhere.
```
kkellerlbl: Maybe we should wait to add this line until it's accurate.  :)

PR #1116, opened 2017-06-23
Create define_env_vars.sh

kkellerlbl:
New (untested) file intended to be sourced inside shell scripts to define Docker image names and versions in the environment.


briehl:
👍

PR #1118, opened 2017-06-27
Develop->master to deploy on next

kkellerlbl:
Also to be able to pull data_source_config.json to appdev and prod.


kkellerlbl:
@briehl does this look reasonable?  If not I can always get data_source_config.json from the develop branch.

kkellerlbl:
I'm planning on pushing up the existing narrative image on next before rebuilding, so this PR wouldn't get pushed up to production.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeManagePanel.js
```diff
@@ -34,9 +34,9 @@ define([
 ) {
     'use strict';
     return new KBWidget({
-        name: "kbaseNarrativeManagePanel",
+        name: 'kbaseNarrativeManagePanel',
```
kkellerlbl: This is the only diff that concerns me; there's a lot in here.  But if it seems okay let's just merge it.

briehl: That was where all the work was done to fix the bug where the wrong narrative was being copied. Also, some minor refactor and cleanup. It should be safe - there have been no complaints from CI or Next.

PR #1119, opened 2017-06-28
TASK-877 new versions of apps with a new name caused errors

briehl:
This case should just make a "version out of date" warning, but prompt with the newer name for the new app to click to update, instead of causing the app cell to crash all together.


briehl:
Agreed about the module_name check, but like you said, it should never fail. Anyway, I'll bump the version, then let's merge and get it on CI.

briehl:
most recent changes were just release notes, don't need more testing.

PR #1120, opened 2017-07-05
bug fixes to kbaseExpressionSampleTableNew.js

thomasoniii:
I don't have an associated TASK ID - these were per request of Arfath who encountered a few bugs while using the ExpressionSampleTableNew widget.

* Now reports an error message if loading the sub-objects also fails
* Only presents a single TPM tab, not constantly adding them


briehl:
👍

PR #1121, opened 2017-07-06
another bug fix to ExpressionSampleTableNew

thomasoniii:
Speaking poorly to our QA work, mostly mine, Arfath discovered yet another bug.

Choosing different expression levels would merely append more rows to the overview table, not redraw it.

Fixed. Also updated to the current data tables api in the process.


briehl:
great!

PR #1122, opened 2017-07-13
Update type listing in data slide out panel, TASK-141

msneddon:
This definitely gives a better listing of types (for the my data and shared with me tabs in the data slideout), but I may still be missing some. The list we had was way out of date.

I updated the reads types to be the new KBaseFile versions. Right now the selection requires that we pick one or the other (which is bad, since types will change over time), but changing that to select multiple types under the same name would be much more work that probably isn't worth it right now.

We at some point could switch to listing types based on what is available in the list, similar to what is done in the data panel. That's also tricky at this point though because I'm not sure we populate the entire list from the WS right away, so we may not have the complete list of types.

We should also update this list once we determine what all the new RNAseq types will be.

At some point, these panels would be replaced by search which may both speed and simplify things a bit.


briehl:
_At some point, these panels would be replaced by search which may both speed and simplify things a bit._

I can't wait! With RESKE mostly there, this might be worth prototyping sooner, rather than later. Or at least trying to patch in the work that @eapearson is doing with the new Search interface.

briehl:
But let's merge this for now to get the changes in.

briehl:
Also, we *don't* populate the whole list right away, especially for cases where there's a ton of items, which is why we can't populate that dropdown. It might be possible for the NarrativeService to do some kind of variant of a get_objects call to just return the list of available types, but that might also put too much pressure on the Workspace.

This has been the case for a long time. I feel like it's worth waiting on the RESKE stuff to mature and just refactor that whole slideout to use that as the search engine.

PR #1123, opened 2017-07-13
new volcano plot; same as the old volcano plot

thomasoniii:
Lots of refactoring work on the volcano widget, inevitably with more to come. But you gotta draw a line somewhere.

The final result is that it's a drop in replacement and it should work pretty much identically to the old one. The only appearance change is (temporarily?) disabling the popups on the genes, since they don't currently work. That needs more investigation.
There may be minor appearance tweaks.

Other things -
* Should be ready (or at least readier) to be updated to use the dynamic service, once it exists.
* Got rid of global dom uniquely generated DOM ids in favor of locally stored refs.
* Re-shuffled around some methods for legibility.
* got rid of embedded html in favor of jquery selectors. Longing to port everything into React and redux.
* The SVG code should still be inspected more thoroughly.
* We may need to genericize things for more types of data. Or maybe not.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/kbaseExpressionVolcanoPlot.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/kbaseExpressionVolcanoPlot.js
@@ -1,8 +1,8 @@
-define(['jquery',
-    'util/string',
+define([
+    'kbwidget',
+    'jquery',
     'd3',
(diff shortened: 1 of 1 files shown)


thomasoniii:
Noting that this is TASK-816. Whoops.

briehl:
great. thanks!

PR #1130, opened 2017-08-07
TASK-959 Make the Narrative sharing panel refresh

briehl:
Whenever the sharing modal opens, it should show the current state of sharing; it shouldn't cache things from the previous time.


briehl:
TASK-960 - This also has a patch for the case where a user with admin permissions on a Narrative could try (and fail) to change permissions for the Narrative's owner. Now it should just show a static string.

That is, if user A owns a Narrative and shares with user B with edit/share permissions, user B should no longer be able to (try to) revoke user A's permission.

PR #1134, opened 2017-08-15
Diff exp matrix set viewer [TASK-816]

thomasoniii:
This is rough around the edges.

This is the initial pass at refactoring the volcano plot widget to use the ExpressionAPI.get_differentialExpressionMatrixSet to load the data.

Note that the widget can still load old-style objects via the workspace, so it's all backwards compatible.

Right now, this adds in a new HO widget called kbaseDifferentialExpressionMatrixSetViewer, in the hopes of having two split tabs - one with some sort of an overview table for the differential expression matrix set, and the second containing a volcano plot.

However - I don't know what that tabular display of the set should look like. Right now, for demo purposes, there are two tabs - The volcano plot is the main one, with "Meta" being the second and only containing the label. This can be refactored to either some sort of table or just dropped in favor of only using the volcano plot. I'm open to suggestions.

But if we're loading data, I need to know where to pull it from. ExpressionAPI.get_differentialExpressionMatrixSet returns volcano plot specific data, so I'm leery of using it for the generalized table. But maybe it's best? I'm open to suggestions.

Finally, I just picked a few random matrix sets that I had access to, and which produced some pretty crappy data. I'd greatly appreciate having a few matrix sets that have data which should look good in a volcano to test with a bit more.


eapearson:
@thomasoniii would you mind posting a shared public ci narrative to demonstrate this? I am a fan of checking out, building, and inspecting things when reviewing PRs (since Javascript is hard to read for correctness, and builds are not always deterministic...). You could include TASK-816 in the narrative title to make it easy to find (although it will be at the front of the line int he dashboard public narratives anyway.)

PR #1142, opened 2017-09-01
TASK-1042 Fix problem with loaded pre-existing code cells through the extension

briehl:


briehl:
Just a little one-liner fix. Merging.

PR #1148, opened 2017-09-07
TASK-1054 Better error message for semaphore timeouts. 

briehl:
## Things this fixes:
* make a better error message rather than just 'Timed out waiting for semaphore "comm" with value "ready"'
* fix issue where all data panel items appear to come from a data palette

## Discussion
As far as I can tell, the main problem (see JIRA TASK-1054) stems from lag that can occur when trying to build a websocket connection. Here's the usual series of events:
1. Page loads with Narrative document.
2a. The main frontend controller waits for the kernel to be ready to build its special comm channel to the backend.
2b. (same time as 2a), each app cell latches onto a semaphore - they aren't supposed to do anything until that comm channel is constructed, otherwise they'll choke while listening for messages that will never come. Also, at this stage the JobManager frontend doesn't exist yet.
3. Kernel connects, triggers "kernel_ready.Kernel" event.
4. JobManager frontend registers a semaphore as unready.
5. ... it then makes that comm channel connection.
6. ... then sets that semaphore to ready.
7. All the app cells get the message, drop the semaphore, and plug in to the JobManager for messaging.

There's lots of fiddly steps here, all focused around getting this set in the right order. The problem is that all of the app cells and the frontend controllers are distinct from each other, though they depend on each other. This is a mix of our own design and a constraint of having each app cell be its own entity (a requirement of the notebook extension interface).

The error occurs if that websocket is never formed, and the "kernel_ready.Kernel" event doesn't get fired in a timely way. The locks on all the app cells all timeout before they even really become relevant.

So I think there's a few solutions to make that error more reasonable.
1. Increase the timeout time. This doesn't really actually solve it, though it might help.
2. Block the whole page until the kernel's ready. This is probably overkill. You can't do anything with jobs (or create data viewers) unless the kernel's there, but you can still view things in place.
3. Add a timeout notification for the kernel. If, after a minute or so, the kernel still hasn't connected, it might be worth popping up an alert with some reasonable "please be patient!" message. Or at least a warning.
4. (in concert with 3) Tie the app cell semaphores to that kernel_ready event. I.e. don't start the timeout countdown until the kernel's connected. I'm pretty sure app cell rendering is blocked at that point, anyway.
4a. Make the "kernel connecting" notification much much more obvious, and link it to a message that says just how little you can do without it. Come to think of it, we should have that anyway.


briehl:
Oops, yeah, that timeout change was just in there for testing and should be put back to 60000 before merging.

From our conversation yesterday and this morning, I'm going to try to let the App Cell render itself, but have some kind of block on it, or some "not ready yet!" flag that doesn't get removed until the semaphore is given up. Then, if it times out, have a warning alert.

briehl:
Ok, this had an overhaul. Now the following happens.

1. On startup, there's a new blob of Jinja template that gets inserted. Rather than the old spinner, there's the flashing animated gif. It has beneath it a list of a few of the larger things that cause lag on load.
2. There's a new 'loadingWidget' that does the job of managing a progress bar on that loading blocker. It also makes checkmarks as things get loaded. Once all things are loaded, that widget removes the loading panel.
3. Most of the logic for responding to load events was already there, except for the app panel. So a jQuery event was inserted there (to be consistent with most of the other events at startup).
4. Some rewiring of the KBaseNarrative object to report those updates to the loading widget.
5. The semaphore lock is back in place now, with a VERY long timeout (10 minutes), until we can really solve that problem.

This is a big enough change (new feature!) that I think it's worth bumping the minor version, so I went ahead and did that.

PR #1154, opened 2017-09-10
Don't show Suggested Next Steps if there aren't any

nlharris:
Some time ago, the Suggested Next Steps stopped showing up (https://kbase-jira.atlassian.net/browse/KBASE-3778). While awaiting a fix for that, I suggested in https://kbase-jira.atlassian.net/browse/TASK-1017 that we should just not show that Suggested Next Steps panel if there aren't any (otherwise it looks like all the apps are dead ends, which they're not).
I am suggesting this fix (which simply moves the ui.setContent('next-steps') call into the if (apps.length > 0) block) but I am not set up to test it. Hopefully @thomasoniii  can test my fix and see if it works. If it does, it would be great to get this deployed to prod ASAP! @RoyKBase @sychan

### DIFF
--- a/nbextensions/appCell2/widgets/tabs/resultsTab.js
+++ b/nbextensions/appCell2/widgets/tabs/resultsTab.js
@@ -117,6 +117,7 @@ define([
             apps = apps.filter(function(app) {
                 return app.info.module_name;
             });
+            // If there are no next apps to suggest, don't even show the Suggested Next Steps panel
             if (apps.length > 0) {
                 appList = apps.map(function(app, index) {
                     return div([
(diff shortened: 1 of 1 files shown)


briehl:
@nlharris this seems fine to merge and put on CI.

But before we do, I've read through that code, and it's simple enough to test things with and fix the actual problem of the next steps not showing up. Are there apps out there that are properly wired to show their next apps? Can you link the display.yaml for one or two here? I looked at a few, and a problem might be that they're not being set correctly. For example, in HISAT2 (https://github.com/kbaseapps/kb_hisat2/blob/master/ui/narrative/methods/align_reads_using_hisat2/display.yaml), it's currently set to use two apps that don't have their full module/app_name. (that's my fault, and I'll go in and fix it after you see this note 😄 ). Those should say kb_cufflinks/assemble_transcripts_using_cufflinks and kb_stringtie/run_stringtie.

If you can't think of any off-hand, I'll make the fix to HISAT2 and see if that behaves properly. If so, it might just be a problem with the various display.yaml files out there, which will be some more work...

nlharris:
"Are there apps out there that are properly wired to show their next apps?" -- my impression (see https://kbase-jira.atlassian.net/browse/KBASE-3778?) is that that stopped working and was never fixed, but I can't swear up and down that there are display.yamls that are currently correct; I'd have to do some digging. The prime candidates that *should* work would be Annotate Microbial Assembly and Assemble Microbial Contigs (I think those are the current names? RAST and ARAST).

briehl:
From looking at that and manually going through the steps it would use to look up the info (those apps don't work in CI right now, for some reason), they fail. It's not the fault of the app cell, but they're suggesting apps that don't exist.

briehl:
Ok, just tested putting the right app id in hisat, and now it doesn't show anything at all. Seems like putting in slashes like "kb_stringtie/run_stringtie" doesn't work in yaml? I'm not sure what's going on.

Either way, that still seems like the root bug. The App cell is properly querying for apps to show, but not getting results. I'll go ahead and merge this and put it on CI soon.

PR #1155, opened 2017-09-12
Fix problem with app insertion API not respecting app tag

briehl:
Even if you use 'dev' it would default to 'release'. Actually, it was defaulting to whatever was set in the App Panel, which is typically release.


thomasoniii:
This does indeed fix my issue!

PR #1159, opened 2017-09-14
Create feature sets buttons [TASK-1057]

thomasoniii:
This pull requires the changes that Bill made to the app injection script. I *think* they're all there already.

Adds in wiring to export a selected volcano plot as a feature set, by bringing up a new pre-populated create feature set app. PLEASE NOTE - this will only work if the widget is given a differential expression matrix set as its input, and will not work with the older style volcano output from the original app. If it isn't given a ref, the export button will not appear.

This pull also piggy backs along the wiring for the explore cluster to re-enable creating as a feature set, BUT that functionality is disabled. Right now the feature set app expects the input to be an object in the current narrative workspace, but for proper functionality we need to give it an object from another workspace. So it doesn't work. Bill needs to make some mods elsewhere in the code to enable it, and once that's done we should be able to just uncomment the lines in the cluster widget to enable it.


briehl:
I think the mods to support external-workspace objects are already there, at least there enough for this case. Though you should probably pass a reference path to that app instead of just the external reference. This will solve the case where a user cannot see the other workspace, but should be able to get the object through the reference.

PR #1169, opened 2017-09-21
Fixes PTV-225

thomasoniii:
This fixes sorting by type, with a caveat -

The type will sort by the fully qualified name, which may be a little confusing to the user. e.g., in the collapsed view of the data, the user may see "GenomeAnnotation" and "Media" and expect "GenomeAnnotation" to be first in the sorted list. It won't be. 

That's because the actual types are "KBaseGenomeAnnotations.GenomeAnnotation" and "KBaseBioChem.Media", and will in turn sort KBaseBioChem.Media first.

I left the behavior as is, since the user could expand the data item and see the full name, but we may want to change this.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
@@ -1492,8 +1492,8 @@ define([
                 .append('type')
                 .on('click', function () {
                     self.sortData(function (a, b) {
-                        let aType = self.dataObjects[a.objId].info[2].toUpperCase().match(/\.(.+)/)[1];
-                        let bType = self.dataObjects[b.objId].info[2].toUpperCase().match(/\.(.+)/)[1];
+                        var aType = self.dataObjects[a.objId].info[2].toUpperCase().match(/\.(.+)/)[1];
+                        var bType = self.dataObjects[b.objId].info[2].toUpperCase().match(/\.(.+)/)[1];
                         if (aType > bType)
                             return -1; // sort by type
                         if (aType < bType)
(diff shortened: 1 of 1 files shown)


briehl:
I think this is a good start, but since the collapsed view is mostly what users see, we should probably sort by that and not the fully qualified name.

Then again, I'm not sure what the best answer is. @sychan? @nlharris?

briehl:
I've thought about this more, and it makes more sense to filter by the shortened name. Other places in that interface (the filter dropdown) just make use of the shortened name, and we should be consistent. Also, the shorter terms like "Media" are the first things that users will see, and I suspect the only things for many users.

nlharris:
I agree. Users will be confused if the sort is by something they don't see. Use the shortened ("pretty") names for sorting.

thomasoniii:
Updated so it'll sort on the non-qualified name.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
```diff
@@ -1492,9 +1492,11 @@ define([
                 .append('type')
                 .on('click', function () {
                     self.sortData(function (a, b) {
-                        if (self.dataObjects[a].info[2].toUpperCase() > self.dataObjects[b].info[2].toUpperCase())
+                        let aType = self.dataObjects[a.objId].info[2].toUpperCase().match(/\.(.+)/)[1];
```
briehl: Sadly, we're still in ES5 (and there's no Babel installed on the build path yet...), so "let" is what's causing the build to fail.

briehl: That took me entirely too long to track down. Sorry!

PR #1173, opened 2017-09-27
Fixes filtering [KBASE-4756]

thomasoniii:
This fixes Ticket KBASE-4756.

The filter list on the data panel only presented the type name to the user, but there are types which exist in multiple modules. As a result, filtering didn't work. If a user had a "KBaseAlpha.Foo" object, but the widget was filtering on "KBaseBeta.Foo", nothing would show up.

This change does a couple of things:
* It removes the hardwired list of types from the widget. It was incomplete and out of date.
* It dynamically builds a new list of types to present to the user, based upon the results of getAndRenderData. That way, the user will only see the option to filter to types that they have, instead of the massive list of all possible types.
* The filter list includes all modules in its filtering. So it'll show "Foo" in the dropdown, but actually filter on "KBaseAlpha.Foo" and "KBaseBeta.Foo", as well as any other "Foo" types in arbitrary modules that the user may have.

There may be other tickets that reference this issue. Kbase-5051 is tangentially related.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataPanel.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataPanel.js
@@ -51,6 +51,26 @@ define([
     GenericClient
     ) {
     'use strict';
+
+    /*
+      as a bit of trivia, in case it comes up....
+
+      I'm pretty sure this could have a race condition which is not currently manifesting.
+      It rebuilds the list of knownTypes inside the getAndRenderData, which is called when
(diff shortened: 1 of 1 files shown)


thomasoniii:
That was way more painful to make work than it ever should've been. :-P

briehl:
Looks okay to me. I think we should lend an eye toward refactoring that whole slideout, for a number of reasons. Eventually, also, that list of available types in both "My Data" and "Shared with Me" should be retrieved by a service.

But I think this will patch the immediate problem. Let's get more eyes on it on CI.

PR #1177, opened 2017-10-04
TASK-1050 some frontend test refactoring

briehl:
Tests should now generate code coverage (still pending some checking to see if that functions properly. Even with more complete tests, it seems like there's some false negatives).

There's also a test utility module that helps manage tokens.

(doing these later)
~~TODO:  
Add util for creating a narrative  
Add util for uploading test data  
Add util for deleting workspaces (maybe. might also just wait until the Mini-KBase stuff is done and operate against that environment)~~


briehl:
There's still some TODOs that'll make life easier, but I think we should get this merged, if only to expose coverage and Auth2 token use.

@eapearson @sychan mind taking a look here?

PR #1178, opened 2017-10-04
Add lock on refresh when changing name [TASK-1114]

tiramisu24:
- move datapanel tooltip to not cover any buttons/information
- add on lock on refresh function when focusing on rename input field of objects
- relinquish after:
    - 15 min of inactivity
    - click rename or cancel button
    - closing panel
    - add, delete, revert object
    - clicking refresh button 
- update release notes

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
@@ -129,8 +129,8 @@ define([
             return obj[6] + '/' + obj[0];
         },
 
-        writtingLock: false,
-        refreshWrittingLock:null,
+        writingLock: false,
+        refreshwritingLock:null,
 
         /**
          * Test if given object is a set.
@@ -329,7 +329,7 @@ define([
             this.serviceClient = new GenericClient(Config.url('service_wizard'), auth);
             this.my_user_id = auth.user_id;
             this.isLoggedIn = true;
-            this.writtingLock = false;
+            this.writingLock = false;
             this.refresh();
             return this;
         },
@@ -359,7 +359,7 @@ define([
         },
 
         refresh: function (showError) {
-            if(this.writtingLock) {
+            if(this.writingLock) {
                 return;
             }
             // Set the refresh timer on the first refresh. From  here, it'll refresh itself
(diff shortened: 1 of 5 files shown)


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
```diff
@@ -355,6 +359,7 @@ define([
         },
 
         refresh: function (showError) {
+            if(this.writtingLock) return;
```
briehl: If (and other control) statements like this should be on a separate line.

We haven't talked about this yet, but we're mostly trying to follow the standard ESLint conventions -
 https://eslint.org/docs/developer-guide/code-conventions  They haven't been enforced much in the past, but now's the time! And we can clean up older code if we see issues. I'm pretty sure VSCode as ESLint built in, if not there's a plugin for it.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
```diff
@@ -901,11 +907,21 @@ define([
                                     .addClass('text-warning')));
                         return;
                     }
+
+                    //lock on refresh expires aftger 5 min
+                    var releaseLock = function(){
+                        if(self.refreshWrittingLock !== null) clearTimeout(self.refreshWrittingLock);
```
briehl: See above about if statements on one line.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
```diff
@@ -901,11 +907,21 @@ define([
...
+                        
+                        self.refreshWrittingLock = setTimeout(function () {
+                            self.writtingLock = false;
+                        }.bind(this), 900000);
```
briehl: .bind(this) isn't necessary since self should still be in scope by closure.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
briehl: Also, I didn't mark them, but there's a few instances where there's an if one-liner that should be surrounded by brackets. E.g.:
```
if (something) 
    doSomethingElse();
```
should be
```
if (something) {
    doSomethingElse();
}
```
I was reminded earlier that doing things like this can easily lead to dangerous bugs...

MrCreosote: In particular https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

tiramisu24:
added brackets to all the if statements

PR #1183, opened 2017-10-06
3.5.1 release

briehl:


briehl:
Going ahead and merging this to get it on next, also so the PR doesn't accrue anything else from develop.

PR #1187, opened 2017-10-19
re add old auth

tiramisu24:
Adding the old auth back since the rest of the file still depends on user variable. Will need to refactor for new auth


briehl:
Just adding a note here - this doesn't add back anything to do with auth, it just makes a local variable to the current user id. There's no need to refactor - that's already done, the previous PR just removed too much code and blew away that variable.

PR #1189, opened 2017-10-19
bug fixes from giant pull request [TASK-1158]

tiramisu24:
-updates button text in all tabs at same time
-fix bug with star
-update app card to look the same as prior card


tiramisu24:
had a stray let which caused the build to fail

PR #1191, opened 2017-10-20
TASK-1128 Introduces a dropdown that populates itself from the staging area

briehl:
## MERGE https://github.com/kbase/narrative/pull/1177 FIRST!
(this includes tests built on the newer configuration)

This is getting a little large, so starting a PR here.
The goal is to create an import widget for an App Cell. This should be a dropdown menu that populates itself by searching for files in the FTP staging area.

The next PR will introduce a mechanism for it to query against dynamic services. But that will likely require some updates to the Narrative Method Store to get the parameters in the App Spec, so it's not yet.

This creates a new parameter type in App Specs called "dynamic_dropdown". This gets interpreted as a string and entered into the Python app running code as a file path. It makes use of Select2, so that when a user clicks the dropdown, it'll do a search against the FTP service (assumes a working FTP endpoint with a '/search/' route) and populate the dropdown based on that search. Also assumes the results are of a specific structure, but that'll also change when this gets adapted for dynamic services.

Still todo on this PR:
- [x] tests
- [x] dynamicDropdownView.js (need a view-only mode)
- [x] revert to FTP endpoint URL in the config (oops)
- [x] more styling to the dropdown (I don't think "last modified" is appropriate for FTP files that are effectively untouchable after uploading. should be "uploaded" with the actual timestamp)


briehl:
Needs tests now.

kbase-extension/static/kbase/js/narrativeConfig.js
```diff
@@ -105,62 +105,59 @@ define([
      */
     function updateConfig() {
         return new Promise.try(function (resolve, reject) {
```
eapearson: this should be just 
`new Promise(function (resolve, reject) {...`
also if anything throws in here (`assertConfig()`?) it will not cause the promise to reject, but rather the constructor will fail and throw.

kbase-extension/static/kbase/js/narrativeConfig.js
```diff
@@ -105,62 +105,59 @@ define([
...
-                        require.config(pathConfig);
-                        config.new_paths = pathConfig;
-                        resolve(config);
-                    }, function () {
-                        console.warn("Unable to get updated widget paths. Sticking with what we've got.");
-                        resolve(config);
-                    });
-                } else {
+            if (window.kbconfig) {
+                return window.kbconfig;
```
eapearson: also, this won't resolve with the config -- it would have if this was original simply Promise.try()

kbase-extension/static/kbase/js/narrativeConfig.js
```diff
@@ -105,62 +105,59 @@ define([
...
-                        console.error('Config: unable to process local configuration options, too! Public and Example data unavailable!');
-                        return Promise.try(function () {
-                            return config;
-                        });
+            return Promise.resolve($.getJSON('static/kbase/config/' + path[path.length - 1]))
+                .then(function (dataCategories) {
+                    console.log('Config: processing local data configuration.');
+                    config.publicCategories = dataCategories[config.environment].publicData;
+                    config.exampleData = dataCategories[config.environment].exampleData;
+                    return Promise.try(function () {
```
eapearson: and these Promse.trys are not necessary since they simply return a value -- just returning the value will continue the promise chain.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -0,0 +1,323 @@
...
+            return model.value;
+        }
+
+        // VALIDATION
+
+        function validate() {
+            return Promise.try(function() {
+                var selectedItem = getControlValue(),
+                    validationConstraints = {
+                        min_length: 1,
```
eapearson: shouldn't the validation constraints be driven from the spec or sdk.js? trawling back through memory of how this is used elsewhere.

eapearson: also in theory wouldn't the validation go against the data source? I mean, in actual usage the value is derived from a dropdown derived from the data source, but as far as the control knows, at some future time this may not be true. That is a problem in general for this control, though, since all files will eventually disappear and thus every instance of this parameter will become invalid. Hmm..

test/unit/spec/appWidgets/input/dynamicDropdownInputSpec.js
```diff
@@ -0,0 +1,70 @@
+/*global define*/
```
eapearson: wow, parameter control specs!!

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
briehl: Yes, but these aren't expected to have string constraints. Those are just there as default values, though sdk.js should cough up the defaults itself, instead of the input widget. I'll change that.

test/unit/spec/appWidgets/input/dynamicDropdownInputSpec.js
briehl: Not perfect, but it's a start! :)
I'm trying to not introduce any new code without a few tests. Hopefully, it'll become a trend...

briehl:
Ok, adjusted stuff with your comments, and fixed tests so they get actually get skipped when no auth is present.

PR #1194, opened 2017-10-27
Small refactor to import panel

briehl:
* Cleaned up some things that made my linter angry - mainly whitespace and quotes.
* Changed staging area viewer to use newer API calls for handling things (fetching files, inserting app cells).
* Tweaked the file staging client to be a little more flexible with its list function.

### DIFF
--- a/kbase-extension/static/kbase/js/api/fileStaging.js
+++ b/kbase-extension/static/kbase/js/api/fileStaging.js
@@ -71,10 +71,7 @@ define([
                 }
                 path += directory;
             }
-            return makeFtpCall('GET', path)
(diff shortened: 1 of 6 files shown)


briehl:
@thomasoniii - let's get this merged before your changes start to conflict too much.

briehl:
I see now. I put that no-op error there just as a convenience. I had it tooled together that if a list call fails, it should show an empty list. But you're right, if an error occurs there should at least be a warning.

Actually, it should be up to whatever invokes the client to catch the error and deal with it there. So, really, the client should just pass that through. I'll make the adjustment.

PR #1198, opened 2017-10-31
Struct to image

JamesJeffryes:
This adds an image to the compound detail page, either from a precomputed directory (if it's a standard compound) or from the structure (if it's novel). Also adds Inchikey & SMILES info to compound details.


kbase-extension/static/kbase/js/widgets/function_output/modeling/KBModeling.js
```diff
@@ -44,7 +44,7 @@ function KBModeling(token) {
             method = 'fbaModelServices.'+method;
             return call_ajax(url, method, params)
         } else if (service == 'biochem') {
-            s_url = 'https://kbase.us/services/service_wizard';
+            s_url = window.kbconfig.urls.service_wizard || 'https://ci.kbase.us/services/service_wizard';
```
briehl: Just making a(nother?) note that we need to eventually update these widgets to get loaded with require like everything else. Then we can just include the config module and get URLs from there, instead of depending on a global variable that may or may not exist.

But for now, this is fine.

kbase-extension/static/kbase/js/widgets/function_output/modeling/KBaseFBA.FBAModel.js
```diff
@@ -331,27 +329,46 @@ function KBaseFBA_FBAModel(modeltabs) {
...
     this.CompoundTab = function (info) {
         var cpd = self.cpdhash[info.id];
+        console.log(cpd)
 		var output = [{
 			"label": "Compound",
 			"data": cpd.dispid,
-		}, {
+		},{
+            "label": "Image",
+            "data":  "<img src=http://minedatabase.mcs.anl.gov/compound_images/ModelSEED/"+cpd.id.split("_")[0]+".png style='height:300px !important;'>"
```
briehl: The url here (or at least the root of the pattern) should be in the same config file - `src/config.json`

kbase-extension/static/kbase/js/widgets/function_output/modeling/kbaseTabTable.js
```diff
@@ -411,16 +411,16 @@ return KBWidget({
         };
 
         this.compoundImage = function(id) {
-            return 'http://bioseed.mcs.anl.gov/~chenry/jpeg/'+id+'.jpeg';
+            return "<img src=http://minedatabase.mcs.anl.gov/compound_images/ModelSEED/"+id.split("_")[0]+".png style='height:300px !important;'>"
```
briehl: add urls to config as above

kbase-extension/static/kbase/js/widgets/function_output/modeling/kbaseTabTable.js
```diff
@@ -411,16 +411,16 @@ return KBWidget({
...
         };
 
-        var imageURL = "http://bioseed.mcs.anl.gov/~chenry/jpeg/";
+        var imageURL = "http://minedatabase.mcs.anl.gov/compound_images/ModelSEED/";
```
briehl: add urls to config as above.

briehl:
The error here is a false-positive - it's Travis being a jerk about not wanting to start up the tests properly.

kbase-extension/static/kbase/js/widgets/function_output/modeling/KBModeling.js
JamesJeffryes: I don't know much about require.js but if we get a refactor of these viewers in the roadmap I'd take a look at it.

JamesJeffryes:
I looked for code examples before reading you comment at the top carefully so I get the url from Config.url not window.kbconfig.urls. Can switch this if the other way is prefered

kbase-extension/static/kbase/js/widgets/function_output/modeling/KBModeling.js
briehl: Sure. It'll be some work to do, mainly since all of the modeling viewer widgets depend pretty heavily on each other being in the global namespace. It's not super high priority yet (though it will be), so don't worry about it for now.

briehl:
Yeah, if the narrativeConfig module is available to those widgets, it should definitely be favored over window.kbconfig.

briehl:
Anyway, this looks fine now. Let's merge.

PR #1200, opened 2017-11-02
Fix Travis-CI testing

briehl:
Apparently Travis-CI no longer likes the Jupyter notebook being started up on `localhost`, but it will handle `127.0.0.1`.

As of, like, Tuesday.

Hopefully this fixes it for the foreseeable future.

PR #1201, opened 2017-11-03
Staging panel mods [TASK-1134]

thomasoniii:
I wasn't sure of the task to tack onto this, but 1134 seemed reasonable.

This is the initial crack at updating the staging panel. We should get it up onto CI as soon as we can.

It includes:
* mods from Paramvir's staging panel mockup, with a few tweaks (which may or may not be desirable)
* integration of the staging service to get extra file info.
* a new abstract RestAPIClient object, with a StagingServiceClient object, to hide the rest calls. It's intentionally a bit rough, and it'd probably be good for various frontend people to discuss refining it. But that's not urgent.
* A rough resize on the drop zone.

It lacks:
* Decompressing files
* info on whether a file was imported.
* The info panel ends up striped in a different color than the file it's giving info on, due to how it's added. This is a bit ugly.

Right now, I'd just like some more eyes on it to see how it's coming along.


kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -40,7 +46,14 @@ define([
         name: 'StagingAreaViewer',
 
         init: function(options) {
+
             this._super(options);
+
+            this.stagingServiceClient = new StagingServiceClient({
+              root : 'https://ci.kbase.us/services/staging_service',
```
briehl: This should be taken from `Config.url('ftp_api_url')`

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -52,6 +65,10 @@ define([
             this.fileStagingClient = new FileStaging(
                 this.ftpUrl, this.userId, {token: runtime.authToken()});
 
+            this.genericClient = new GenericClient(Config.url('service_wizard'), {
+              token: Runtime.make().authToken()
```
briehl: There's already an initialized local handle on Runtime. So just `runtime.authToken()` works here.

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -52,6 +65,10 @@ define([
             this.fileStagingClient = new FileStaging(
```
briehl: We should pick one Staging Service client and roll with it. We both rolled our own. I gotta say, I like yours better because it comes with a more generic REST client that we could use elsewhere.

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
thomasoniii: All changes made.
* I got rid of references to the FileStaging() client in favor of StagingServiceClient
* It can't use the ftp url, since it doesn't have the same routes. I added a staging_api_url to src/config.json, including the fake pointers for appdev, next, and prod services which don't exist yet.

briehl: the FTP url is, i guess, deprecated at this point. we can remove it later. I'm planning on scouring through the repo and ditching all kinds of crap - that can be part of it. :)

briehl:
Hey, that's the same Travis-CI problem that's fixed in #1200! The rest looks fine. We can merge and redeploy if you're ready.

thomasoniii:
I'm good to go whenever you are. As long as that travis thing is a non-issue, go ahead and pull the trigger.

thomasoniii:
Just as an FYI, at the moment the staging_service is tossing a 500 error on a list/ hit, which means that no files are listed and showing up on the panel. I just tweaked it so that if it fails in that case it reports to the user instead of just logging to the console.

I'm pretty sure it's a service issue instead of the widget, since I see the same 500 error if I try to hit the URL directly. It'd probably be good to hold off on merging and deploying this until everything is working properly.

PR #1203, opened 2017-11-07
Take ContigSet out of Example Assemblies

nlharris:
Fixes https://kbase-jira.atlassian.net/browse/PTV-795


nlharris:
You are right, sorry for being so un-thorough. See revised PR.

briehl:
Great. Thanks!

PR #1205, opened 2017-11-08
Delete lots of dead code

briehl:
Most of the old service clients, some MG-RAST things that incredibly early versions of the Narrative made some use of, and so on.

None of this is currently referenced by the Narrative, so it should go.


sychan:
I remember deleting the Github repos for most of those clients!

PR #1206, opened 2017-11-09
Upload Reads Demo with staging area dropdown

briehl:
Includes tweaks to the staging viewer to set the path as the file input, not just the filename.


scanon:
Looks safe enough... Just CI.

PR #1208, opened 2017-11-09
fix problem with showing subfolders in staging area

briehl:
~~not ready to merge yet~~ - still need to get the metadata dropdown working in subfolders.
Also fixes problem with searching in the new dynamic dropdown input.

Edit: work's done, working on some kind of tests. But ready to review.


briehl:
@thomasoniii I was working on this anyway since I was blocked by similar problems. I think I have that fixed. Mind taking a look?

briehl:
Could use some more tests, but the import dropdown widget should be mostly done.

PR #1211, opened 2017-11-16
fix release notes for 3.5.1

briehl:
The release notes got mangled somehow. This fixes them.


briehl:
Yup, that's right. Don't need to build a new image.

Also, tests are failing because of a problem with the test runner that's been resolved in the develop branch. This is safe to merge anyway, since it's just the release notes doc.

PR #1213, opened 2017-11-17
Staging panel updates

thomasoniii:
This integrates changes from the feedback doc here:
https://docs.google.com/document/d/1LLhFVipqMRkZ16IGFwwbKeX5-VDecMkombH1hk-6_2g/edit#

* Age is now rendered with "short" names, i.e. "88 secs" instead of "88 seconds ago"
* The remove button is moved to the right, and framed in a button. Still inside the expanded area, though.
* Changed formatting of Info field - bolded text, changed size, removed colons, added commas to line count.
* first 10 / last 10 are in a slightly different gray and monospace font
* fixed the bug that could cause metadata to not appear if sorted
* added the decompression button, which only shows up for files assumed to be compressed.
* Added "All files will be deleted after 90 days" footer.

errata -
* I also added a change to swap the directional arrow on the data panel. It's not *quite* perfect, since the uploader button won't toggle it back. But it's real close. I'll roll it back if we don't want it.
* I was unable to reproduce the first 10 / last 10 unresponsiveness.
* I left it so that it remains open unless explicitly closed, due to ambiguousness of request. Will revisit if we're sure we want to do this.
* I'm confirming with Josh, but it sounds like the "Imported as" metadata item (which lists any UPAs the file was inserted into) won't work without updates to the importers to add the metadata. So the field is there, but I think the importers need to be updated to populate it. Will re-visit if I was mistaken and the widget can address it.

### DIFF
--- a/kbase-extension/static/kbase/js/util/timeFormat.js
+++ b/kbase-extension/static/kbase/js/util/timeFormat.js
@@ -236,6 +236,7 @@ define([], function() {
      * edited from: http://stackoverflow.com/questions/3177836/how-to-format-time-since-xxx-e-g-4-minutes-ago-similar-to-stack-exchange-site
      */
     function getTimeStampStr (objInfoTimeStamp, alwaysExact) {
+
         var date = new Date(objInfoTimeStamp);
         var seconds = Math.floor((new Date() - date) / 1000);
         if (seconds < 0) {
@@ -298,6 +299,22 @@ define([], function() {
         return pluralizeTimeStr(Math.floor(seconds), 'second');
     }
 
+    // There was a request that the staging panel tighten up the display a little bit, which was using getTimeStampStr to format the age.
+    // So this function will just take the output of regular getTimeStampStr and shorten it up a little bit.
+    // * Replace "long" durations with shortened ones.
+    // * drop " ago" from the end.
+
+    function getShortTimeStampStr (objInfoTimeStamp, alwaysExact) {
+      var longTimeStampStr = getTimeStampStr(objInfoTimeStamp, alwaysExact);
+      longTimeStampStr = longTimeStampStr.replace(/month/, 'mon');
+      longTimeStampStr = longTimeStampStr.replace(/hour/, 'hr');
+      longTimeStampStr = longTimeStampStr.replace(/minute/, 'min');
+      longTimeStampStr = longTimeStampStr.replace(/second/, 'sec');
+      longTimeStampStr = longTimeStampStr.replace(/\s*ago\s*/, '');
+
+      return longTimeStampStr;
+    }
+
     /**
      * @private
      * @method
@@ -345,6 +362,7 @@ define([], function() {
         reformatDate: reformatDate,
         reformatISOTimeString: reformatISOTimeString,
         getTimeStampStr: getTimeStampStr,
+        getShortTimeStampStr : getShortTimeStampStr,
         readableTimestamp: readableTimestamp
     };
 });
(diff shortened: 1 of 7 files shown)


kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -269,29 +296,39 @@ define([
             $tabsDiv.empty();
             var data = JSON.parse(dataString);
 
+            var ulCSS = {
```
briehl: why not make this a class?

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -269,29 +296,39 @@ define([
...
+              'font-family' : 'OxygenBold',
+              width : '115px',
+              display : 'inline-block',
+              'font-size' : '90%',
+            }
+
+            var linesCSS = {'white-space' : 'pre', 'overflow' : 'scroll', 'font-family' : 'monospace', 'font-size' : '90%', 'background-color' : '#DDDDDD'};
```
briehl: This one, too?

kbase-extension/static/kbase/js/util/timeFormat.js
```diff
@@ -298,6 +299,22 @@ define([], function() {
         return pluralizeTimeStr(Math.floor(seconds), 'second');
     }
 
+    // There was a request that the staging panel tighten up the display a little bit, which was using getTimeStampStr to format the age.
+    // So this function will just take the output of regular getTimeStampStr and shorten it up a little bit.
+    // * Replace "long" durations with shortened ones.
+    // * drop " ago" from the end.
+
+    function getShortTimeStampStr (objInfoTimeStamp, alwaysExact) {
```
briehl: A few tests should be added here. Check out `test/unit/spec/Util/timeFormatSpec.js`

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
thomasoniii: I hate external class files. :-P

I think it's tougher to keep 'em in sync with the code, and cruft accumulates too easily when usage of the class is deleted but the actual class still hangs around in the stylesheet.

Nonetheless, I retooled to use a class, and added them into kbase-extension/static/kbase/css/kbaseNarrative.css

Lemme know if they should be in a different file.

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
thomasoniii: Same as above

kbase-extension/static/kbase/js/util/timeFormat.js
thomasoniii: I added in several tests to verify the new function.

While I was at it, I discovered that the existing getTimeStampStr would favor a larger interval of smaller units if it was divisible by one.
i.e., instead of saying "1 day ago" it would be "24 hours ago". Or 1 hour -> 60 minutes, 1 minute -> 60 seconds.

I re-tooled that to favor smaller increments of the larger interval, and added a few more tests to confirm that behavior as well, on both functions.

briehl:
Looks good to me. Let's merge!

PR #1214, opened 2017-11-18
[WIP] TASK-1270, TASK-1271 Update Output cell mechanism

briehl:
The first stage here is to update the output/viewer cells for some consistency. Before merging, these cells should (see checklist below):

(see here for some details on requirements, etc.: https://docs.google.com/document/d/1Ie7qb8rEvnxXi_jrgYu2TIWmhNIuHjqaYA4J3AFCof8/edit )

Note that this work might get separated into multiple PRs, I'm just making a checklist here of the broad strokes.

- [x] Have code that calls to biokbase.narrative.widgetmanager to make the viewer.
- [x] Update WidgetManager.show_data_cell to just generate a kbaseNarrativeOutputCell with the directly requested viewer widget, NOT a kbaseNarrativeDataCell widget.
- [x] Tests for above.
- [x] Include an UPA field in all generated JS code.
- [x] Cleanup functional (but sloppy) WidgetManager.infer_upas function.
- [x] kbaseNarrativeOutputCell should serialize/deserialize UPAs automatically before running the viewer.

These will be worked on in another PR.
- Include an auto-updater for older widgets (or maybe just flag those cells to be executed again if the Narrative was last saved before some date).
- Every last freaking viewer needs to be updated to take one (or more) UPAs as input. That'll be later.


briehl:
Okay, some other work crept in here, but here's the main list of architectural changes. I think all the creep is below, too. The core target - update the viewer cell instantiation mechanism while leaving it backward-compatible - is still the same.

1. Update backend calls that create viewer widgets to only make kbaseNarrativeOutputCells. (e.g. remove the kbaseNarrativeViewer or kbaseNarrativeDataCell constructs that just wrapped around kbaseNarrativeOutputCell).
2. Update backend calls to automatically do all the mapping from an object type -> viewer widget and the parameters for it (deprecates kbaseNarrativeDataCell and narrativeViewers JS modules). This means adding a mapping for all current widgets so we know the semantics of each parameter (see src/widget_param_mapping.json). Hopefully, future widgets will just have an UPA field.
3. Change kbaseNarrativeOutputCell to serialize/deserialize UPAs that are passed in to it.
4. Change kbaseNarrativeOutputCell to automatically inject all UPAs into the viewer widget it wraps.
5. (as a test) update kbaseRNASeqPieNew.js to handle upas as input. (the rest will come later)
6. A bunch of tests for the above.

extra feature creep
1. Adds the first pass at a general information header for all data cells. From here, the user can toggle which version of the object is being seen. This gets serialized as part of the UPA, so it should be restored on page load.
2. Remove the (essentially redundant) bootstrapAlert module. Adds a 'alertOnly' option to bootstrapDialog - set that option to be true gives the same functionality as the other module.
3. A bunch of code cleanup from seemingly random places - remove commented out code, boilerplate comments, x'd out css classes, etc.
4. There was a piece of rarely used code in the backend that would try to parse user ids from the current auth token. It does the right thing now.
5. Fell down a rabbit hole of adding a number of other tests to repos with holes in them.

kbase-extension/static/kbase/js/api/upa.js
```diff
@@ -108,13 +108,65 @@ define([
...
+            if (typeof upas === 'string') {
+                return serialize(upas);
+            }
+            else if (Array.isArray(upas)) {
+                var serialArr = [];
+                upas.forEach(function(upa) {
+                    serialArr.push(serialize(upa));
+                });
+                return serialArr;
+            }
```
thomasoniii: I never remember what's available...do we have map in our environment? If so, this could just be:

return upas.map( function(upa) { return serialize(upa) } );

Not actually necessary to fix, just a little prettier. Even better when we could have

return upas.map( upa => serialize(upa) )

kbase-extension/static/kbase/js/api/upa.js
```diff
@@ -108,13 +108,65 @@ define([
...
+            else {
+                var serialSt = {};
+                Object.keys(upas).forEach(function(key) {
+                    serialSt[key] = serializeAll(upas[key]);
+                });
+                return serialSt;
+            }
```
thomasoniii: Same comment as above, could be:
return upas.reduce( function(acc, key) { acc[key] = serializeAll(upas[key]); return acc }, {} );

...but doesn't need to be.

kbase-extension/static/kbase/js/api/upa.js
```diff
@@ -108,13 +108,65 @@ define([
...
+                    deserialArr.push(deserialize(upa));
+                });
+                return deserialArr;
+            }
+            else {
+                var deserialSt = {};
+                Object.keys(upas).forEach(function(key) {
+                    deserialSt[key] = deserializeAll(upas[key]);
+                });
+                return deserialSt;
```
thomasoniii: Same commentary on 145...156 - could be a map and a reduce, but not urgent. It's such a big pull I feel obligated to add notes. :-)

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeOutputCell.js
```diff
@@ -25,12 +33,43 @@ define([
...
+                }
+                return metadata;
+            }
+        },
+
         init: function (options) {
+            // handle where options.widget == null.
+            options.widget = options.widget || this.options.widget;
+            // handle where options.widget == 'null' string. I know. It happens.
+            options.widget = options.widget === 'null' ? this.options.widget : options.widget;
```
thomasoniii: Before coming to CSHL, I worked at Edison Schools and we imported school data from their SIS systems into our servers every night. One night, it broke at a school. I dug into it. Turns out they had a new kid whose name was "Brandon Null", and our importer failed because it said that his last name was a required field.

So our CSR called up the school and said he was missing his last name. They replied, "What are you talking about? His name is Brandon Null. That didn't go through? Why not?" Poor kid. He's gonna have all sorts of problems in life.

The quick fix was when I realized that the importer only failed on insert, not update - so I just change his name to "Brandon NotNull", imported it, and then went into the database to manually change back to "Brandon Null". After that, his record would update without issue.

We never encountered that specific bug again during my tenure.

test/unit/spec/Util/BootstrapDialogSpec.js
```diff
@@ -72,5 +78,54 @@ define (
...
+            });
+
+            var btns = alert.getButtons();
+            expect(btns.length).toBe(1);
+            expect(btns[0].innerHTML).toBe('Close');
+        });
+
+        it('Should set the title based on type', function() {
+            var d = new Dialog({
+                title: 'Foo',
```
thomasoniii: I'd recommend making this something more user-friendly. I know that the title should be set based on the type, but just in case there's some odd condition that cause it not to be set...maybe make the default either a blank string or something like "Title should have been set by type"

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeOutputCell.js
briehl: Ha! Well, all over the place in various app specs, there's the JSON value `null` mixed in with the string `"null"`.

Makes for some fun trickiness there, too.

kbase-extension/static/kbase/js/api/upa.js
briehl: Yeah, map's available. Not sure why I never think of it. I'll switch that out.

briehl:
This is great so far. Thanks for the feedback! Added the suggestions (and twiddled some tests around, too). I keep forgetting that reduce is just built in to Javascript...

briehl:
Ok, after some conversation, I think this'll get frozen for now and merged next week, after staging up a new release.

PR #1219, opened 2017-12-12
PTV-833 Fix genome icons in data panel

briehl:
A recent update replaced the genome icons in the data panel with the default icon (document). This brings them back.

The problem was with how the object type was getting interpreted by the new KBaseDataCard abstraction. I added a new "viewType" parameter to that (so the "type" string could continue to have things like the scientific name), and shuffled around how the actual data type gets read.

Also did a little bit of general cleanup on a few things.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseDataCard.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseDataCard.js
@@ -1,215 +1,198 @@
 /**
  *  kbaseDataCard.js -- used making object cards in narrative
  *
- *  Authors: zzheng@lbl.gov
+ *  Authors: zzheng@lbl.gov, wjriehl@lbl.gov
  *
- *   Example and expected out put:s
+ *  Example and expected outputs
  *
- *     var $card = new kbaseDataCard(
-                {   
-                    //expected values
-                    object_info: object_info (array)
-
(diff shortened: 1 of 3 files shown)


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseDataCard.js
```diff
@@ -1,215 +1,202 @@
...
+        //params
+        var shortName = entry.name ? entry.name : object_info[1],
+            version = entry.version ? entry.version : ('v' + object_info[4]),
+            date = entry.date ? entry.date : TimeFormat.getTimeStampStr(object_info[3]),
+            editBy = entry.editedBy ? entry.editedBy : (' by ' + object_info[5]);
+
+        // in order - entry.viewType > entry.type > parsing it out of the object_info type string.
+        var objectType = entry.type ? entry.type : object_info[2].split('.')[1].split('-')[0],
+            viewType = entry.viewType ? entry.viewType : entry.type;
+        viewType = viewType ? viewType : objectType;
```
tiramisu24: viewType set twice seems pretty confusing. Maybe break it out into if statements?

briehl: You're right. Got too clever for my own good. Fixed.

PR #1224, opened 2017-12-20
adds contents of metadata file to metadata tab [STC-284]

thomasoniii:
This will add an additional tab to populate file metadata, if it exists.

Notes:
* It manually finds the metadata file by looking at the first part of the filename before the first period, and then appends ".metadata" to the end. So that if your file is "foo.bar.baz.fasta.gz", it'll look for a metadata file called "foo.metadata". This is reasonably brittle, in that if the naming convention changes it won't find metadata. Arguably, this should be wrapped up into the staging service client, but as of yet this is the only way to get at it.

* It was mentioned on the call yesterday that metadata files will no longer be visible to the user - we need to make sure that this method still works even if the metadata files cannot be seen. If not, then we're back to what was said above about adding it into the service call itself.

* While testing, I discovered an issue with a NaN in the JSON, which is elaborated upon here: https://kbase.slack.com/archives/C4E7KUGTD/p1513782220000601

* There's a momentary blink before the metadata shows up, since it's two requests which happen in parallel. Again, I don't have a good way to fix this unless it's added into the service as an explicit key associated with the original file's metadata.

* If the metadata is created with line breaks, this method will fail. It assumes all contents are on the first line.

* The original ticket alluded to a "jgi_metadta" key, which I didn't see. This just dumps out the entire metadata file on a tab called "Metadata", assuming one exists as per the conditions up above.

### DIFF
--- a/kbase-extension/static/kbase/css/kbaseNarrative.css
+++ b/kbase-extension/static/kbase/css/kbaseNarrative.css
@@ -31,7 +31,8 @@
  overflow : scroll;
  font-family : monospace;
  font-size : 90%;
- background-color : #DDDDDD
+ background-color : #DDDDDD;
+ max-height : 200px;
 }
 
 .kb-dropzone {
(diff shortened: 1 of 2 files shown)


briehl:
> Arguably, this should be wrapped up into the staging service client, but as of yet this is the only way to get at it.

It would be nice if the staging area call would return file paths for metadata files associated with actual data files. Maybe a later version?

Anyway, let's get this on CI for testing today.

PR #1227, opened 2018-01-06
add 'is_temporary' field to saved narrative objects, not just workspace metadata

briehl:
All this needs to do from here (other work is in kbaseapps/NarrativeService) is to ensure that there's an "is_temporary" field in the saved object metadata, and it's set to "false" on save.

The user should be prevented from saving an "untitled" narrative elsewhere.

### DIFF
--- a/src/biokbase/narrative/contents/narrativeio.py
+++ b/src/biokbase/narrative/contents/narrativeio.py
@@ -144,7 +144,7 @@ class KBaseWSManagerMixin(object):
         self._test_obj_ref(obj_ref)
         try:
             if content:
-                nar_data = self.ws_client().get_objects([{'ref':obj_ref}])
+                nar_data = self.ws_client().get_objects([{'ref': obj_ref}])
                 if nar_data:
                     nar = nar_data[0]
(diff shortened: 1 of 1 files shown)


src/biokbase/narrative/contents/narrativeio.py
```diff
@@ -211,6 +211,7 @@ def write_narrative(self, obj_ref, nb, cur_user):
                 meta[u'job_ids'][u'apps'] = list()
             if 'job_usage' not in meta[u'job_ids']:
                 meta[u'job_ids'][u'job_usage'] = {u'queue_time': 0, u'run_time': 0}
+            meta[u'is_temporary'] = 'false'
```
MrCreosote: OK, so according to bill in this case, as opposed to the narrative service, Untitled is a fine name for a non-temporary narrative. Saving a narrative here implies that 1) the user has changed the name or 2) the user has run an app. Hence the narrative is no longer empty and not temporary.

Effectively, once the narrative service creates an empty narrative, the front end cannot save temporary narratives.

PR #1229, opened 2018-01-09
SCT-291 add LIST endpoint helper func

Tianhao-Gu:


Tianhao-Gu:
@briehl Adding LIST endpoint staging service helper function. Just try to see how it will look like in narrative. Then I will add more endpoint helpers. 

I would want to add some unittest. But I cannot get narrative test running successfully. Maybe we can sync up on this. :)

src/biokbase/narrative/staging/helper.py
```diff
@@ -0,0 +1,121 @@
...
+
+"""
+KBase staging.helper class
+"""
+__author__ = "Tianhao Gu <tgu@anl.gov>"
+
+
+class Helper(object):
+
+    def __token(self):
```
briehl: There's an auth client you can use that'll provide this for you - it's in biokbase.auth, and you can invoke with:

```import biokbase.auth
token = biokbase.auth.get_auth_token()
```

briehl: This'll return None if there's not token, or the current token otherwise.

src/biokbase/narrative/staging/helper.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+
+        token = os.environ.get('KB_AUTH_TOKEN')
+
+        if not token:
+            raise ValueError('Cannot retrieve auth token')
+
+        return token
+
+    def __staging_url(self):
```
briehl: URLs should be added to `src/config.json`

They can be fetched with the `biokbase.common.url_config` class as:
```from biokbase.narrative.common.url_config import URLS
my_url = URLS.key_from_config_file
```

Actually, I just checked and the staging service is already there. You can fetch that url with:
```
staging_url = URLS.staging_api_url
```

It will automatically know what environment you're in, too.

src/biokbase/narrative/staging/helper.py
Tianhao-Gu: 👍

src/biokbase/narrative/staging/helper.py
Tianhao-Gu: 👍

Tianhao-Gu:
@briehl  Thanks! Updated to use biokbase fetch env variables.

PR #1231, opened 2018-01-11
Update to use jgi-metadata method [SCT-284]

thomasoniii:
We now have a jgi-metadata endpoint to vend back jgi-metadata files.

So the logic is the same as before - when the expanded view on a file is opened, it attempts to load jgi metadata. If any is found, it is tossed into a new tab, now named "JGI Metadata" instead of just "Metadata".

While I was at it, I slowed down the load by a fraction of a second but got rid of the "JGI Metadata" tab blinking into view after the other tabs. I felt this was the lesser of the evils.

IMPORTANT NOTE - this is explicitly *not* backwards compatible with older styles of storing jgi metadata. Previously, the importer would create a second file with a .metadata extension (foo.jgi.fasta and foo.jgi.metadata), and the frontend would look for that file. It doesn't do that any more. So any older imports would not get the JGI tab automatically. A user could still get at the metadata (though not pretty-printed) by opening the extended view on the metadata file, looking at the first 10 lines, and copying and pasting.

Newer files (from the last 22 days onward, at least) function properly.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
@@ -328,11 +328,6 @@ define([
               filePath += '/';
           }
 
-          // we need to chop up the file to see if a metadata file exists. Assume that the first part of the file name is the ID
-          // and that it ends in .metadata. Route it into the appropriate subfolder.
-          var fileParts = fileData.name.split('.');
-          var metaDataFilePath = filePath + fileParts[0] + '.metadata';
-
           filePath += fileData.name;
 
           this.stagingServiceClient.metadata({ path : filePath }).then( function(dataString, status, xhr) {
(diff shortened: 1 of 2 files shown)


kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -401,33 +404,35 @@ define([
...
+              );
+
+              // finally, empty and append the tabs container.
+              $tabsDiv.empty();
+              $tabsDiv.append($tabsContainer);
+            })
+            .fail(function (xhr) {
+              // if we failed, then there's no JGI metadata. That's fine. We still want to empty and append the tabs container.
+              // yes, yes, I could abstract this out into a separate function call so the code is DRY and not WET, but it's a duplicate
+              // of two lines that are right next to each other. I felt this was easier.
```
briehl: You could also put those in a `.always` or `.finally` (I forget if this is a Promise or a jQuery deferred...) so it's always done at the end.

But it is nice to have a separate, explicit `.fail` block.

PR #1235, opened 2018-01-13
Struct to image

JamesJeffryes:
Some minor tweaks to the modeling viewers:
1. If there is no enzyme data for a reaction, the entire viewer won't crash
2. Compound names in reactions will link to their respective compound pages


kbase-extension/static/kbase/js/widgets/function_output/modeling/KBaseFBA.FBAModel.js
```diff
@@ -242,6 +242,7 @@ function KBaseFBA_FBAModel(modeltabs) {
 
     this.ReactionTab = function (info) {
         var rxn = self.rxnhash[info.id];
+        console.log(rxn)
```
briehl: probably isn't necessary, outside of testing.

PR #1236, opened 2018-01-17
SCT-291 add more helper functions

Tianhao-Gu:


Tianhao-Gu:
@briehl Adding couple more helpers. The Helper class should now cover pretty much everything the current staging_service can do.

PR #1242, opened 2018-01-25
General Cleanup

briehl:
This PR will have some general cleanup tasks and minor bugfixes to get merged up to next and beyond.

- [x] PTV-225 (not sure where else to reference) - change data sort icons to be more descriptive when sorting by name or type (A->Z icon).
- [x] PTV-886 - no scrollbar in the Narratives panel in the Narrative interface.
- bump version (probably minor version bump to reflect shiny new Import tools)
- add release notes


briehl:
There will probably be a few more tickets over the course of the day. I'll add these as I go.

briehl:
Agreed. If that grows in the future, we should rewrite that whole sorting feature to be more easily expandable. Thanks for the feedback!

PR #1248, opened 2018-01-31
Updating phytozome genomes in public data configuration

samseaver:


samseaver:
@briehl Can you review this for me? There were also some duplicate entries under prod

PR #1249, opened 2018-02-02
App panel cats [SCT-531]

thomasoniii:
Originally PTV-840.

This pulls in the categories.yml file defined in kbase-ui-plugin-catalog (now a bower dependency) and uses it to display the same category names as are shown on the app catalog page.

I opted to keep the sort alphabetical (by mapped name, natch) instead of using the ordering provided in the categories file, but we can easily change that if desired.


briehl:
I can't seem to get this to compile. It's trying to resolve multiple versions of js-yaml and zonking out during an automated install. Messing around locally, I think I made it work with js-yaml 3.3.1 (that's what require-yaml wants).

thomasoniii:
Yep, explicitly setting to 3.3.1 seems fine. I updated the bower.json file to just wire it to that version.

PR #1250, opened 2018-02-03
SCT-401 Inject the correct auth token on kernel restart

briehl:
It was using a somewhat hidden, cached copy instead of straight from the browser cookie. Now, when the kernel restarts, it should do so with the browser's cookie, which should be up to date.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,15 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v4.4.1 (more notes will follow).
 
+### Version 3.6.1
+- SCT-533 - Remove the accidental test uploader that crept into production.
+- SCT-516 - Set staging panel to auto-refresh after various updates.
+- SCT-531 - Updated App Panel to have the same category names as the external App Catalog.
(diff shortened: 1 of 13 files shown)


briehl:
I don't think this will completely solve SCT-401, but it might help some variants.

briehl:
It doesn't really monitor whose token that is at any given time, as long as it's valid. The "switch" is unchanged - when the kernel lets the browser know it's been restarted, the browser injects the token. Just where it's getting that token from is what's changed.

So, yeah, it could be invalid, or belong to a different user. If it's invalid, the watcher will just trigger a logout/redirect. If the user's changed, well, that's different, and will cause a number of different issues. Probably out of scope here, though.

This is not terribly likely to happen in the wild anyway. A user would have to have a running narrataive, log out from a different window (which would redirect the running narrative back to the login page), log back in, then refresh the kernel in their original narrative (NOT refresh the page). That's something I do occasionally when I'm running locally, or when I'm trying crazy things to reproduce the SCT-401 problem. I still feel that's something deeper in the system, however.

briehl:
This now also bumps the version to 3.6.1. Forgot I was on this branch...
Still, this should get merged anyway.

PR #1251, opened 2018-02-06
added auto-refreshing of staging panel [SCT-516]

thomasoniii:
This pull does a couple of things -

* kbaseNarrativeDataPanel and kbaseNarrativeStagingTab are updated to allow a widget on the panel to have an activate() / deactivate() method to be called when the panel they are on respectively comes into view / leaves view. These methods are not required to be defined. But they're generic in nature, so that if we want a tab to do extra processing when it comes to the foreground, it can.

There is a gotcha in that an active widget will not be told to deactivate() if the panel is closed by any means other than hitting the toggle button. This should be re-visited.

* stagingAreaViewer now auto-refreshes its view every 30 seconds. The timer will start when the widget has been sent an activate() and stopped when the user sends it a deactivate(). Per the above listed gotcha, if the uploader tab is active, and the data panel is closed via any means other than hitting the toggle button, it will continue to refresh needlessly in the background.

The rest of the mods in the stagingAreaViewer are just to ensure that when the auto-refresh occurs, that any detailed metadata views of the files remain open, and that the tab is scrolled to the same point the user was at. Sometimes there is a blink for a fraction of a second as it redraws, but not always. It's very minor, and I worry that eliminating it entirely would provoke major effort.

It may be reasonable to move the code to append the open metadata on files to outside data tables' row callback, but for now I'm leaving that as an exercise to the reader.

If there was already a method in place for noting when a tab becomes active/inactive, I missed it. If that's the case, please let me know and I'll rip this out and refactor to use the existing method.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataPanel.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataPanel.js
@@ -201,6 +201,11 @@ define([
                     this.$slideoutBtn.tooltip('hide');
                     this.trigger('hideGalleryPanelOverlay.Narrative');
                     this.trigger('toggleSidePanelOverlay.Narrative', this.$overlayPanel);
+
+                    // NOTE - this will be missed and a widget will remain active if the panel is closed by means other than clicking this button.
+                    // This should be re-visited at some point.
+                    this.deactivateLastRenderedPanel();
+
                     //once we've clicked it 10 times, meaning we've open and shut the browser 5x, we reveal its TRUE NAME.
                     if (++numDataBrowserClicks >= 10) {
                       this.$slideoutBtn.attr('data-original-title', 'Hide / Show Slidey McSliderface')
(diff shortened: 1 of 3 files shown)


briehl:
I don't think there's an event that occurs when an app becomes active/inactive. There are a couple events associated with the slideout coming and going, though they're mostly used to control that action. We should probably add a few that get triggered after it's done appearing or disappearing, and could be used for control over what happens in that panel. 

The current ones happen in kbaseNarrativeSidePanel.js - 'showSidePanelOverlay.Narrative' and 'hideSidePanelOverlay.Narrative'

briehl:
Other than that, this all seems reasonable.

It doesn't have to be part of this work, but in the near future, I think we should fix some of the wiring issues with the slideout and how events get managed around it. Right now, the toggled arrow icon in the data panel only flips (-> to <-) if the slideout gets triggered using that button. Clicking outside of the slideout also removes it, but doesn't change the arrow, Neither does the red "Add Data" button in a new Narrative. And the arrow in the App panel doesn't move at all.

I think those could all be related to having finer control over interior panel states as well.

PR #1258, opened 2018-02-15
Fix problem with opening read-only narratives

briehl:
That's a fragile piece of code. :P


briehl:
Travis build doesn't want to start. gonna merge anyway and test on ci. tests pass locally.

PR #1260, opened 2018-02-15
remove "search data" from menu in Narratives

nlharris:
(it links to the old search)
as per https://kbase-jira.atlassian.net/browse/PTV-958


nlharris:
requesting review by @sychan (or whoever)

Issue #1266, opened 2018-03-06
Need to force tornado version under 5 in requirements.txt or change event loop code


bio-boris:
On a fresh install of the narrative repo, the requirements.txt has a tornado requirement for tornado > 4
However, the latest tornado, v5, doesn't have io_loop anymore in the init.
You get the following error when running narratives

`   TypeError: __init__() got an unexpected keyword argument 'io_loop'`

The workaround is to change the requirements.txt to a lower version of tornado, such as
tornado==4.5.2

Some discussion of this I found online:
https://github.com/bokeh/bokeh/issues/7308

```
Uncaught exception in /narrative/api/kernels/c2416ccc-02db-4467-8cc7-fe9871fc46b2/channels
    Traceback (most recent call last):
      File "/workspace/narrative/narrative/narr-venv/lib/python2.7/site-packages/tornado/websocket.py", line 498, in _run_callback
        result = callback(*args, **kwargs)
      File "workspace/narrative/narrative/narr-venv/jupyter_notebook/notebook/services/kernels/handlers.py", line 260, in open
        super(ZMQChannelsHandler, self).open()
      File "workspace/narrative/narrative/narr-venv/jupyter_notebook/notebook/base/zmqhandlers.py", line 176, in open
        self.send_ping, self.ping_interval, io_loop=loop,

```

briehl:
Thanks for finding this, and for the PR!

PR #1267, opened 2018-03-06
Update requirements.txt

bio-boris:
This is the latest version of Tornado before Tornado5. See https://github.com/kbase/narrative/issues/1266


briehl:
The test error here is a red herring - that's an issue with the test itself that's fixed in #1265

PR #1268, opened 2018-03-08
SCT-628: Add a viewer widget for CompoundSet

JamesJeffryes:


JamesJeffryes:
@briehl So i'm not familiar with how the Narrative testing works. I added a test spec for the new type but that's still showing a lack of coverage on the new lines

briehl:
Sorry, I didn't see this until now. This is fine, for now, I think, at least loading the module into memory to see if it crashes. It would be nice to get some better testing, but I think there needs to be some better architecture in to support it - hopefully when the Narrative gets integrated with mini kbase, this'll become easier. 

The FBA modules are also tricky, since they are all loaded in a different context from the other widgets.

PR #1269, opened 2018-03-14
SCT-559 App cell "Bus Error" on page start

briehl:
There's a bunch of chatter on that ticket that I won't put here. But I want to get a PR set up anyway.

This addresses a number of race conditions that pop up with a slow connection and/or a slow computer. It also upgrades the Jupyter notebook to 5.4.0 (ignore the branch name...) and IPython 5.5.0.

**Things needed to do:**
- [x] style fixes
  - [x] icons
    - [x] chevron icons in left side panels
    - [x] refresh icon in the same
    - [x] chevron icons in app panel category mode
    - [x] refresh icon in data slideout
    - [x] need to have cursor:pointer in various cards (apps, data)
    - [x] favicon back to KBase style (I think that's fixed - there's some new JS in Jupyter that changes it on the fly, though. needs testing.)
  - [x] "my favorites" -> "My Favorites"
  - [x] extra scrollbar on right side attached to header
  - [x] multiple versions of font awesome icons
  - [x] vertical blue bar on selected cells
- [x] Fix cell minimization
- [x] Adjust all nbextensions so they shouldn't loop over all cells multiple times - should only scan cells after the notebook has been fully loaded (by the jquery event "notebook_loaded.Notebook" and the _fully_loaded flag)
- [x] Fix tests here
- [x] Tests for app cell load race condition fix
- Smoke tests for various nbextensions -- already present, browser vs. CPU vs. internet race conditions are hard to catch...
- [x] Update narrbase image to use new versions of jupyter products
- [x] Bump version to 3.7.0
- [x] Add release notes
- [x] Fix missing glyphicons for datatables viewers
- [x] Remove persistent "Trusted" notification

**Stretch goals:**
(reserved for other PRs after this one)
- [x] Do something about long narrative names in header causing overlap (#1270)
- [ ] Batch auth calls for username lookup for shared narratives
- [ ] Make service to provide app info to app panel and app cells. Shouldn't have to make a zillion Catalog/NMS calls on startup.


briehl:
Done and ready for review, and possibly testing on CI.
The stretch goals should be factored into (or associated with) JIRA tickets, and addressed as individual PRs. This one's already big enough.

briehl:
Another note - all the widgets being touched here were tweaked to remove some parallel/redundant dependencies. They all depended on both Datatables and the Bootstrap wrapper around Datatables. Just including datatables-bootstrap would include the other one, so the redundancy was removed.

briehl:
There's a lot of files updated here, but they're in semi-related chunks.
1. all of the nbextension/***Cell/main.js files needed an update to do the majority of addressing the race condition bug.
2. a large number of viewer widgets (~25, mostly in the function_output directory) were updated to remove a redundant import of datatables. This is mostly optical, but it's a better and more sustainable practice.
3. several of the core page widgets (the App Panel, the Data Panel, etc.) were updated to migrate from using glyphicons -> Font Awesome icons in an attempt to be consistent, and just use Jupyter's install of Bootstrap that doesn't include the glyphicons font pack
4. ...which kinda failed because DataTables didn't want to work. So an updated version of Bootstrap (mainly to include the Glyphicons pack) was included.
5. page.html and notebook.html were updated to support the new Jupyter version, as were any files ending in .py
6. tests were all updated to include needed dependencies from the new Jupyter version.
7. the usual extra files (package.json, bower.json. biokbase.narrative.__init__.py, config.json) were updated to show that we're on Narrative 3.7.0

In all, the majority of the changes were in each of the nbextensions' main.js to ensure that the notebook was fully loaded before updating all cells. A few other internal API pieces were updated to also fight race conditions. The rest (so, the majority of the changed files) were updates so we can include the newest version of Jupyter Notebook.

eapearson:
Here is an issue I see in this PR and not prod, some tables are awry. See the screenshots. This example is for the data view cell for a genome. There is one screenshot from prod, two from CI -- the one that looks normal is straight from CI, the malformed one is rendered locally from the PR. YMMV?

screen shot 2018-03-20 at 2 22 50 pm
screen shot 2018-03-20 at 2 22 12 pm
screen shot 2018-03-20 at 2 22 05 pm

eapearson:
YMMV meaning that if you don't see this, it may be an artifact of my local deploy, which works fine most of the time, other than Docker going into slow decay and forcing a (Docker) restart periodically.

briehl:
No, I see it too, now. I hate datatables. Have I said that recently? Because I do.

briehl:
I take it back, that might not just be a datatables thing. Investigating.

eapearson:
I hadn't even looked at the console.

briehl:
Found it. A sneaky CSS rule that sets all tables to table-layout: fixed. Making a patch. There's another rule there that sets all tables to text-align: right. I don't know why they would do that.

eapearson:
cool.
are you seeing these missing source map files as well? 

screen shot 2018-03-20 at 3 02 29 pm

eapearson:
oh, and preact? Oh, Jupiter, why not be different? I guess it is a new available dependency? I know you are thinking about that...

briehl:
Yeah, I'm seeing those. I don't think the maps are included in the Jupyter install. Running vanilla Jupyter will cause the same issues, unless I have some kind of funky environment.

They have one (relatively new) module that does optional things. I think it's called shortcuteditor, and it uses preact and the others you see there. And they don't load via AMD. And they depend on a specific load order. And they're not really in a canonical location.

Those were great fun to get working with our stack...

briehl:
I'm hopeful that Jupyterlab really will be easier to extend and modify, once we get time to explore that.

briehl:
For the record, as much as I'd like to explore other view layer things like React, I'm certainly not throwing anything like that in here. :)

eapearson:
is there a way to reliably trigger the race condition, say a heavy narrative in CI?

eapearson:
So I've opened a bunch of the largest public narratives I could find in CI, and they load fine.
In fact, since running the narrative locally the last couple of weeks, I have experienced occasional collapses of docker; narratives would start to be slow or fail to load (no app cell errors, though, once started), the ui would get fidgety and sometimes a url here or there would timeout. Bad behavior was associated with connections > 50 or so, with failure setting in above 150 (just doing a simple lsof -p $(pgrep vpnkit) | wc -l). Haven't had time to diagnose much, but early googling had led to to a docker/macos issue with dangling connections which eventually choke docker. 

But Narrative was also involved because killing the narrative container would also sometimes (but not always) clear it up.

But now ... throwing 10 big narratives at the same time, no problems at all. A few narratives got stuck on the loader page for about 10-20 seconds, but they eventually loaded fine.

briehl:
If you're using Chrome, there's a dev option (in the Performance tab) to throttle the CPU. Setting that to 6x slowdown would reliably cause that problem, along with other related ones sometimes - ever see the "Narrative doesn't have a workspace ref" problem? That's related, too.

I'm not sure if there's similar options in Firefox or Safari.

briehl:
Huh, no idea what could have caused that, but I'm glad that this PR clears it up. Might have been the older Jupyter notebook. Might have been inconsistent interactions among cell extensions all vying for CPU cycles. Shrug.

eapearson:
Scanned changes top top bottom, nothing stands out as far as I can grok.
Using the narrative from this PR it operates normally, as I commented above. Now however I noticed in the console the following error (chrome):

```
utils.js?v=20180321221210:963 Uncaught TypeError: Cannot read property 'resetEquationNumbers' of undefined
    at utils.js?v=20180321221210:963
    at Function.execute (MathJax.js?config=&delayStartupUntil=configured:19)
    at cb (MathJax.js?config=&delayStartupUntil=configured:19)
    at Object.Process (MathJax.js?config=&delayStartupUntil=configured:19)
    at Object.call (MathJax.js?config=&delayStartupUntil=configured:19)
    at Function.WAITEXECUTE (MathJax.js?config=&delayStartupUntil=configured:19)
    at cb (MathJax.js?config=&delayStartupUntil=configured:19)
    at Object.Execute (MathJax.js?config=&delayStartupUntil=configured:19)
    at Object.loadComplete (MathJax.js?config=&delayStartupUntil=configured:19)
    at AMSmath.js?rev=2.6.0:19
```

Where utils.js is https://ci.kbase.us/narrative/static/base/js/utils.js?v=20180321221210

and the problematic call is:

screen shot 2018-03-21 at 3 29 12 pm

briehl:
I've seen that, too, then it went away. Though that may have been on a different branch.

eapearson:
Other than that, it looks fine to me.

briehl:
That particular call resets the equation numbers generated by TeX for those who use it.
I'm not sure if this is a side-effect, but LaTeX isn't rendering. Might be caused by that error making MathJax crash out?

eapearson:
The function that is failing is indeed "typeset", but I don't know now the mathjax queue system works, but I would hope that the failure of one queue item would not disrupt the entire mathjax processing queue? That line is wrapped in a try ... catch but the actual error is thrown when the queue item is executed...

In any case, I can replicate, I'm not getting latex rendering in a Narrative with just a single markdown cell.

briehl:
The MathJax config template variable isn't getting passed in properly. Not sure why. Putting the default value ("TeX-AMS-MML_HTMLorMML-full,Safe") in the notebook.html template fixes the problem, so we can always fall back on that. I'd rather have it work how it's supposed to, though.

briehl:
Ok, merging. Thanks for testing!

PR #1270, opened 2018-03-21
SCT-690 - fix narrative navbar with long name

briehl:
Long narrative names cause the navbar to hang over the rest of the app and prevent seeing the tops of cells and such. Other apps (including Jupyter notebook) hide long names with an ellipsis. This PR does the same, but mousing over the truncated name will temporarily show the whole thing. Gif below.

out


scanon:
Best use of animated gif goes to Bill

briehl:
Oops, forgot to mention that #1269 should be merged first.

briehl:
Sheesh, I thought merging all the other changes in that other PR would shrink this one. I guess not! Closing and re-opening as it's own.

PR #1271, opened 2018-03-21
SCT-590 cache busting for data_source_config.json

briehl:
It doesn't happen often, but when the data source config (the config file for updating the sources for Public data in the data slideout) gets changed, the Narrative tends to not pick it up as it's cached in the browser.

This adds an explicit cache busting call, instead of just being lazy and using $.getJSON.

Also, this expects #1272 to be merged first.


briehl:
This is pretty minor, going to just merge.

PR #1273, opened 2018-03-23
SCT-706 - fix app panel search not accepting spaces

briehl:
The fix was reasonably simple, but playing around exposed that the search boxes in the app panel and data panel were different. They (more or less) should act the same, but looked slightly different. To avoid keeping duplicate code around, this factors that search box out into a generic, though specialized, widget. It might be a little much, but it keeps things consistent.

The widget - BootstrapSearch - makes a simple Bootstrap input form with an attached clear button. Options sent to it on initialization set its functionality, mainly what to do when a user types.

This also includes some whitespace and style cleanup to the App panel.


briehl:
Great. Thanks!

PR #1275, opened 2018-03-29
Sorta genome browser widget [SCT-670]

thomasoniii:
Well, I bungled this one.

Anyway, attached is a very very minimal little modification of a narrative cell - all it does is show a Genome Browser tab (and only on CI, for security purposes) that just displays a genome browser hardwired to some rhodobacter data, since I haven't gotten a dynamic service up to vend out the data as of yet. In short, there's really nothing here.

It's solely just "here's a small genome browser with a gif track, hardwired to a random genome, that only exists on CI." Hopefully someone can look at it and come up with some useful ideas about something to do with it.

Things that I'm taking away from it -
* Supposedly JBrowse can be configured to load raw files, I haven't found the proper incantation to do it yet, so until then IGV is here.
* BUT - as I dug more into this, I discovered just gigantic fasta files in our data, in the neighborhood of 100+ megs. Loading those into IGV takes a minute or more, so it's not user friendly at all. We need a different way to load the data in, which presumably is what JBrowse's interim track conversion process would do for us.
* This minimal implementation only includes annotations from a gff file and nothing else. Someone needs to define what info it should present, and where it's loaded from. Taking into account file size as listed above.
* Tacking this thing into a landing page is straightforward, once the data is vended out in some sort of useful manner.
* The dynamic service should be straightforward for someone with more knowledge of building one, and who knows python. Between Bill's original kb_GenomeBrowser app and the stuff I pulled out and ported to perl (which no one wants), the bones of a service are in place - but it needs the final wrapping into something restful that can be loaded via a GET to hand into a browser. Something similar could be done to load the individual json tracks into jbrowse.
* This shouldn't be that difficult to do as a concept, I just wasn't able to figure it all out in a month.


thomasoniii:
If there's a better place to stick the arbitrary fasta/gff file I hardwired it to instead of hosting it externally, just lemme know and I'll bump the URLs. I just didn't have a better place to stick it and didn't want to commit it into the repo here.

sychan:
Do you have a screen shot anywhere?

briehl:
👍

PR #1277, opened 2018-03-30
remove old import tab, remove (New) tags from new one

briehl:


briehl:
Passed before updating release notes, going ahead and merging.

PR #1280, opened 2018-03-30
update npm in the dockerfile

briehl:


briehl:
This won't have any effect on Travis-CI tests, only when deploying on CI. Merging anyway.

PR #1284, opened 2018-04-04
SCT-396 Widget test bed for copying/sharing problem

briehl:
This is somewhat incomplete, so hopefully someone else can pick it up. It'll likely crash right now on test/unit/run_tests.py, maybe.

I wanted to get this in place to be something helpful while I'm out.

The main parts of the widget test bed are under test/casper. You can run it (though it might fail) with `make test-casper`. Or, make sure you have a running Narrative, and run `casperjs test --includes=<path_to_venv>/jupyter_notebook/notebook/tests/util.js test/casper/name_of_file.js` smoketest.js is a good one to start with.

There's some incomplete things here right now, so this probably won't be useful until it works.
My original goal was to have all of the workspace sharing and modification things written in JS, run as part of the test harness. Turns out, it's a huge pain to deal with asynchrony in running tests in Casper, especially doing non-testing things. So I decided to do it all through test/casper/run_tests.py - that would run the widget testing part, then shut down the notebook, then use the Narrative's own Python code to manage copying/sharing, and finally do the second half of the testing to verify that shared things still work right.

You should still be able to run widget tests now. The genome widget test is in `test/casper/widgets/kbaseGenomeView.js`, which makes use of `test/casper/WidgetTestBed.js`. So check those out, and modify as needed.


briehl:
https://kbase-jira.atlassian.net/browse/SCT-396

As detailed above, the flow for doing sharing tests will also involve some magic automation. The main script that controls this is `test/casper/run_tests.py`, which (currently) handles running each individual headless browser test. All it does right now, is use some configured info to open a given narrative as a specific user, click on a data object in that narrative to add a viewer widget, then inspect that widget to make sure it's hunky-dory.

The next phase here is to expand that script (and config) to do the following. I've split out the tasks between `Py` for the run_tests.py script and `JS` for the CasperJS testing script:
1. Change config (for each widget) to include public data (from CI for now. ideally, it would include a data object to get uploaded into mini-kbase each time... but we'll get there.)
2. `Py` Create a new narrative and workspace for user A.
3. `Py` Copy public data into user A's workspace.
4. `JS` Do the completed test steps, but save the narrative this time (simulate clicking on the Save button).
5. `Py` Share that workspace with user B.
6. `Py` User B copies that narrative / workspace.
7. `Py` User A revokes share privilege.
8. `JS` User B opens that narrative and inspects the already-made widget cell to see that it a) renders right and b) references the copied data object.

briehl:
This finally all works. Huzzah!
There's been a few changes since last time.
1. Changed the test command from `make test-frontend-casper` to `make test-widgets`. Also added a variable to only test one widget - `make WIDGET=kbaseGenomeView test-widgets` will work, for example. The widget has to show up in `test/integration/testConfig.json` for that to work.
2. Renamed test/casper -> test/integration to make it a little more agnostic.
3. Got all the steps in the post above to work.
4. Wrote all the documentation I could think of.

briehl:
Ugh. PhantomJS is dead. Long live any other headless browser for testing!

Closing this for now. Needs some serious reworking.

PR #1285, opened 2018-04-04
update readme with authenticated test info

briehl:


briehl:
also has a fix for a dependency not being loaded.

PR #1287, opened 2018-04-09
Fix broken pangenome viewer [SCT-793]

eapearson:
The problem was a unconstrained dependency on the "file-saver" library. The "^" operator causes the narrative build process to pick up the most recent version of the library; a change in 1.3.6 (current version is 1.3.8) broke the library by, um, omitting the javascript file which implements the library.


scanon:
Jim said this looks good and I can't see any issues.  Merging.

PR #1288, opened 2018-04-09
prevent successfully completed app from re-inserting deleted output cell upon reload [SCT-791]

eapearson:
- also some code cleanup and removal of long-unused features
- note that the narrative must be saved after the output cell is deleted

### DIFF
--- a/nbextensions/appCell2/widgets/appCellWidget.js
+++ b/nbextensions/appCell2/widgets/appCellWidget.js
@@ -71,12 +71,7 @@ define([
         div = t('div'),
         span = t('span'),
         a = t('a'),
-        table = t('table'),
-        tr = t('tr'),
-        th = t('th'),
-        td = t('td'),
         pre = t('pre'),
-        input = t('input'),
         p = t('p'),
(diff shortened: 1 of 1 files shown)


eapearson:
Hi @thomasoniii I added you as a reviewer so we could move this one along too. It is an important usability fix especially important for large narratives like the ones Miriam is working on now, because it prevents output cells from popping back into place after being deleted.

I'm going to spend the rest of the afternoon getting a handle on narrative testing, see if I can add a test for the various cases of output cell behavior.

BTW in case you don't use it enough to remember, to get rid of the whitespace formatting noise (it was too late when I noticed it) you can add ?w=1 to the end of the url when viewing the diff and it will remove all just-whitespace diffs.

eapearson:
Hi @thomasoniii would you mind reviewing this? Then let me know and I'll ask Shane or Steve to merge it.

thomasoniii:
Looks reasonable to me.

eapearson:
@sychan or @briehl would you mind having a look and merging?

PR #1289, opened 2018-04-10
prepare release 3.7.1 [SCT-793]

eapearson:
- update semver in all places that 3.7.0 exists
- and added release notes for 3.7.1

Able to test this locally against CI; still working to test it against prod locally.


scanon:
Hope this works.

PR #1291, opened 2018-04-12
Fix FeatureValues widgets [SCT-804]

eapearson:
Fix the primary bug with these widgets, which was a mistake in the AMD module dependencies.
Along the way of course I needed to repair the code to satisfy lint.
Some programming errors were fixed as well, most notably global module references.
Also hard-coded usage of the 'dev' tag for calls to FeatureValues service were removed.

I'd appreciate a review. I'll be setting up a testing Narrative in CI which can demonstrate the proper operation of of these widgets. It is not necessarily as trivial as it might be because some objects involved suffer from the copy problem and need to be generated from scratch within the narrative.


eapearson:
Change highlights:

- The main culprit is that in kbaseExpressionConditionsetBaseWidget.js and kbaseExpressionGenesetBaseWidget.js  the 'jquery-dataTables' module id was introduced before 'kbaseFeatureValues-client-api' without introducing an argument to the module's function to receive it. Since dataTables is actually a jQuery plugin and does not define any useful module objects, it should be placed at the end of the dependencies and not be associated with any module argument.
- The "Workspace" service client was being referenced in more than one module as a global reference -- which is strongly frowned upon. These were replaced usages of generic client.
- Other non-promise usages of the narrative's generic client were replaced with usages of abase-common-js's promises-only generic client.
- There were many (hundreds) of violations of eslint rules. Purely to enable me to sanity check the modules, all linting violations needed to be cleaned up.
- All affected modules were set to strict mode, and strict mode enabled in the eslint rules file (which works for Visual Studio Code, don't know about other environments.)
- since FeatureValues is a dynamic service, no longer a core service, all references (in affected files) to the core service, including the config.json file, were removed. This reduces code and human confusion.

scanon:
This looks good as far as I can tell.  I can merge it if needed.

eapearson:
That would be great @scanon, from my point of view, since these are broken in production.

PR #1292, opened 2018-04-16
SCT 657: narrative data panel public tab NCBI RefSeq Genomes should use KBaseSearchEngine as data source

eapearson:
This set of changes converts the "NCBI RefSeq Genomes" item in the "Public" tab of the Data Slideout from using the old SOLR search service to the new KBaseSearchEngine service.

The changes include not just the change in data source code, but also in the search input and the search layout.

The search input was changed to require the user the press the Enter/Return key or the new Search button in order to conduct the search. This replaced a keypress-based search in which user entry would passively cause a search to be conducted based on partial user input. This change is required for two primary reasons. First, the new search service can experience higher latencies than the old one and conducting the search based on partial input, as the user types, could cause searches to easily overlap, pauses in typing, and overall sluggish responsiveness. In addition, the new search service does not perform partial matches, so a partial search term would likely and confusingly provide either no results are irrelevant (to the fully typed search) results.

The second change was to modify the search layout to use an embedded table for the small "metadata" detail contained within the results. The previous layout was ragged and difficult to read. In addition the new layout, for the refseq data only, breaks out the kbase id and refseq id into two lines. They were combined in the previous layout, leading to one very long long which could be difficult for the user to read.

A later set of changes, perhaps in a separate PR, may contain integration testing. The framework for these tests has been started, but is not included in this PR yet because they do not yet include the test scripts for the search scenario, which is rather complex and will take some time to set up.


scanon:
I would like Bill to review this later.  But I'm going to merge so we can test this on CI.  Bill may have some style comments later.

PR #1293, opened 2018-04-17
Sct 762

JamesJeffryes:
This PR converts the feature set viewer to use GenomeSearchUtil to get feature data. This abstracts any changes to the genome object allowing features from all feature arrays to be searched.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/kbaseFeatureSet.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/kbaseFeatureSet.js
@@ -113,8 +113,6 @@ define (
         },
 
         search: function(genome_ref, query, limit) {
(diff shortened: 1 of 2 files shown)


thomasoniii:
I'd recommend getting rid of the various informational console.log() lines. They just clutter things up in production.

Especially on the promise returned from the search function around line 125 - it's got a then and a fail that do nothing other than log the info. I'm less concerned about the fail() catch - I'd prefer that there's a catch on the call to search() that logs it (instead of inside the search method), but I can see the utility of just having one centralized. But the then() is superfluous.

And note that the travis build failed - Oracle JDK 8 was not installed, apt-get failed, exited with 100 error (more info in the log). So presumably there's something to resolve there, I just don't know what exactly it is.

PR #1294, opened 2018-04-19
Add "info" item to view cell menu for inspecting app info [SCT-821]

eapearson:
- shows app info for view cells
- some code quality fixes along the way
- general cleanup of viewCell extension
- viewCell wasn't (now is) pinning non-release to a git commit hash
  - removed unused code
  - removed commented out code
  - removed unused files
- run_local_app was not pinning beta & dev to a git commit hash as app cells do. Cannot think of a reason this should be so, and it is unexpected to have different behavior. It can of course also break if a dev or beta changes, which it will.


nbextensions/viewCell/widgets/codeCellRunWidget.js
```diff
@@ -1,170 +0,0 @@
-/*global define*/
```
scanon: Was this all just dead code?

eapearson: These are directives to "jslint", which was the old linter we used, now replaced by eslint, which has a different way of configuring lint options. So in this case the eslint directive "env.amd = true" takes care of this as well as other globals.

We should actually make sure we are all on the same page with this. I use the config file .eslintrc.yml in the project root which integrates as well with Visual Studio Code, but don't know how other folks set this up.

eapearson: I agree about separating code cleanup from functional changes, although it can be tricky to separate them because the cleanup scope expands as the code change scope expands. Still, an initial crack at it would reduce the noise.

Perhaps an initial code cleanup commit which absorbs the bulk of at least the mechanical fixes (e.g. fixing all linting errors, which may be functional as well, but not part of the main logic changes.)

PR #1296, opened 2018-04-20
improve side panel behavior in view-only mode [SCT-866]

eapearson:


eapearson:
@scanon So I went ahead with this since there were also some bugs lurking -- the warning dialogs which should have appeared when attempting to add a data viewer or app in view-only mode were not displaying.

I opted for hiding the app panel in a different way. The app panel is simply collapsed, as if the user had used the collapse control in the app panel header. This naturally allows the data panel to expand to full height. The current, problematic approach used a different method to have the data panel take over that side panel, but this caused at least two different problems -- the 50% height problem and also incorrect behavior when leaving view-only mode.

With this change, the app panel is collapsed when view mode is entered (either a read only narrative or using the view-only toggle button), and is restored when leaving view mode (clicking the view-only toggle again). There is more to it than that, but that is the basic behavior. It should achieve (a) hiding the apps initially and allowing the data panel to expand to the full height (b) using the existing mechanism which, since it is a user control, (c) allows the user to inspect the apps if they want to.

Also, in this PR I've separated the changes into three commits -- the logical changes, code quality changes to these same files, and finally a couple of changes in other unrelated files to ensure that some unwelcome logging lines are removed. Basically, like a good Boy Scout (I was not a good one), leave it better than you found it.

http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule

PR #1311, opened 2018-05-10
Volcano plot updates [SCT-380]

thomasoniii:
Updates to the volcano plot.
* calculates logs of q-values, since those are no longer provided.
* Also adds in the ability for the user to adjust the min/max x/y values. This is rough around the edges, but we want it up on CI for Ranjan/Sunita to start playing with it.

Yes, it redraws the svg in a very lazy manner when the min/max values change, but my d3 is a little rusty and as I said, it's rough around the edges.

Once it's up on CI, we'll probably need to iterate more, but we don't necessarily need to prevent an unintentional release.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/kbaseDifferentialExpressionMatrixSetViewer.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/kbaseDifferentialExpressionMatrixSetViewer.js
@@ -46,22 +46,7 @@ define (
           this._super(options);
 
           var $self = this;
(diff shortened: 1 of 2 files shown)


thomasoniii:
Got it. Silly me had typed "renderSVG() {...}" instead of "renderSVG : function() {...}"

These old syntax bugs are just impossible to track down, since they (a) run and (b) sure do look correct.

Anyway, I don't know if we have time to get this onto CI before the demo today, but if we could that'd be awesome.

PR #1314, opened 2018-05-17
SCT-962, SCT-975 batch backend

briehl:
This adds a `run_app_batch` function to the appmanager. Instead of taking a single dictionary of inputs, it takes a list of dictionaries. These get built out (based on the app their calling), then bundled together in a call to the kb_BatchApp.run_batch app.

This all might change eventually, but it's a place to start from.


eapearson:
Otherwise, looks good to me; sort of understand what is happening. For supporting a batch app, I suppose there will be an addition to the app spec to make this clear to the narrative, something like a flag to enable batch operations, constraints (min, max) on the # of batch submissions, perhaps batching which ranges across one or more parameters, etc.

PR #1315, opened 2018-05-18
SCT-976 batch jobmanager

briehl:
There really isn't much to do for the batch jobmanager, as defined now. But this wires things together to use the mock job checking service. Also there's a couple minor fixes to some frontend things.

So it's temporary, but reasonably small. It does also depend on #1314 getting merged first, though.


eapearson:
Looks good to me.

PR #1320, opened 2018-05-25
[WIP] SCT-961 batch frontend update

briehl:
**DO NOT MERGE**
I'm just making a checklist here of things that should go in this PR.  

This will NOT cover all the changes to make batch apps work. It's mainly focused around the dummy first-pass at setting a batch mode and setting inputs. At the end, a user should be able to enter app inputs (in some form, probably as either a Python variable or textarea input, maybe as a spreadsheet-ish thing), and have the cell generate a call to `run_app_batch`.

This'll need the following:
- [x] Update PythonInterop to generate a call to `AppManager().run_app_batch()` when it sees a list of inputs
- [x] Add a Batch toggle button to the main App Cell.
- [x] Have that Batch toggle turn on batch mode.
- [ ] Toggling back to single input mode should remove any memory of batching stuff, and warn the user about it (if there's inputs).
- [x] Have batch mode show some dummy input thing of some sort as a first pass, then we can iterate.
- [ ] Have the validator work against that dummy input thing and generate the proper code.
- [ ] While running (view only mode), show a view only batch parameter widget.


briehl:
merging what we have to have something for the demo. not finished, but.... eh.

PR #1321, opened 2018-05-25
 add report icon to data panel data object toolbar [SCT-930]

eapearson:


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
```diff
@@ -573,34 +585,76 @@ define([
...
+
+            var workspace = new GenericClient({
+                module: 'Workspace',
+                url: Config.url('workspace'),
+                token: this.token
+            });
+
+            var reportRegex = /^KBaseReport.Report-/;
+
+            return narrativeService.callFunc('list_objects_with_sets', [{
```
briehl: What does this do, exactly? It looks like you're stitching in to the Data Panel populator and looking up everything that references each object, and filtering for reports? That should either just be part of the NarrativeService call. Or, to make it faster, just find the report when the it's requested. In a large narrative, this might take forever.

Also, copied objects (i.e. everything in a copied narrative) lose that link. You can find it by tracing back up the copy chain, but that's also something that shouldn't be applied to every object in a Narrative, and only on request.

briehl: I'm working on a function in NarrativeService right now that should do what you need.

eapearson:
Yes, you got it. I should have added a note, but presumed because of the earlier conversation that you might pick up the NS side of it if you had time, as you have. Otherwise, it works well enough as is to iterate upon.

briehl:
I'll have to pick this up later in the weekend. Got a function mostly there, but not entirely tested. In the meantime, would this work? Think anything else would be helpful?
https://github.com/briehl/NarrativeService/blob/report_lookup/NarrativeService.spec#L393-L423

I figure the service itself should return a list of reports (if there's more than one), and the interface can do with that what it wants.

eapearson:
Took a look at the NS changes finally. Looks like it covers the needs just fine ... 
I'm wondering about including the error within the result itself. Might it be better to include it as a second return value? I've used this pattern, and it makes it very easy to check for the outcome of the call ... either an exception that falls through and generates a 500, or a result, error pair. The presence of a non-null error indicates an error, and the value some error value.
Also, it isn't altogether clear to me how an object can be referenced by more than one report. Could you explain?

briehl:
Hey, before merging this, would you mind looking at (and maybe merging) https://github.com/kbase/narrative/pull/1323 ? I suspect lingering nonsense with PhantomJS is one of the causes of our recent testing woes.

PR #1322, opened 2018-05-30
Clean up logs2

tiramisu24:
-replace log viewer with jobLogViewer Widget
-remove dead code


briehl:
Tests are being very wonky right now and breaking at the end. But this code looks fine, and it works as expected locally.

PR #1328, opened 2018-06-04
Dockerize

sychan:
New dockerize based build of the narrative. Uses the current build/release approach where certain branches are automatically built in Travis if they pass testing, and images are pushed to Dockerhub. The entrypoint is the dockerize program that expands the /src/config.json.templ file into /src/config.json at container startup. The environment variable CONFIG_ENV is used to set the value for the config dictionary value in config.json. The environment variable VERSION_CHECK is used to set the URL to the narrative_version json file. The default is to use the old value, but it can be overridden to point to the new location.

This version also has a Dockerfile2 file that builds a very small ( 15mb ) docker image called "narrative_version" who's sole function is to serve up the narrative_version file. This gets around the need to write this file into the kbase-ui service's directories. This is intended as a stopgap until the narrative is modified to use NMS for for setting/getting the most recent narrative version. By serving the new narrative_version file at a location such as "/services/narrative_version" via nginx, and setting VERSION_CHECK to "/services/narrative_version" when starting narratives, narratives can continue to check for the latest version number.


eapearson:
So far so good. Built locally (takes quite a while the first time as it builds narrabase), started a container, and was able to run with it locally (via the kbase-ui proxy), without problem.

This is how I normally run the Narrative locally.

The only notable issues so far:

- the build ends in a 404, triggered by the final curl command.

```html
<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx/1.12.2</center>
</body>
</html>
```

Also, I use a local container for narrative development and testing, and I think others would be too if it was straightforward. For this usage, I have needed to alter the Dockerfile ever so slightly to comment out the "grunt minify" step. This is necessary to force the Narrative to load all JS files directly from source, so that on docker run the local files can be volume mounted on top of the container files. I usually tweak that by hand when setting up a batch of narrative work, but there could also be an improvement to the Dockerfile to add something like

RUN if [ -n "$SKIP_MINIFY" ]; then echo "skipping minification"; else grunt minify; fi

with a corresponding change in build_narrative_container.sh

--build-arg SKIP_MINIFY=$SKIP_MINIFY \

Although arguably the minification step is part of the narrative build and should be inside install_narrative.sh.

sychan:
The last curl command is a legacy thing - I would be happy to remove it, as it just seems to be noise. I can also look into the skip-minification - it seems straightforward enough. Though I would make a separate make target for a fast local build, instead of monkeying around with environment variables.

Might not be able to get to it until Friday-ish, since I want to get this logstash/docker monitoring stuff off the ground before I switch back to narrative

eapearson:
Works fine for local development so far.
Next is to try testing running tests within the container.

briehl:
Finally got this to build. It all works fine locally, and tests seem to continue to pass in Travis-CI.

I think this is worth merging now. This work was intended to build out the deployment environment to update how we dockerize things. I think local testing within a container and getting it to play nice with the kbase-ui dockerized apparatus should be a different PR.

briehl:
However, there seems to be a merge conflict. @sychan mind fixing when you can?

sychan:
Fixed the merge conflict. Waiting for travis tests to complete before merging

sychan:
Based on Bill and Erik's responses, this looks good to go after fixing the merge conflict.

PR #1335, opened 2018-06-11
Batch log

tiramisu24:
@briehl


.gitignore
```diff
@@ -17,7 +17,10 @@ install.log
 .projectKnowledge
 nbproject
 static/dist
-my_narrative_env/
+my_narrative_venv/
+kbase-extension/
```
briehl: the bulk of the front end code shouldn't be gitignored...

nbextensions/appCell2/widgets/appCellWidget.js
```diff
@@ -682,8 +683,8 @@ define([
 
         function startTab(tabId) {
             var selectedTab = controlBarTabs.tabs[tabId];
-
             if (selectedTab.widgetModule) {
+                debugger;
```
briehl: probably shouldn't check in `debugger` statements

PR #1337, opened 2018-06-14
restore "job_service_mock" to "job_service"

eapearson:
- it looks like this was left in place from development
- job_service_mock causes job queries fail outside of CI


briehl:
Aw, shoot. Yes. Forgot about that. I'll make a similar PR to master.

briehl:
Just to be clear, there's new things in #develop now that shouldn't be pushed to master yet, so we should do a similar PR out of band.

PR #1340, opened 2018-06-19
ConditionSet Changes

JamesJeffryes:
I refactored the condtionset viewer substantially. I also added extensions to datatables to add buttons for download and hiding columns. It would be nice to have them integrated for demos on Fri


bower.json
```diff
@@ -36,7 +36,12 @@
     "bower": "*",
     "install": "^1.0.4",
     "kbase-ui-plugin-catalog": "^1.2.0",
-    "bootstrap": "3.3.7"
+    "bootstrap": "3.3.7",
+    "pdfmake": "^0.1.36",
```
briehl: Is this needed? Are you making PDFs with this widget? It's in the paths, too, but not referenced anywhere else.

kbase-extension/static/kbase/js/widgets/function_output/kbaseConditionSetViewer.js
```diff
@@ -73,29 +75,28 @@ define (
...
-                        // init with optional values
-                        var row = {"unit": "", "unit_ont_ref": "", "unit_ont_id": ""};
-                        for (var key in cs.conditions){
-                            row[key] = cs.conditions[key][index]
+                    var rows = [];
+                    var cols = [{title: "Sample ID"}];
+                    cs.factors.forEach(function(factor){
+                        cols.push({title: factor.factor});
+                    });
+                    for (var _id in cs.conditions) {
```
briehl: In this block, if that's how `rows` is getting initialized, you could use the `map` function instead. It looks like you're converting from key-value pairs to an array of arrays (e.g. from {'foo': 'bar'} to [['foo', 'bar']]). This'd be something like:

```
rows = Object.keys(cs.conditions).map(function(_id) {
    return [_id, cs.conditions[_id]];
});
```

JamesJeffryes: That's not quite what I need to do but there is still a nicer way to write this with map that I switched to

bower.json
```diff
@@ -36,7 +36,9 @@
     "bower": "*",
     "install": "^1.0.4",
     "kbase-ui-plugin-catalog": "^1.2.0",
-    "bootstrap": "3.3.7"
+    "bootstrap": "3.3.7",
+    "datatables.net-bs": "^2.1.1",
+    "datatables.net-buttons-bs": "^1.4.2",
```
briehl: The trailing comma makes the JSON parser break.

PR #1346, opened 2018-06-26
Display a warning text when a typed object has no viewer associated with it.

briehl:


src/biokbase/narrative/widgetmanager.py
```diff
@@ -585,6 +585,8 @@ def show_data_widget(self, upa, title=None, cell_id=None, tag="release"):
                 }
             }
         else:
+            if len(type_spec['view_method_ids']) == 0:
```
JamesJeffryes: What comes back from the service manager if "view_method_ids" is absent from the spec? Do we want to support that case as well?

src/biokbase/narrative/widgetmanager.py
```diff
@@ -585,6 +585,8 @@ def show_data_widget(self, upa, title=None, cell_id=None, tag="release"):
                 }
             }
         else:
+            if not type_spec.get('view_method_ids'):
```
briehl: This should cover the case where 'view_method_ids' doesn't exist, OR when it does and it's an empty array.

PR #1348, opened 2018-06-27
Version 3.8.0 again

briehl:
The previous PR had some issues.

@thomasoniii this has the recent updates to the expression matrix viewer. Is that stable enough?
@eapearson this also has the updates to the regression in the public data tab, and bower.
@sychan this carries the new Dockerize stuff, but I don't think that gets used, right?


sychan:
@briehl @eapearson @kkellerlbl The current dockerized setup is not backwards compatible - the old release process ran things like sed against the config.json file, the new setup requires that an environment variable be set. The default is setup for the CI environment. I was hoping that it would be a quick fix but I haven't been able to focus on getting it done. That being said, I'm currently testing the changes to the Lua code that will enable the environment variable to be set in the nginx proxy container and passed through to the narrative. So hopefully in the next day or two the narrative container can run outside of the CI environment without changes.

thomasoniii:
Yes, the front end for the expression matrix updates is very stable.

It's written so that it'll use the newer methods provided by the ExpressionAPI service, and if that service is non-responsive for any reason, it'll fall back to using the older method instead, and populate those columns with "data not currently available". So it all degrades nicely for the user if there are any backend issues.

PR #1353, opened 2018-06-29
[WIP] Batch frontend work

briehl:
This should *mostly* all be working now. It's hard to tell without being able to test, since the CI queue is clogged (as of 7pm PDT). But I want to get this staged.

It's primarily for the Job Status tab of the app cell, tracking progress of individual jobs.

Things to verify before merging:
- [x] Transition from queued state to any other state should happen seamlessly.
- [x] Child jobs should appear as they start.
- [x] Logs should appear for each job.
- [x] Currently selected job should be highlighted.

Things still to-do (possibly next sprint):
- Wire output for kb_BatchApp so it shows up. --> this will get its own PR.
- [x] Placeholders for each non-started job in the status tab (take advantage of job metadata for count).
- [ ] Improve tests.


briehl:
This is getting a little long and crazy, and I'd like to get it reviewed before adding any more features (and/or bugs...)

@eapearson or @tiramisu24 are probably closest to this code - mind having a look?

kbase-extension/static/kbase/js/util/jobLogViewer.js
```diff
@@ -1108,14 +1111,12 @@ define([
 
         function stopJobUpdates() {
             if (listeningForJob) {
```
tiramisu24: how come it no longer need to emit request-job-completion?

briehl: The `request-job-completion` command will tell the lookup thread to stop looking up the job status. If there's any other widgets waiting on job information (like the status widget), it'll never see a change. Instead, I just have the log viewer stop listening for updates.

nbextensions/appCell2/widgets/tabs/jobStateList.js
```diff
@@ -22,52 +22,119 @@ define([
...
+                model: model
+            });
+            widgets[jobNumber].start({
+                node: createTableRow(jobNumber),
+                name: isParentJob ? 'Parent Job' : 'Child Job ' + (jobNumber+1),
+                jobId: jobId,
+                initialState: initialState,
+                isParentJob: isParentJob ? true : false,
+                clickFunction: function(jobRow, jobId, isParentJob) {
+                    Array.from(container.getElementsByTagName('tr')).forEach((elem) => {
```
tiramisu24: cant't you just getElementsByClassName()

nbextensions/appCell2/widgets/tabs/jobStateListRow.js
```diff
@@ -71,156 +64,64 @@ define([
...
+                jobId = arg.jobId;                  // id of child job
+                name = arg.name;
+                isParentJob = arg.isParentJob;
+                clickFunction = arg.clickFunction;  // called on click (after some ui junk)
+                updateRowStatus(arg.initialState);
             });
         }
 
         function stop() {
-            return Promise.try(function() {
```
tiramisu24: why is stop empty?

nbextensions/appCell2/widgets/tabs/jobStateListRow.js
```diff
@@ -71,156 +64,64 @@ define([
...
+            clickFunction,
+            isParentJob;
+
+        function updateRowStatus(jobStatus) {
+            jobStatus = jobStatus ? jobStatus : 'Job still pending.';
+            var jobIdDiv = '';
+            if (jobId) {
+                jobIdDiv = div({'style': 'font-size:8pt; color:gray'}, [jobId]);
             }
-            runtime.bus().emit('request-job-update', {
```
tiramisu24: since request-job-update is removed, how does the state get updated now?

nbextensions/appCell2/widgets/tabs/jobStateList.js
briehl: Duh. Yes. Yes I can.
:)

nbextensions/appCell2/widgets/tabs/jobStateListRow.js
briehl: Related to below, it no longer does anything that needs to stop, but the stop function still needs to be present to be a valid widget.

nbextensions/appCell2/widgets/tabs/jobStateListRow.js
briehl: The jobStateList now gets its update that contains all child job information. It pushes the child job info into each jobStateListRow widget. Before, each widget was individually requesting and waiting on results from the kernel for its job, so it was bogging things down. Now, on the usual update loop, the owner widget just updates the state for each child job.

briehl:
Hm, tests are failing. Looking in to that.

kbase-extension/static/kbase/js/util/jobLogViewer.js
tiramisu24: the status widget share the same thread?

nbextensions/appCell2/widgets/tabs/jobStateListRow.js
tiramisu24: ops skimmed over that

nbextensions/appCell2/widgets/tabs/jobStateListRow.js
tiramisu24: clickFunction starts a widget?

briehl: Yes, indirectly. It toggles between what child jobs are being displayed. There's 3 display widgets - Job Params, Job Status, and Job Log. Clicking on a different child job calls `stop()` on each of those, then `start(jobId)` on the new jobId (ok, that's not the real syntax, but you get it).

Clicking on 'View Configure' tells everything under the 'Job Status' tab to stop, which includes all of those widgets as well.

tiramisu24: ohh okay so basically the row doesn't have anything it needs to stop as the child job display widgets stop and restart with the clicked jobId.

briehl: Yup, exactly. It's now just a little thing that formats and displays whatever state gets pushed to it from jobStateList.

PR #1364, opened 2018-07-24
Sct 1210

JamesJeffryes:
Allow a match to a module level type spec file if no more specific match is found allowing for the definition of more generic viewers and download functions
Allow the specification of only a type's parent modules as a "valid_ws_type" to allow any types from that class to be submitted


briehl:
Add tests, please.

The python code tests are in `src/biokbase/narrative/tests`.

briehl:
Build is "failing" because of the r.js issue that's fixed in another PR. This one's fine.

PR #1365, opened 2018-07-25
Remove dead code

JamesJeffryes:
I removed what I thought was dead code from 4 locations:
* kbase-extension/static/kbase/js/kb_sdk_clients
* kbase-extension/static/kbase/images/data-icons
* src/biokbase/narrative/common
* src/biokbase-doc


briehl:
Not sure how you chose these, but you got a little too delete-happy.

kb_sdk_clients/genericClient is still needed, as are those icon pngs (I believe).
In fact, I think everything under kb_sdk_clients is used somewhere.

Can we do this in two passes? I think the python / backend code culling is fine, so let's do that in one PR. The frontend JS code is a bit trickier to sort out all dependencies, so let's do that separately.

briehl:
That's funny, it deletes the file, then adds it back in the diff.
Eh, that's okay. You can leave it as-is.

PR #1366, opened 2018-07-26
Batch results tab

briehl:
This updates the code that runs the results tab. So this PR does several things.

1. Add a 'partial-complete' stage to the 'processing' mode in the App Cell FSM. This gets triggered when the app has child jobs, and at least one is "complete" - either actually completed or in an error state.
2. Refactor the results tab to support batch mode. This entails:
  * Separating the results viewer into its own separate widget that's not entangled with the results tab itself.
  * Adding the jobStateList to the Result tab as is done in the Job Status tab.
  * Added a simple error renderer to the result tab. This gets used to render subjob error. When there's just a single job, we don't even see the Result tab, just an error tab. But with a batch job, some subjobs might be completed fine, while others error, so the result tab needs to be able to show both.

So now, as child jobs finish, the Result tab should become clickable and have ongoing results show up.


briehl:
Added the missing resultsViewer file. This should be ready to test for real now, @eapearson @tiramisu24 .

briehl:
I'll go ahead and put it on CI so we can test there.

PR #1367, opened 2018-07-27
Batch backend fix

briehl:
Child apps were getting kinda orphaned (sometimes) when trying to map their outputs. This was causing ugly errors.

Also, the Job object creation on starting the batch run received an incorrectly formatted set of parameters, causing more render errors. Oops!

Also-also, this builds on #1366 , so that should get merged first.


briehl:
Oops, I probably left that file out of version control. It's new.

briehl:
Hm. actually, this PR builds on #1366, so i'll deal with it in that branch.

PR #1368, opened 2018-07-27
Test updates

briehl:
* Fix issue with App Panel tests trying to invoke a kernel that doesn't exist (and doesn't need to be tested, either)
* Update npm dependencies. Funny how npm depends on some apparently vulnerable things. 
Note that those npm artifacts aren't used in production, only for building and testing. So having a few vulnerabilities is safe enough.


package.json
```diff
@@ -7,29 +7,30 @@
   "devDependencies": {
     "bower": "1.8.4",
     "geckodriver": "1.6.1",
-    "grunt": "1.0.1",
+    "grunt": "^1.0.3",
```
eapearson: I would advise a pinned version number so the build doesn't change when the upstream deps change (or rather have a new version released.) But I know opinions differ on this.

briehl: Good point. Just updated and pinned versions in both package.json and bower.json.

briehl:
Thanks. Once the pinned versions of things passes tests, I'll go ahead and merge.

PR #1369, opened 2018-07-27
fix dynamic_dropdown tests

JamesJeffryes:
Adds the mock and the missing values to the dynamic_dropdown spec. Also fixed the case where the dynamic_dropdown_options are missing


nbextensions/install.sh
```diff
@@ -31,3 +31,6 @@ jupyter nbextension enable advancedViewCell/main --sys-prefix
 
 jupyter nbextension install ${dir}/codeCell --symlink --sys-prefix
 jupyter nbextension enable codeCell/main --sys-prefix
+
+jupyter nbextension enable --py --sys-prefix widgetsnbextension
+jupyter nbextension enable --py --sys-prefix clustergrammer_widget
```
briehl: clustergrammer's not installed at this point. So this'll bomb.

PR #1370, opened 2018-07-28
SCT-1234: Clustergram widget

JamesJeffryes:
**Ready for review**
Compatible data to test this viewer is available here: https://ci.kbase.us/narrative/ws.28852.obj.1 but the viewer should also work with expression or differential expression matrices. 

To invoke in WS code cell:
```
from biopython.narrative import viewers
viewers.view_as_clustergram(<ws_ref>)
```
- [x] Install Dependancies 
- [x] Add JS viewer

### DIFF
--- a/nbextensions/advancedViewCell/widgets/advancedViewCellWidget.js
+++ b/nbextensions/advancedViewCell/widgets/advancedViewCellWidget.js
@@ -433,7 +433,7 @@ define([
             var content = div([
                 p([
                     'Deleting this cell will remove the data visualization, ',
-                    'but will not delete the data object, which will still be avaiable ',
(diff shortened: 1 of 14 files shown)


JamesJeffryes:
I had to add pandas, sklearn and scipy to the requirements for the tests to pass even though they are in the base image. I think that's consistent with the philosophy though

JamesJeffryes:
Duly noted and fixed. Thanks Bill.

briehl:
I think adding pandas, scipy, etc., to the requirements file should be ok. It shouldn't interfere with the base image. I mean, they're already installed in the base image, so "installing" them again should be a quick operation.

PR #1374, opened 2018-08-06
SCT-886 Make a default viewer widget for typed objects

briehl:
If an object doesn't have a registered viewer, right now it throws a big ugly error that might be confusing. This adds a default widget that has some links to the typespec and to the landing page. Not perfect, but better than an error.

In reality, this doesn't happen very often, but it's been a hole in the system for a while that needs a little plugging. The widget looks like this:

screen shot 2018-08-06 at 3 53 00 pm


src/biokbase/narrative/widgetmanager.py
```diff
@@ -571,7 +571,9 @@ def show_data_widget(self, upa, title=None, cell_id=None, tag="release"):
             All objects are related to their viewers by an app. This is the tag for that app's
             release state (should be one of release, beta, or dev)
         """
-        widget_name = 'kbaseNarrativeError'  # default, expecting things to bomb.
+        widget_name = 'kbaseNarrativeError'             # default, expecting things to bomb.
+        default_widget = 'widgets/function_output/kbaseDefaultObjectView'
+        widget_name = default_widget
```
JamesJeffryes: Can't we just condense this to a single line? or at least remove the first line

src/biokbase/narrative/widgetmanager.py
```diff
@@ -620,9 +623,9 @@ def show_data_widget(self, upa, title=None, cell_id=None, tag="release"):
 
                 (input_params, ws_refs) = validate_parameters(app_id, tag,
                                                               spec_params, input_params)
-                (output_widget, output) = map_outputs_from_state([], input_params, app_spec)
-                widget_name = app_spec['widgets']['output']
-                widget_data = output
+                (widget_name, widget_data) = map_outputs_from_state([], input_params, app_spec)
+                # widget_name = app_spec['widgets']['output']
+                # widget_data = output
```
JamesJeffryes: Drop these commented lines?

briehl:
Good ideas, both. In trying to do some formatting, I forgot about the metadata part. Oops. I'll add that.

I can make the warning smaller, but I don't want to remove it or minimize it too much. I think it's useful to tell the user that this isn't necessarily how we want them to view their data - we just don't have a better way yet for that particular object.

briehl:
Ok, while this is testing, an update.
I added separate tabs now, one for the overview (what you see above), and one that just for metadata. It expects key-value pairs, so don't expect any fancy JSON unpacking in the values. I guess we could do that if it's necessary, but I didn't in this first pass. Also removed the big (scary?) warning and made it more subdued.

Also, I wrote some tests that actually mock out the workspace calls, which took far longer than I had hoped, but hey, now it's an example of how to do this in any viewer widget. So, no more excuses for no widget tests!

screen shot 2018-08-08 at 6 05 55 pm
screen shot 2018-08-08 at 6 06 16 pm
screen shot 2018-08-08 at 6 17 56 pm

src/biokbase/narrative/widgetmanager.py
```diff
@@ -614,15 +615,13 @@ def show_data_widget(self, upa, title=None, cell_id=None, tag="release"):
                 obj_param_value = upa if (is_ref_path or is_external) else info_tuple[1]
                 upa_params = list()
                 for param in spec_params:
-                    if any((t == bare_type or t == type_module) for t in param['allowed_types']):
+                    if param.get('allowed_types') is None or any((t == bare_type or t == type_module) for t in param.get('allowed_types', [])):
```
JamesJeffryes: Thanks for adding this in

JamesJeffryes:
I was able to get the local tests to run after updating npm packages. Looks good to me

PR #1379, opened 2018-08-10
SCT-1171 Batch API

briehl:
I just need a checklist, basically, and this seems as good a place as any...

Hope to get this complete-ish by the demo tomorrow. This has functions and documentation to help with building parameter sets for a batch app.

**get_input_scaffold**
- [x] decent docstring
- [x] implement
- [x] test

**list_objects**
- [x] decent docstring
- [x] implement
- [x] test

**list_files**
- [x] docstring
- [x] implement
- [x] test

**generate_input_batch** -- goes in another PR based on this one.
- docstring (revise first!)
- implement
- test

PR #1380, opened 2018-08-15
SCT-1171 Batch input generator

briehl:
* note - this builds on PR #1379 , so that should get reviewed first.

This should be the last piece of the batch input API.
This contains a function `generate_input_batch` that takes in some loosely defined inputs and generates a batch of app run inputs from them.

For example, for running trimmomatic, you could do:

```python
import biokbase.narrative.jobs.batch as batch
ins = {
    'input_reads_ref': '15/45/1',
    'palindrome_clip_threshold': [5, 7],
    'simple_clip_threshold': (0, 2, 10),
    'translate_to_phred33': [0, 1],
    'output_reads_name': 'run_output_pct-${palindrome_clip_threshold}_${run_number}'
}
batch_inputs = batch.generate_input_batch("kb_trimmomatic/run_trimmomatic", **ins)
```

This does a few things.
1. It uses a constant input ref value in input reads ref
2. It uses a list of possible`palindrome_clip_threshold` values
3. It creates a range of values for `simple_clip_threshold`. The syntax of that tuple is (start, interval, end), so this case will make the array [0, 2, 4, 6, 8, 10]
4. It uses a templated string for the output name. This case uses the value of `palindrome_clip_threshold` for that run, and the "magic variable" `run_number` - just the index of the run that gets made.

It doesn't do too much validation or error checking, other than guaranteeing that the values are present. It does do the shuffling of group parameters, though, so you can vary those as well.

Tests are still lacking, so don't merge this yet, but it's mostly functional.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ branches:
     - staging
 
 before_install:
-  - pip install virtualenv coveralls
+  - pip install virtualenv coveralls scikit-learn[alldeps] pandas
   - gem install coveralls-lcov
   - nvm install $TRAVIS_NODE_VERSION
   - npm install -g bower
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,18 @@
 ### OVERVIEW
 The Narrative Interface allows users to craft KBase Narratives using a combination of GUI-based commands, Python and R scripts, and graphical output elements.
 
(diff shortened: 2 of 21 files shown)


briehl:
Updated - tests should work now.

sychan:
Looks good, feel free to merge on the checks complete

PR #1382, opened 2018-08-16
puppeteer [SCT-1191]

thomasoniii:
You probably don't want to merge this pull yet, it's here for informational purposes.

Errata and rough spots are all in the README.txt file.

This adds a few things to test/frontend 

Run it by : cd test/frontend && ./puppeteer.js --config=./config.json

* puppeteer.js actually runs tests. It'll do a full UPA copy-of-a-copy test for a piece of public data, plus any other widget related/browser tests that you specify elsewhere.
* There's a sample config file to run the above, which is almost exactly the same one as the previous casper.js file.
* There are rough spots, mostly around integrating with other tools or behaving nicely as a real automated test harness. But the overall functionality should be there, and with a config file populated with accessible data we should be able to do full copy-of-a-copy testing at a minimum.


thomasoniii:
I added in another file to here that'll let you delete narratives en masse from your account. Say, for example, that you're getting all the wiring in place to make puppeteer work and end up half-running your script _a lot_ and it bombs out and doesn't properly clean things up. That's a good candidate for this. It's in /scripts.

i.e.,
./delete_narratives.js --token=MYTOKEN --criteria='n.ws[0] > 35000'

PR #1387, opened 2018-08-29
SCT-1219 Fix multiple output handling

briehl:
Multiple output objects should now throw an error when trying to give the same name to multiple output objects in the same app run.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/appWidgets2/input/newObjectInput.js
+++ b/kbase-extension/static/kbase/js/widgets/appWidgets2/input/newObjectInput.js
@@ -28,13 +28,15 @@ define([
                 value: undefined
             },
             dom,
-            runtime = Runtime.make();
+            runtime = Runtime.make(),
+            otherOutputParams = config.closeParameters.filter((value) => {
+                return value !== spec.id;
+            });
 
         // Validate configuration.
         // Nothing to do...
 
         options.enabled = true;
-
 
         /*
          * If the parameter is optional, and is empty, return null.
@@ -51,12 +53,12 @@ define([
 
         function setModelValue(value) {
             return Promise.try(function() {
-                    if (model.value !== value) {
-                        model.value = value;
-                        return true;
-                    }
-                    return false;
-                })
+                if (model.value !== value) {
+                    model.value = value;
+                    return true;
+                }
+                return false;
+            })
                 .then(function(changed) {
                     render();
                 });
@@ -64,8 +66,8 @@ define([
 
         function unsetModelValue() {
             return Promise.try(function() {
-                    model.value = undefined;
-                })
+                model.value = undefined;
+            })
                 .then(function(changed) {
                     render();
                 });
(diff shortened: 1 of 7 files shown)


JamesJeffryes:
Thanks for taking this on! My understanding of the buses is pretty fuzzy

briehl:
Added another edge case fix. When multiple outputs are successfully set, then one is changed to something invalid, it was showing the error, but still allowing the 'Run' button to be activated.

Here's the total set of fixes added:

* Update the newObjectInput widget to have awareness of the values of all output objects on validation. On value change it will also send along whether it was changed to an invalid state or not.
* Update the newObjectInput widget to know how to validate unique values, compared to a list of "closeParameters" that's passed to it on initialization.
* Update the appParamsWidget to send the list of ids of "closeParameters" (which can be context dependent - the only context implemented so far is sending all output parameters to each widget labeled as "output" by the spec.
* Update the appParamsWidget to respond to any field widget requests for a list of parameter values (needed to do cross-validation)
* Update the appCellWidget, during validation, to take an optional error field in the 'parameter-changed' bus message. If this is set to true, then it will always treat the parameters as invalid. Otherwise, it'll go through and validate each parameter individually again, which doesn't do any cross-comparison.

Just want to add some notes here, from this little dive into the backwaters of the app cell validation code.
1. We gotta write down what bus messages get sent where and what the format is, etc. I spent most of my time just deciphering that.
2. Individual input field widgets validate themselves, then, once they're deemed valid, the appCellWidget validates them all again, using code that seems to be copied/pasted into different places. This is a little redundant, and hard to maintain, since the same validation code is duplicated.
3. There's a few places where direct workspace calls are made to validate inputs/outputs. To be fair, those are done through a little wrapper - this wrapper should be redirected to a more central caching API that talks to the NarrativeService (or, ideally, browser-cached info from the Data Panel) instead.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/newObjectInput.js
```diff
@@ -80,49 +86,65 @@ define([
...
+                        diagnosis: 'disabled'
+                    };
                 });
+            }
+            else {
+                var rawValue = getInputValue();
+                return validateUniqueOutput(rawValue)
+                    .then((isUnique) => {
+                        if (!isUnique) {
+                            return Promise.try(() => {
```
eapearson: this doesn't need to return a promise, just the value, since it is already wrapped in then, which itself always returns a promise.

eapearson:
Without having run it locally yet, but reading the notes and scanning the code, the changes make sense. I especially like the addition of the carrying error state in the parameter value flow as well as the closeParameters to allow a parameter to react to multiple other parameters (if I'm understanding that right.) Of course, the devil is in the details.

Having used knockout extensively over the last year or so, I can still really see the advantages of this approach. Although the baby bus in the narrative is not really the same thing, it is being used that way. 

It would be instructive some day to try a real reactive framework like knockout, rxjs, etc. to wire this stuff together. They are very black-boxy, but because of that one can do a more sophisticated composition of things (with the various hacks required to work around traps hidden inside the black boxes...)...

briehl:
Agreed. It would be nice to take some recent lessons learned and come back to this with some kind of higher-level framework.

It would also be a nice project to someday make this more general for the Jupyter (and maybe others) community. I've chatted with some folks who wanted to make use of something like this, but we're so deeply intertwined with KBase that this is impossible right now.

PR #1388, opened 2018-08-29
Clustergrammer build

JamesJeffryes:
So unfortunately I was editing the wrong file earlier. This version should actually enable the widget in the built image

### DIFF
--- a/scripts/install_narrative_docker.sh
+++ b/scripts/install_narrative_docker.sh
@@ -26,6 +26,10 @@ console "Installing biokbase modules"
 cd $NARRATIVE_ROOT_DIR/src
 console "Running local 'setup.py'"
 ${PYTHON} setup.py install
+console "installing sklearn & clustergrammer_widget'"
+# We install clustergrammer_widget and sklearn specially here so that it does not
+# clobber dependencies in the base conda image
+pip install --no-dependencies sklearn clustergrammer_widget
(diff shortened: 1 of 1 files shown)


scripts/install_narrative_docker.sh
```diff
@@ -110,5 +114,6 @@ jupyter nbextension install $(pwd)/codeCell --symlink --sys-prefix
 jupyter nbextension enable codeCell/main --sys-prefix
 
 jupyter nbextension enable --py --sys-prefix widgetsnbextension
+jupyter nbextension enable --py --sys-prefix clustergrammer_widget
```
JamesJeffryes: It might make sense to replace this logic with a call to `nbextention/install.sh`

JamesJeffryes:
At this point I wonder if it might almost be easier to update the narrative base image but I don't know how to do that.

briehl:
I was thinking that might be the way to go, too. I don't think that's written down anywhere, either, since the most recent update.

I'll dredge my memory for how to do it, document it, and make the update. I think having this setup for running it locally is fine, too.

JamesJeffryes:
@briehl Should we merge this or wait till the base image is updated? I'd kinda like to get this working for demos Fri

briehl:
Let's merge now and see if it works. I can work on the base image in the meantime.

PR #1394, opened 2018-09-04
GenomeClassifier Viewers

JamesJeffryes:
Adds simple viewers for GenomeCategorizer and GenomeClassifierTrainingSet

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/GenomeCategorizer.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/GenomeCategorizer.js
@@ -28,7 +28,7 @@ define ([
     'use strict';
 
     return KBWidget({
-        name: 'kbaseGenericMatrix',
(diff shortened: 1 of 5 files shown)


JamesJeffryes:
@briehl I haven't added in tests yet. I'm going to try to replicate your test for the default viewer but may have to pester you a bit

briehl:
Ok. Go ahead and pester if you need to. :)

JamesJeffryes:
@briehl  OK, so I kind of cheated on the mocks. In one case, i'm pulling two different WS objects so I was not sure how to make that happen with `jasmine.Ajax.stubRequest`. My searching was not fruitful so I got lazy and inject the expected data to make sure the rendering code is functional. That's the section most prone to change so I think i've covered the most critical lines with these tests.

briehl:
Looks fine to me.

PR #1395, opened 2018-09-07
Remove IGV viewer

JamesJeffryes:
This was always a proof of concept and getting the right environment variable set in prod so it goes away seems to be non-trivial. I say we just yank it.


briehl:
You're right. And from Jim's notes, it seems that IGV might not be the best option in the long run. I'll extract those notes then merge this.

IGV is also a dependency that was manually installed in ext-packages (https://github.com/kbase/narrative/tree/ddf4f16f346e72ca02680bd1d25c30865509effc/kbase-extension/static/ext_packages/igv) so we should remove that, too, if you don't mind.

PR #1396, opened 2018-09-07
bump version in the config template

briehl:


briehl:
looks like a lot, but it's all whitespace.

PR #1398, opened 2018-09-10
Sct 1315

JamesJeffryes:
0 should be represented as 0 not null. Also renamed deprecated keys for jquery-datatables


briehl:
Looks good.

PR #1400, opened 2018-09-13
Sct-1253: Refactor Alignment Viewer

JamesJeffryes:
This drops the old kbaseRNAseqPie viewers (which did not display pie charts...) and replaces them with a simpler new viewer. Also I added a unit test for the new viewer.


briehl:
This might cause problems - those older modules have their names embedded in the Narrative output viewer, so deleting the files would cause those cells to error out. You can probably deal with this by redirecting the names through the narrative_paths file to your new widget. But you'll have to support what inputs they take, too.

Another option might be to tweak the updater.py module (src/biokbase/narrative/updater) to scan and look for things like that before the Narrative even gets opened, but that's another hairball.

JamesJeffryes:
Oh, I can do the first option because we just changed the output of NarrativeViewer app. I didn't realize that the NarrativeViewer app wouldn't get called when the notebook was reopened though

PR #1401, opened 2018-09-13
Changing the text in the drop zone to recommend Globus for large files. SCT-1370

landml:
@briehl  - Users need a clearer warning that Drag-n-drop stops working around 20GB and they need to switch to Globus. Roy suggested that we need to specify when and why Globus is offered as an option. The documentation is being updated for SCT-1370.

This PR is a revision to the text users will see in the drop zone of the Import panel.

### DIFF
--- a/kbase-extension/static/kbase/templates/data_staging/dropzone_area.html
+++ b/kbase-extension/static/kbase/templates/data_staging/dropzone_area.html
@@ -11,7 +11,16 @@
         </div>
     </div>
     <div class="dz-message">
-        <p>Drag and drop data files (or click) in this box to upload them to your staging area.<br><br>
+        <p>Three ways to add data to the staging area:</p>
+        <ul>
+          <li>Click in this box
+          <li>Drag and drop data files 
+          <li>For large files (over 20GB) or a large number of files use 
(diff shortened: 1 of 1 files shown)


briehl:
This is a good idea, but there's a couple issues.

Here's what the text looks like:
screen shot 2018-09-13 at 2 51 42 pm

I think it might be better with the bulleted text left-justified, and the bullets scooted over to the right to compensate.

And here's what happens at that globus link: 
screen shot 2018-09-13 at 2 52 00 pm

I think the `"&destination_path="` part needs to end with the user's name, like `"&destination_path=wjriehl"`. That'll take an adjustment to the template. It looks like the username is passed through, but unused, so you can set `"destination_path={{username}}"`.

scanon:
@briehl Could we somehow add a check when the user click on Globus to confirm if that have a Globus account linked to their KBase account?

briehl:
@landml @scanon That link below the box automatically puts the username in there. If that's not working, then they probably don't have a linked globus account. I'll investigate how to check that. I bet there's something in Auth that'll help. Maybe we change that text around to prompt them based on that info?

briehl:
I made the changes we talked about in a separate branch. Closing this PR to open a new one.

PR #1406, opened 2018-09-20
SCT-683 Add script to fix workspace info for search

briehl:
This is mostly complete (but no tests yet), so I wanted to get some feedback.

This contains a little script that updates the metadata for a single workspace to clean out some issues. Most of these issues get automatically sorted out whenever a user saves their Narrative, but lingering problems have been sticking around from old Narratives from folks who aren't going back. 

To that end, this does the following:
1. Make sure the "narrative" metadata field contains an int that points to the narrative.
2. Make sure the "narrative_nice_name" metadata field is correct.
3. Make sure the "is_temporary" metadata field exists and is correct.
4. Adds a count of the number of narrative cells.
5. Does nothing at all if there's > 1 narrative in the workspace.

That last one is tricky - users could have fixed that themselves with the tooling built into the narrative. I suspect there aren't many of those, though, and the logic gets fuzzy, especially if the object id in the workspace metadata is wrong. I propose leaving it out, unless there's an objection.

Still todo:
- [x] Tests. This ain't getting run without a pile of tests.
- [x] Write a wrapping function that iterates over all workspaces, or build this out into a command line tool.
- [x] Add code to update the number of cells in Narrative and NarrativeService (on new narrative creation). This gets triggered:
  * on new narrative
  * on copy
  * on save


briehl:
Also, maybe this belongs in a different repo. I'm just putting it up here for a point of discussion.

src/scripts/fix_workspace_info.py
```diff
@@ -0,0 +1,85 @@
+"""
+Fixes workspace info to do the following.
+1. Make sure the "narrative" metadata field contains an int that points to the narrative.
+2. Make sure the "narrative_nice_name" metadata field is correct.
+3. Make sure the "is_temporary" metadata field exists and is correct.
+4. Adds a count of the number of narrative cells.
```
MrCreosote: Presumably this will also need to be updated on narrative save, right?

src/scripts/fix_workspace_info.py
```diff
@@ -0,0 +1,85 @@
...
+5. Does nothing at all if there's > 1 narrative in the workspace.
+Note that while this fetches the Narrative object, it doesn't modify it in any way.
+"""
+from biokbase.workspace.client import Workspace
+
+ws_url = "https://ci.kbase.us/services/workspace"
+
+def fix_workspace_info(ws_id, token, verbose=False):
```
MrCreosote: It'd be more useful if this took an admin token and used admin methods to fix things up (I don't think there's an alter_workspace_metadata admin fn yet, so the script would have to set and unset permissions)

src/scripts/fix_workspace_info.py
```diff
@@ -0,0 +1,85 @@
...
+    assert(str(ws_id).isdigit())
+
+    new_meta = dict()
+    ws = Workspace(url=ws_url, token=token)
+
+    if verbose:
+        print("Checking workspace {}".format(ws_id))
+
+    # test if there's exactly 1 Narrative object in the workspace
+    narr_obj_list = ws.list_objects({'ids': [ws_id]})
```
MrCreosote: This will list all objects in the ws, not just narratives. Filter by type would be helpful here.

src/scripts/fix_workspace_info.py
MrCreosote: Note there's no error handling for if the ws throws an error.

src/scripts/fix_workspace_info.py
briehl: Yup, we'll need another PR to the narrative repo.
Just wanted to throw this out there before adding extra work. But I'll make this a checkbox in the todos.

src/scripts/fix_workspace_info.py
briehl: That should work in to the calling function. If this is being called by some kind of shell script, then that should catch it when this crashes (and I'll change this to take in the url as a parameter or something). If we want this here to be an all-in-one script, then whatever manages the loop should deal with it.

briehl: Sure, we can do that. I mostly left that aside, since I'm not sure the best way for it to run. I kinda assumed that there's some magic way to get a superuser token. Would it be hard to write a fn like that?

MrCreosote: You just have to get a token for a user that's a workspace admin (or we make you one), and then use the administrative interface: https://ci.kbase.us/services/ws/docs/administrationinterface.html

briehl:
Update: unittests done and committed with dummy test data.

Next (final) step - write integration tests against mini_kb workspace. Shouldn't be too much work.
Steps here:
1. start up mini kb (assume no data)
2. add user(s) to auth
2a. use pre-configured wsadmin user with token as ws admin
3. add Narrative data type to workspace
4. add similar 7-10 workspaces based on unittest data to cover states of workspace metadata
5. add dummy narratives to workspace (create some dummy json data files to upload)
6. Tinker with ws meta to get it in the right/wrong/broken state as dictated by unittest data
7. Run fixit script.
8. Validate all workspaces.

Sounds like a lot, but most of the work will be coming up with the test cases, inserting them into the workspace (and tracking ws ids, etc, for things that should be skipped), then validating results. E.g., for the case where there's > 1 narrative in a workspace, and semi-configured metadata, it should be skipped.

briehl:
The script itself is as done as it's going to be.
You'll need to have the workspace client on your Python path, then run the script with:
`python fix_workspace_info.py -w <workspace_url> -a <auth_url>, -t <ws_admin_token> -m <max workspace id> -o <output file>`

or just `python fix_workspace_info.py -h` to see the options.

`workspace_url` = the service endpoint: http://ci.kbase.us/services/ws
`auth_url` = the auth endpoint: http://ci.kbase.us/services/auth
`ws_admin_token` = an active auth token for a workspace admin on that host
`max_workspace_id` = an int for the highest number of workspace that should be checked. Any that don't exist (or are deleted or locked) will be ignored
`output_file` = the name of an output file for the results. These are in JSON, and generally give the workspace id and the last time it was modified. There's 4 high-level keys in the results:
1. `fail` - those workspaces that threw an error somehow when looking them up. They're either deleted, don't exist, or are locked. Whatever the reason, the error message is also given.
2. `multiple` - these workspaces have multiple narratives. Their moddate is also given. If they're used relatively recently, it might be worth fixing them manually. The logic to do it automatically gets hairy.
3. `skip` - these workspaces are skipped because they have no narratives.
4. `update` - these workspaces all had their metadata updated.

briehl:
So... anyone need this change anymore?

eapearson:
This includes setting the searchtags to "narrative" for both the fixup script and narrative save? Yes! I saw a/the ticket move my way yesterday, I'll try to attend to it. 
This sprint is the best opportunity to get the corresponding changes into search ui (i.e. using the new searchtag).

PR #1408, opened 2018-09-22
SCT-1370 Show how to link an account to Globus

briehl:
* Change the dropzone uploader text to tell users to use Globus.
* Change the text all together to go to docs on how to link an account to Globus if it's not already.

Still needs tests, so don't merge yet.


briehl:
assuming the tests pass, this should be good enough.

PR #1410, opened 2018-09-25
SCT-1395: Convert object ref and IDs to common names for download

JamesJeffryes:
I suspect some UPA/ref path work resulted in objectId becoming a numerical ID rather than an object name. I added in a WS call to fetch the common name and update the objectId and ref accordingly.

I also added what tests I could (but was not sure how to simulate this particular change)


briehl:
That's certainly more tests than that module had before. :)
I think this looks fine. If tests pass, let's merge and get it up on CI.

PR #1421, opened 2018-10-15
Sct-1526

JamesJeffryes:
This new viewer widget accepts a few params (matrix_ref and output_type for now), generates some python code, launches and executes a Code Cell with that code and then removes itself (which ensures that a duplicate code cell is not inserted when the narrative is reloaded).

It's a working demo but have a few open issues:
Should the python code go into `pythonInterop.js` or should we keep that to just narrative functions?
I haven't decided how to test the widget correctly yet (I have an idea though).
Product Team will have some feedback on which visualizations should be supported and what parameters should be able to be set.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
+++ b/kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
@@ -1,0 +1,136 @@
+/**
+ * Output widget to generate a code cell with vis output.
+ * @public
+ */
+
+define ([
+    'uuid',
+    'jquery',
+    'kbwidget',
+    'kbaseAuthenticatedWidget',
+    'kbaseTabs',
+    'narrativeConfig',
+    'base/js/namespace',
+    'common/utils',
+    // For effect
+    'bootstrap',
+    'jquery-dataTables',
+    'datatables.net-buttons',
+    'datatables.net-buttons-bs',
+    'datatables.net-buttons-html5',
+    'datatables.net-buttons-print'
+], function(
+    Uuid,
+    $,
+    KBWidget,
+    kbaseAuthenticatedWidget,
+    kbaseTabs,
+    Config,
+    Jupyter,
+    utils
+) {
+    'use strict';
+
+    return KBWidget({
+        name: 'codeCellLauncher',
+        parent : kbaseAuthenticatedWidget,
+        version: '1.0.2',
+        options: {
+            output: null,
+            workspaceID: null,
+            loadingImage: Config.get('loading_gif')
+        },
+
+        init: function(options) {
+            this._super(options);
+            // Create a message pane
+            this.$messagePane = $('<div/>').addClass('kbwidget-message-pane kbwidget-hide-message');
+            this.$elem.append(this.$messagePane);
+            return this;
+        },
+
+        launchCodeCell: function (code) {
+            const thisCell = this.$elem[0].parentElement.parentElement.kb_obj.cell,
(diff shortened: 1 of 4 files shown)


briehl:
Oh, I see what you're doing now. Yeah, this is fine for a demo. Another solution would be for the App Cell to know how to make these, based on info from the app spec. It would then just generate one directly at the end of an app run instead of a viewer widget that coughs up the code cell and deletes itself. But I think this is worth demoing the concept, then refactoring.

kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
```diff
@@ -0,0 +1,121 @@
...
+                    '    "layout": go.Layout(title="Stacked bars" xaxis={"title": "Columns"}, yaxis={"title": "Row Values"}, barmode="stack")\n' +
+                    '})'
+
+            } else {
+                this.showMessage(`Unsupported output type ${type}`);
+            }
+            return code;
+        },
+
+        loggedInCallback: function(event, auth) {
```
briehl: You aren't really doing anything that needs authentication here. Might as well just put this logic in the init function.

kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
```diff
@@ -0,0 +1,121 @@
...
+            console.log(this.options);
+            const type = this.options.output_type,
+                matrix_ref = this.options.matrix_ref;
+
+            const code = this.get_python_code(type, matrix_ref);
+            if (code) this.launchCodeCell(code);
+            return this;
+        },
+
+        loggedOutCallback: function() {
```
briehl: Also unnecessary.

kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * Output widget to generate a code cell with vis output.
+ * @public
+ */
+
+define ([
+    'uuid',
+    'jquery',
+    'kbwidget',
+    'kbaseAuthenticatedWidget',
```
briehl: You could leave this out, too (see other comments)

kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
```diff
@@ -0,0 +1,121 @@
...
+        },
+
+        showMessage: function(message) {
+            let span = $('<span/>').append(message);
+
+            this.$messagePane.append(span);
+            this.$messagePane.show();
+        },
+
+        hideMessage: function() {
```
briehl: Nothing ever calls this. Is it needed?

kbase-extension/static/kbase/js/widgets/function_output/codeCellLauncher.js
```diff
@@ -0,0 +1,121 @@
...
+            if (code) this.launchCodeCell(code);
+            return this;
+        },
+
+        loggedOutCallback: function() {
+            this.isLoggedIn = false;
+            return this;
+        },
+
+        showMessage: function(message) {
```
briehl: You could probably just embed this in the one place that calls it. I would just have it do something like:

`this.$messagePane.append('Some message')` 

and that would work just as well.

JamesJeffryes:
> Another solution would be for the App Cell to know how to make these, based on info from the app spec.
I prototyped this but decided that this would actually be preferable because it cuts out the delay of launching and executing the app.

PR #1422, opened 2018-10-15
Add 'staging' button to some types for download

briehl:
The allowed types are hard-coded for now. We can change this later. Just putting something together for the demo.

@Tianhao-Gu

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
@@ -915,6 +915,7 @@ define([
                         type: type,
                         objId: objId,
                         ref: objRef,
+                        objName: objData.objectInfo.name,
                         downloadSpecCache: _this.downloadSpecCache
                     });
                 });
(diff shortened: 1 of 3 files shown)


kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataList.js
```diff
@@ -915,6 +915,7 @@ define([
                         type: type,
                         objId: objId,
                         ref: objRef,
+                        objName: objData.objectInfo.name,
```
JamesJeffryes: I knew there had to be a better way to do this when I wrote my patch

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDownloadPanel.js
```diff
@@ -116,27 +100,46 @@ define (
...
+    		                self.runDownloader(self.type, self.ref, self.objName, descr);
     		            }));
     		};
-    		
-    		var downloaders = self.prepareDownloaders(self.type, self.ref, self.objId);
-    		for (var downloadPos in downloaders)
-    			addDownloader(downloaders[downloadPos]);
-		
+
+            let addStagingButton = () => {
```
JamesJeffryes: Minor but do you want to go ahead and use ES6 declarations throughout?

briehl:
I thought about updating that module to the newer syntax and so on, but I think the whole thing could be rewritten for some clarity and such, and don't have the time right now to devote to it.

PR #1430, opened 2018-11-02
SCT-1509 Added link to add globus ACL before going to link

bio-boris:
@briehl

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
@@ -162,22 +162,16 @@ define([
             // Add ACL before going to the staging area
             var globus_flag = false;
             this.$elem.find('.globus_link').click(function (e) {
-
-                if(globus_flag){
-                    globus_flag = false;
-                    var url =  window.location.hostname + "/services/staging_service/add-acl";
-                    $.ajax({
-                        async: false,
-                        url: url,
-                        xhrFields: {
-                            withCredentials: true
-                        }
-                    }).done(function (data) { console.log(data); });
-                    return;
+                if (!globus_flag) {
+                    e.preventDefault();
+                    this.stagingServiceClient.addAcl()
+                        .done(
+                            data => {
+                                globus_flag = true;
+                                $(this).trigger(e.type); //e.click Throws an error for 2nd call of double call
+                            }
+                        )
                 }
-                e.preventDefault();
-                globus_flag = true;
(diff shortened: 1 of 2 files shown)


kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -153,11 +153,33 @@ define([
...
 
+            // Add ACL before going to the staging area
+            var globus_flag = false;
+            this.$elem.find('.globus_link').click(function (e) {
+
+                if(globus_flag){
+                    globus_flag = false;
+                    var url =  window.location.hostname + "/services/staging_service/add-acl";
+                    $.ajax({
+                        async: false,
```
bio-boris: Window.location.hostname might be bad and theres a better way to get it?
async:false might be bad?

briehl: Yeah. There's already a staging service client that'll probably do what you need it to. I forgot about that. :P It's here: https://github.com/kbase/narrative/blob/develop/kbase-extension/static/kbase/js/api/StagingServiceClient.js You'll have to add the wrapper for the add-acl endpoint. Add a line something like:
```
addAcl: {method: 'get', path: 'add-acl'}
```
to that client.

Then run it similar to other parts of this file:
```
    this.stagingServiceClient.addAcl()
    .then(function(data) { - other logic - });
```

briehl: And I think you're still going to run into a race condition against the call to the service.
You're right, setting `async: false` isn't a great idea. If the network's down, it'll lag the browser.

It looks like the logic here is to test if the globus_flag is set. If not, then set it and prevent the link from happening, and trigger another click. If it is set, then unset it, and go to that endpoint.

A way to do this have have it maintain asynchrony would be to reverse it.
Something like:
```
if (!globus_flag) {
    e.preventDefault();
    this.stagingServiceClient.addAcl()
        .then(data => {
            globus_flag=true;
            $(this).trigger(e.type);
        );
}
else {
    globus_flag = false;  // not doing anything, and not preventing the default event will just let it happen.
}
```
I'm actually unsure if that last `globus_flag=false` is really necessary. Resetting it will trigger that ACL thing every time, which isn't really necessary past the first.

Also, what happens if that call fails? Should there be an alert? Should it prevent sending the user to Globus? If you want to just ignore any errors, then change `.then` to `.done` - "done" is activated whenever the call finishes, regardless of result state.

PR #1431, opened 2018-11-06
Handle double-clicking on globus link

briehl:


bio-boris:
@briehl It's very fast, and works well. Do you think we should do the same for the other globus link right above it? I just noticed there's a second one "For large files over 20GB..."

briehl:
@bio-boris 
Ok, made the change to the link in the actual dropzone.

I also updated the Globus link - apparently the endpoint we're using is deprecated and will be removed in April. We don't **need** to change it right now, but I don't see a reason not to.

PR #1437, opened 2018-12-05
SCT-1602 Allow workspace id for Narrative URLs

briehl:
Currently Narrative URLs all look like this: https://narrative.kbase.us/narrative/ws.###.obj.###

This will adjust to allow urls like this: https://narrative.kbase.us/narrative/xyz where xyz = a numerical workspace id. The changes in this PR handle the following cases:
1. workspace has a single narrative, set in the workspace metadata.
2. workspace has a single narrative, but NOT set in the ws metadata.
3. If the workspace has either 0 or > 1 narrative (and the ws metadata is not set), it will throw a 404 error.

This wound up requiring a refactor of some deep chunks of the ContentsManager that's plugged in to Jupyter, so there's a bunch of changes and more tests are needed before merging (but I want them tested in Travis-CI as well).

Also, there's some front end adapter work that needs to be done, still.

### DIFF
--- a/kbase-extension/static/kbase/js/common/jupyter.js
+++ b/kbase-extension/static/kbase/js/common/jupyter.js
@@ -140,7 +140,6 @@ define([
         getCells: getCells,
         getCell: getCell,
         disableKeyListenersForCell: disableKeyListenersForCell,
-        // getWorkspaceRef: getWorkspaceRef,
         onEvent: onEvent,
         uiModeIs: uiModeIs,
         canEdit: canEdit,
--- a/kbase-extension/static/kbase/js/common/runtime.js
+++ b/kbase-extension/static/kbase/js/common/runtime.js
@@ -100,10 +100,7 @@ define([
          * The kbaseNarrative object does this, but it also does a lot more...
          */
         function workspaceId() {
-            var wsInfo = window.location.href.match(/ws\.(\d+)\.obj\.(\d+)/);
-            if (wsInfo && wsInfo.length === 3) {
-                return wsInfo[1];
-            }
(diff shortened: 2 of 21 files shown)


eapearson:
Wow, there are cases where "workspace has a single narrative, but NOT set in the ws metadata."? Given that most other calls to fetch a narrative from the WS will fail without proper metadata, what do you think about making this a 404 as well? I'm sure you've thought that through, just curious about the reasoning.

What do you think about a Narrative Service method which implements the same logic? I think a method to provide basic narrative info, even if just the ws and obj info, which takes just the workspace id, would be quite useful.

MrCreosote:
I would just find them and fix them rather than including special logic just in case...

briehl:
Did that. Well, I wrote a script that'll do that. It should be run across CI first, and maybe reviewed again.
https://github.com/kbase/narrative/pull/1406

briehl:
Still, anyone with some knowledge of the system can easily mess up their workspace metadata (by incorrectly copying a narrative, or copying just the narrative object to another workspace, etc.). I think it's a useful check to keep, even if it won't get used often.

MrCreosote:
I would just throw an error if there's no narrative ID in the metadata or it doesn't match a narrative. Teach them dorkwits to Not Do That.

briehl:
Front end changes now work and support the new url structure. That was annoying.

I'll bow to the majority of @eapearson and @MrCreosote and just throw an error if the workspace metadata isn't set up right. Changes still upcoming.

briehl:
ok, ready for review again.

PR #1438, opened 2018-12-06
Follow WS refs in subdataInput

JamesJeffryes:
- Added the ability to prefix a portion of path_to_subdata with "WSREF" in order to load the object referenced there and list options in the referenced object.
- Added a minimal test (cuz I was sick of JS at this point)
- Removed a duplicated copy of subdataMethods and fixed a reference in customSubdataInput.js


kbase-extension/static/kbase/js/widgets/appWidgets2/input/subdataMethods/manager.js
```diff
@@ -71,158 +87,150 @@ define([
...
+                    return item;
+                });
+            };
+            if (subdataSelection.service_function) {
+                dataCall = genericClientCall(subdataSelection, subObjectIdentity);
+            } else {
+                dataCall = workspaceCall(subObjectIdentity);
+            }
+            let subdata_path = subdataSelection.path_to_subdata[0];
+            if (subdata_path.startsWith("WSREF(")) {
```
eapearson: not a big deal, but better to set WSREF as a constant, so that the slice below is not vulnerable to changes in it. Not likely in this deep code, but e.g. someone might want to change it to UPA or whatever and not know that they need to change the 6 below to a 4.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/subdataInput.js
```diff
@@ -10,7 +10,7 @@ define([
     'common/ui',
     'common/props',
     'base/js/namespace',
-    '../subdataMethods/manager',
```
eapearson: rather than using one subdataMethods (I saw the comment about consolidating to one version, which is good) based in the input parameters directory place it one level above so that it is clear it is shared from that root; a dependency on a sister directory is a big confusing (things are in input or view directories specifically because they are or might exhibit behavior specific to being input parameters or view-only parameters.)

eapearson:
Hi James, 
It passes tests, so I haven't bothered to test it locally, but I did review the code, such as I can, and had only a couple of comments which are more stylistic. 
We did talk about removing the custom subdata, which I see is included here. I assume that you've looked for those parameters in extant apps and found them absent?

JamesJeffryes:
I think didn't actually remove `custom subdata` (at least not intentionally) rather just one of the duplicated `subdataMethods` directories. I don't know of anything using customSubdata, growthCondition or the like but I left them in for the time being

eapearson:
ah, ok.

PR #1445, opened 2019-02-03
improved google analytics support

eapearson:
- add per-deploy-env google analytics tracking id; previously was hardcoded to production
- update google analytics usage to use latest recommended google js lib
- improvements to local-dev-with-docker docs and script

This PR strives to improve google analytics usage, as part of the current effort to improve GA support at KBase. To wit:

1. Update the analytics library to the recommended google client. That is, since the Narrative already uses the GA client downloaded from google via the code snipped published by google, this replaces the current google client code snippet with the current release.  The previous client was "analytics.js", and the current one is "gtag/js". Usage is a bit different. This matches the new google analytics usage in the static app catalog.

2. Utilize unique a GA tracker id per KBase deployment environment. Previously the GA id was hardcoded into the page.html template. This caused all Narrative usage to send page hit stats to the production GA account. We have individual accounts for each KBase deployment environment, each with their own individual GA tracker id. 
This prevents development and testing usage from interfering with production stats, and enables real-time and historical monitoring of all kbase deployment environments. It also allows us to more easily test the usage of GA, as we can conduct tests in CI etc. without interference form other environments.
The PR also includes a config stanza for narrative-dev, since I needed to test this locally against each deployment environment.

I have assumed that the current deployment strategy for Narrative still uses config.json.templ. If this is not the case, please indicate the current deployment config in a comment and I'll adjust the PR accordingly.

Also included in this PR are some changes to the local developer tool and doc for dockerized local deployment.

### DIFF
--- a/docs/developer/local-docker.md
+++ b/docs/developer/local-docker.md
@@ -2,8 +2,6 @@
 
 It is useful to run the Narrative within a local docker container. E.g. this makes it easy to work on Narrative ui locally integrated with a local instance of kbase-ui.
 
-
-
 ## Narrative
 
 The following changes are required:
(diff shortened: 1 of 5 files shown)


briehl:
Looks right.

PR #1447, opened 2019-02-08
Merge to master for 3.10.0 release

briehl:
Note that this includes a feature that lets users send a notification to others, requesting access to their narrative. This means that both feeds and NarrativeService (along with ws admin and service tokens) need to be up to date on prod before deploying this.


briehl:
Pinging @sychan and @scanon so they have eyes on this. Especially the need to update NarrativeService with tokens in prod before deploying. Ping!

PR #1449, opened 2019-02-14
SCT-1783 Some refactor of the manage panel

briehl:
It's not perfect, but only makes a few calls now, instead of one for each user.

I also refactored and deleted a bunch of code.
Honestly, this widget is still a mess of mashed together logic and view and whatever, but it's much more contained.


briehl:
@eapearson I think you were the last to work on that panel - mind taking a look?

PR #1451, opened 2019-02-16
Delete unused modules

briehl:
There might be more, but this is a good first pass.


briehl:
Heard no comments or anything despite multiple requests, so I'm assuming this is fine. Merging.

PR #1452, opened 2019-02-20
Share with org

FlyingUnicornSF:
added request to add  narrative to org.


package.json
```diff
@@ -5,8 +5,8 @@
   "private": true,
   "repository": "github.com/kbase/narrative",
   "devDependencies": {
-    "bower": "1.8.4",
-    "geckodriver": "1.6.1",
+    "bower": "^1.8.4",
```
eapearson: IMO It is dangerous to use non-deterministic dependencies specs like ^SEMVER in a top level app like this -- it means that the build will change over time even though no-one made changes to the codebase. (Libraries must use them, because of the way JS dependency managers work.) Even though this usage of ^SEMVER _should_ ensure that the build doesn't break (because it will only allow changes in the second and third server positions, which are always supposed to be backwards compatible), in the JS world "safe" dependency changes can sometimes break. 

I do think it is good to update the deps from time to time not just to to pick up improvements but more importantly to reduce the amount of upstream code change when one does update deps. In the best of worlds, one would inspect the diff in the upstream dependency to see what exactly has changed to evaluate whether it is (a) worth the dependency update and (b) to understand what is different and whether it might affect something. In reality, that is quite difficult to do!

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSharePanel.js
```diff
@@ -72,6 +73,40 @@ define ([
...
+                json: true,
+                headers:{
+                    "Authorization": token,
+                    "Content-Type": "application/json",
+                },
+            })
+            .then(response => response.json())
+            .then(response => {
+                this.updateOrgList(response);
+                console.log("Success:", JSON.stringify(response))
```
eapearson: imo console.log lines should be removed before release.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSharePanel.js
```diff
@@ -325,6 +471,42 @@ define ([
...
+                json: true,
+                headers:{
+                    "Authorization": token,
+                    "Content-Type": "application/json",
+                }
+            })
+            .then(response => response.json())
+            .then(response => {
+                if(response.error) {
+                    console.log(response.error.message)
```
eapearson: ditto here re console.log

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSharePanel.js
```diff
@@ -181,6 +220,104 @@ define ([
...
+                        $applyOrgBtn.addClass('disabled');
+                    }
+                });
+                
+                $addOrgDiv.find('span.select2-selection--single')
+                .css({'min-height': '32px'});
+                            
+                $shareWithOrgDiv.append($addOrgDiv); // put addOrgDiv into shareWithOrgDiv 
+            } else {
+                $shareWithOrgDiv.append('<p style="margin-top: 18px;">You must be the owner to request to add this narrative.</p>')
```
briehl: semicolon

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSharePanel.js
```diff
@@ -181,6 +220,104 @@ define ([
...
+                tabSwitch($(this), $tab1);
+                $shareWithOrgDiv.css("display", "inherit");
+                $shareWithUserDiv.css("display", "none");
+            });
+            $tabDiv.append($tab2);
+            
+            function createTab(color, text){
+                return $('<div class="shareTab">')
+                .css({'width': '50%', 'display': 'inline-block', 'padding': '10px', 'border': 'solid', 'border-width': '1px 1px 0px'})
+                .css('background-color', color)
```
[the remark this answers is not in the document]
FlyingUnicornSF: -why 2 .css - i figured 1) "color" is a param so separate it to make it obvious how param is used 2) the line 250 is pretty long.... 
I will combine it. 👍

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeSharePanel.js
```diff
@@ -311,18 +459,53 @@ define ([
...
+                    "Authorization": token,
+                    "Content-Type": "application/json",
+                }
+            })
+            .then(response => response.json())
+            .then(response => {
+                if(response.error) {
+                    this.reportError(response);
+                }
+                console.log("Request to Add Narrative:", JSON.stringify(response))
```
briehl: missed a console log line. :)

PR #1454, opened 2019-02-21
Fix "Associate" spelling in share panel

briehl:
Not sure how I missed this in review...


briehl:
Just fixes a typo and removes a console log line. Merging.

PR #1461, opened 2019-03-08
Python 3 update

briehl:
Updates the Narrative interface to use Python 3.

* Now depends on anaconda as an environment manager (not virtualenv)
* Cleaned up all backend code to adhere to Python 3 standards, if not all idioms
* Updated all tests to continue to run with nose, but python3 code
* Updated README with info on how to install (should be easier and more stable now, I think...)
* Builds up a new narrbase:6.0 image, which depends on kbase/python:python3
* The main image now deploys the narrative internally inside the 'base' conda environment
* There might still be some idiosyncrasies with how Jupyter starts and behaves - this needs A LOT of extra testing and QA on CI before release!


.travis.yml
```diff
@@ -24,20 +24,28 @@ branches:
...
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda info -a
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
+  - source activate test-environment
+  # - pip install -r src/requirements.txt
+  # - pip install coveralls scikit-learn[alldeps] pandas
+  # - pip install --no-dependencies clustergrammer_widget
```
JamesJeffryes: Do you want to keep these commented lines?

Dockerfile
```diff
@@ -50,12 +43,12 @@ WORKDIR /kb/dev_container/narrative
 RUN mkdir -p /kb/deployment/ui-common/ && ./src/scripts/kb-update-config -f src/config.json.templ -o /kb/deployment/ui-common/narrative_version
 
 # Install Javascript dependencies
-RUN npm install && ./node_modules/.bin/bower install --allow-root --config.interactive=false
+RUN npm install -g grunt-cli && \
+    npm install && \
+    ./node_modules/.bin/bower install --allow-root --config.interactive=false
 
 # Compile Javascript down into an itty-bitty ball unless SKIP_MINIFY is non-empty
 # (commented out for now)
```
JamesJeffryes: It's not commented out anymore is it?

narrbase-image/Dockerfile
```diff
@@ -18,13 +20,20 @@ RUN conda update -n base conda && \
...
 
+# Install apt-get prereqs for node, R
+RUN apt-get update && \
+    apt-get install -y gfortran gnupg
+
+# Install nodejs at a useful version
+RUN curl -sL https://deb.nodesource.com/setup_${NODEJS_VERSION}.x | bash - && \
+    apt-get install -y nodejs
+
 # Install misc R packages not available on Coda
```
JamesJeffryes: Conda

scripts/install_narrative.sh
```diff
@@ -71,29 +69,9 @@ function usage () {
     printf "If the virtualenv does not yet exist, the script will attempt\n\tto make one for you with default options.\n"
```
JamesJeffryes: Is this true anymore?

scripts/install_narrative_docker.sh
```diff
@@ -22,26 +22,19 @@ function console () {
...
 
-# Install KBase data_api package
-# ------------------------------
-# git clone https://github.com/kbase/data_api -b develop
-# cd data_api
-# pip install -r requirements.txt
-# $PYTHON setup.py install
-# cd ..
-# rm -rf data_api
-
```
JamesJeffryes: Good riddance!

src/biokbase/narrative/contents/kbasewsmanager.py
```diff
@@ -177,11 +177,11 @@ def exists(self, path):
         return self.file_exists(path)
 
     def _wsobj_to_model(self, nar, content=True):
-        nar_id = u'ws.{}.obj.{}'.format(nar['wsid'], nar['objid'])
-        model = base_model(u'{} - {} - {}'.format(nar['saved_by'], nar_id, nar['name']), nar_id)
-        model[u'format'] = u'json'
-        model[u'last_modified'] = nar[u'save_date']
-        model[u'type'] = u'notebook'
+        nar_id = 'ws.{}.obj.{}'.format(nar['wsid'], nar['objid'])
+        model = base_model('{} - {} - {}'.format(nar['saved_by'], nar_id, nar['name']), nar_id)
```
JamesJeffryes: In future, I think we should use f-strings where possible but I doubt it's worth changing all the existing code.

src/biokbase/narrative/contents/kbasewsmanager.py
```diff
@@ -210,35 +210,35 @@ def get(self, path, content=True, type=None, format=None):
         path = path.strip('/')
         model = base_model(path, path)
         try:
-            if self.exists(path) and type != u'directory':
+            if self.exists(path) and type != 'directory':
```
JamesJeffryes: Shadows a built-in but probably beyond the refactor's scope

src/biokbase/narrative/jobs/appmanager.py
```diff
@@ -768,15 +768,15 @@ def _map_inputs(self, input_mapping, params, spec_params):
...
+                    bck_slash = "\u244A"
+                    fwd_slash = "\u20EB"
                     temp_string = target_prop.replace("\\\\", bck_slash)
                     temp_string = temp_string.replace("\\/", fwd_slash)
                     temp_path = []
                     for part in temp_string.split("/"):
                         part = part.replace(bck_slash, "\\")
                         part = part.replace(fwd_slash, "/")
-                        temp_path.append(part.encode('ascii', 'ignore'))
+                        temp_path.append(part.encode('ascii', 'ignore').decode("ascii"))
```
JamesJeffryes: Do we want to strip non-ascii chars or should we support them now?

src/biokbase/narrative/jobs/batch.py
```diff
@@ -274,7 +274,9 @@ def generate_input_batch(app, tag='release', **kwargs):
     batch_inputs = list()
     param_ids = input_vals.keys()
     product_inputs = [input_vals[k] for k in param_ids]
-    batch_size = reduce(lambda x, y: x*y, [len(p) for p in product_inputs])
```
JamesJeffryes: Appreciate you refactoring out the map and reduce expressions

.travis.yml
briehl: Nope. Deleted!

Dockerfile
briehl: Right

narrbase-image/Dockerfile
briehl: Done. conda-forge has the latest nodejs LTS, so that should be ok.

scripts/install_narrative.sh
briehl: Nope, removed that text and tweaked the rest, and some other options.

scripts/install_narrative_docker.sh
briehl: 🎉

src/biokbase/narrative/contents/kbasewsmanager.py
briehl: Yes! Maybe that can be a separate PR.

src/biokbase/narrative/contents/kbasewsmanager.py
briehl: Yeah, that should get changed. That module is inherited from a base class in the Jupyter stack that deals with file management, etc. I suspect that API's been updated along the way, too.

Probably a separate PR.

src/biokbase/narrative/jobs/appmanager.py
briehl: That's all an ugly bit of parsing to handle some side cases that I don't think are documented anywhere, and I'm really not sure anything uses.

Short version = the `target_property` field in an app spec can be set to something like "foo/bar/baz", and it'll put the value in a dictionary like {foo: {bar: {baz: <value>}}}. That weirdness up there handles including slashes in the name, especially escaped slashes. So a `target_property` of "foo/bar\/baz" should be {foo: {bar/baz: <value>}}.

Whoever wrote this bit (Roman, I think), did some temporary things by turning slashes into their unicode values, then used the `encode` function to restore them. In Python 3, `encode` always returns a `bytes` object, so we have to decode that (again) to make it a useful string.

For that purpose, everything always supported ascii, and would fail at the Catalog service anyway. (er, probably. Either way, it's expected that the app specs this code sees are up to snuff).

src/biokbase/narrative/jobs/batch.py
briehl: 👍  (they didn't work anyway 😬)

narrbase-image/Dockerfile
JamesJeffryes: 😆 I'm totally in-favor of this change but I actually was just talking about the spelling of the comment "Install misc R packages not available on *Coda*"

briehl: Ha. Oops! Fixed it. 😝

briehl:
This should be pulled into a shiny new branch, so closing this for now.

PR #1472, opened 2019-04-10
Gtag update

FlyingUnicornSF:


kbase-extension/kbase_templates/page.html
```diff
@@ -54,8 +53,15 @@
...
-        'cookie_domain': 'kbase.us'
-      })
+        'username': '{{userName}}',
+        'Page_location': document.location,
+        'page_path': document.location.pathname,
+        'page_title': document.location.pathname
+      });
+      gtag('set', {'user_id': '{{userName}}'});
+      if (document.location.hostname.includes('narrative')){
+        gtag('config', 'AW-753507180');
```
briehl: So that only gets set in production? And shouldn't be set otherwise?

FlyingUnicornSF: AW tag is for google Ads so it won't be any good used for what internal dev humans do....

PR #1477, opened 2019-04-26
fix(?) string sanitization test

briehl:


briehl:
Agreed to both of those. Walked through that module and updated let -> const where appropriate. And fixed a few " -> ' issues (in one of the tests, it's explicitly testing the use of single quotes or double quotes and how they get escaped, so that had to be inverted).

PR #1482, opened 2019-05-29
Lazy-load the Narratives tab

briehl:
Took a little tweaking to a few places.

In the parent widget class for the Narrative Manager widget, I added an `activate()` method. This gets invoked whenever that tab gets clicked on by the user. It's a no-op in the parent widget, so each child widget can extend it. Only the Narrative manager widget uses it to do its first rendering, then it doesn't do so again.


briehl:
Merging to build and test on CI. This loads correctly across all browsers.

PR #1483, opened 2019-06-04
Data Panel Speedup

briehl:
This refactors big chunks of the data panel slideout to (hopefully) speed it up, and to make it more legible.

I didn't rewrite everything, as that would take forever. But here's the highlights:
* Factored the main My Data and Shared With Me tabs into a separate module, `dataBrowser`
  * This uses a new NarrativeService function, `list_all_data`, to get all user data, pre-filtered, and sorted.
* Wrote a small `dataProvider` to do all the data fetching for a single workspace - anything that needs to know what's in the current workspace can use this.
* Changed the filters around to do a specific data lookup, instead of fetching everything again and parsing through them.
* Modified the data card to be a little more detached.

Still to-do:
* Tests for dataProvider.
* Ensure the data viewer limits are observed and in place.


briehl:
Ugh. Silly test apparatus. It should be mostly updated, but on the way it exposed a bunch of poorly maintained modules that needed a little TLC.

briehl:
Going to merge this as a prototype fix so eyes can get on it on CI, we can come back and do a deeper review soon.

PR #1486, opened 2019-06-06
Merge for release of v3.11.2

briehl:


briehl:
We're not using dockerhub to do builds.

PR #1494, opened 2019-06-26
update release notes

briehl:


briehl:
just a little release note text change, has no code effect, so I'm just gonna merge

PR #1498, opened 2019-07-17
DO NOT MERGE - Py2to3 narrative search script

briehl:
**This is based on the Python 3 branch of the Narrative**  
So, don't merge this in to develop yet.

Also includes the Narrative metadata updater/fixer script, since this has some shared code.

So I'll close that PR.


briehl:
This runs now, and has been run on prod. It's not open code, and needs some super special privileges. Merging now for posterity, and in case it's needed again.

PR #1503, opened 2019-08-27
Exporter

briehl:
Shinier version of the exporter that can do app cells and other various types of modern cells.

Todos for this PR:
(removed - now in a separate service)
~~[x] docs on how to run, requirements, limitations, etc.~~
~~[ ] test the direct_html report case (also find an app that does that, or make one...)~~
~~[ ] better tests to investigate resulting html file.~~
~~[ ] better configuration to set paths to images, icon fonts~~
~~[ ] configuration to point to KBase for jquery, bootstrap, and font-awesome (not cloudflare's cdn)~~


briehl:
Plans change! Most of this was migrated to the https://github.com/kbaseapps/StaticNarrative service.

The remaining parts are all front end changes to act as an interface to that service.

briehl:
This is all prototype code that doesn't impact other things, so I'll go ahead and merge so we can ~~break~~ try it out on CI.

PR #1506, opened 2019-09-04
JobManager error on startup

briehl:
NJS 1 had an occasional (very rare these days) problem where it would throw an error on returning a job. These were captured during the `initialize_jobs` function and forwarded to the browser.

This is different from a job being in an error (or the new terminated) state. Those are still valid and shouldn't raise an overall exception.

I'm guessing things will be a little more stable, too, so this just eliminates that individual job lookup error, especially now that they're all looked up in a batch, so the lookup will all either work at once or fail all together.

This also fixes some issues with the `list_jobs` function.


briehl:
@Tianhao-Gu

PR #1510, opened 2019-09-05
fix result tab

Tianhao-Gu:
Screen Shot 2019-09-04 at 11 14 04 PM


Tianhao-Gu:
Fix job logs as well. 

Screen Shot 2019-09-04 at 11 38 42 PM

Tianhao-Gu:
Fix job status as well

Screen Shot 2019-09-05 at 12 40 06 AM

Tianhao-Gu:
Fix job status as well

Screen Shot 2019-09-05 at 12 40 06 AM

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -874,6 +874,8 @@ def _get_all_job_states(self, job_ids=None):
                 state['run_id'] = self._running_jobs[job_id]['job'].run_id
                 if status == 'finished':
                     self._completed_job_states[state['job_id']] = dict(state)
+                state['job_input'] = literal_eval(state.get('job_input', '{}'))
```
briehl: This has bothered me for a few days, and I think I finally figured out why - the API shouldn't be returning goofy non-standard JSON strings that require an AST to eval, but the actual unpacked structure. I think that should be revisited.

nbextensions/appCell2/widgets/tabs/jobStateViewer.js
```diff
@@ -26,14 +26,26 @@ define([
...
+                break;
             case 'suspend':
                 label = 'error';
                 color = 'red';
                 break;
+            case 'error':
+                label = 'error';
+                color = 'red';
+                break;
             case 'canceled':
```
briehl: Is "canceled" still an option from the API? What are the states that are available?

nbextensions/appCell2/widgets/appCellWidget.js
```diff
@@ -1786,9 +1786,9 @@ define([
                     // to inform us about what processing stage the
                     // error occurred in -- we need to inspect the job state.
                     var errorStage;
-                    if (jobState.exec_start_time) {
+                    if (jobState.running) {
```
briehl: Can this be figured out by the job status? And not guessing from which fields are present?

nbextensions/appCell2/widgets/tabs/jobStateViewer.js
Tianhao-Gu: no. `canceled` status is no longer available. I updated the state viewer code to remove all `canceled` instances.

src/biokbase/narrative/jobs/jobmanager.py
Tianhao-Gu: 👍 updated ee2 API.

nbextensions/appCell2/widgets/appCellWidget.js
Tianhao-Gu: hmm it's already in `jobState.status` case block.

briehl:
I think that failing frontend test is a red herring right now. There's other things to do on that branch, anyway, that should ameliorate tests. Let's go ahead and merge this in now.

PR #1511, opened 2019-09-18
fixed link to kbase-how-to page to link globus account

FlyingUnicornSF:
the template makes two separate links to globus account and kbase instruction page how to link globus account.  Both have .globus_link class name, and on click event was handling only when the link was redirecting to user globus account.  Resulting the redirect page to stall without being able to open globus account.  PTV-1308 for mode details. 

added if/else statement to check if the target href is to the kbase.us or not.


kbase-extension/static/kbase/js/widgets/narrative_core/upload/fileUploadWidget.js
```diff
@@ -46,21 +46,28 @@ define([
...
+                        token: Runtime.make().authToken()
                     });
+                    var globusWindow = window.open('', 'dz-globus');
+                    globusWindow.document.write('<html><body><h2 style="text-align:center; font-family:\'Oxygen\', arial, sans-serif;">Loading Globus...</h2></body></html>');
+                    stagingServiceClient.addAcl()
+                        .done(() => {
+                            window.open($(e.target).attr('href'), 'dz-globus');
+                            return true;
+                        });
+                } else if((e.target.href).includes("http://kbase.us/transfer-data-from-globus-to-kbase")){
```
briehl: This can just be `else {`, as that's the only other case you're checking.

briehl: This can just be an `else {` - that's the only other case you're tracking here. You're really only adding special behavior if that isn't the link, otherwise it should proceed as normal.

kbase-extension/migrated
```diff
@@ -0,0 +1 @@
+2019-09-18T17:37:24.457793
```
briehl: Nothing you did, but this file should probably just be deleted and added to `.gitignore`.

kbase-extension/static/kbase/js/widgets/narrative_core/upload/fileUploadWidget.js
```diff
@@ -46,21 +46,28 @@ define([
...
+                    window.open("http://kbase.us/transfer-data-from-globus-to-kbase", "_blank")
```
briehl: Probably best to just use the link in the template - `e.target.href`, so if that template ever gets updated you don't have to copy/paste it everywhere.

briehl: Needs a semicolon.

kbase-extension/static/kbase/js/widgets/narrative_core/upload/fileUploadWidget.js
```diff
@@ -46,21 +46,28 @@ define([
...
                     });
+                    var globusWindow = window.open('', 'dz-globus');
+                    globusWindow.document.write('<html><body><h2 style="text-align:center; font-family:\'Oxygen\', arial, sans-serif;">Loading Globus...</h2></body></html>');
+                    stagingServiceClient.addAcl()
+                        .done(() => {
+                            window.open($(e.target).attr('href'), 'dz-globus');
+                            return true;
+                        });
+                } else {
+                    window.open(e.target.href, "_blank")
```
briehl: semicolon. :)

PR #1513, opened 2019-09-19
Compatible with ee2

Tianhao-Gu:


src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -195,10 +195,10 @@ def list_jobs(self):
 
             if not len(status_set):
                 return "No running jobs!"
-            status_set = sorted(status_set, key=lambda s: ObjectId(s['_id']).generation_time.timestamp())
+            status_set = sorted(status_set, key=lambda s: ObjectId(s['job_id']).generation_time.timestamp())
```
briehl: This can just sort on s['created'] now, right?

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -195,10 +195,10 @@ def list_jobs(self):
...
 
             for status in status_set:
-                status['creation_time'] = datetime.datetime.strftime(ObjectId(status_set[0]['_id']).generation_time, "%Y-%m-%d %H:%M:%S")
+                status['creation_time'] = datetime.datetime.strftime(ObjectId(status['job_id']).generation_time, "%Y-%m-%d %H:%M:%S")
```
briehl: Same here? we can just use 'created' instead of 'creation_time', I think?

briehl:
Hey, not sure if you're done working on this or not, but I threw in a couple things I saw.

src/biokbase/narrative/jobs/jobmanager.py
Tianhao-Gu: 👍

src/biokbase/narrative/jobs/jobmanager.py
Tianhao-Gu: 💯

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -866,7 +868,7 @@ def _get_all_job_states(self, job_ids=None):
 
         for job_id in jobs_to_lookup:
             state = fetched_states.get(job_id, {})
-            state['job_id'] = state.get('_id')
+            state['job_id'] = state.get('job_id')
```
briehl: that's a little redundant.

Tianhao-Gu: ahhh yea. removed it.

PR #1519, opened 2019-10-01
Bugfix for dynamic dropdown result processing

MrCreosote:
Fixed two bugs:

The dynamic dropdown was hardcoded to only handle the case where the selection
items were in the first position of the return array and were at the first level
of the object at that position.
Now the position of the object in the return array and the path into the object
to the selection items can be specified.

Some dynamic service functions don't like empty input, but the DD will always
query the server regardless of the input contents. This behavior can now
be configured.

These changes require https://github.com/kbase/narrative_method_store/pull/74

There are no meaningful tests for the DD, and developing  a test harness
for the DD is way out of scope, so the changes were tested manually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ static/dist
 my_narrative_venv/
 narrative_venv/
 new_narrative_env_2/
+js-coverage
 
 # From https://github.com/github/gitignore/blob/master/Python.gitignore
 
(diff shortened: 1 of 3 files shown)


MrCreosote:
No idea why the build failed.looks like a travis internal thing?

MrCreosote:
Also I tested generating the app cell and getting the list of putative names in Bill's app spec tester

PR #1521, opened 2019-10-02
minor dynamic dropdown fixes

briehl:
1. fetchData needs to always return a Promise (even on a no-op, empty array return)
2. Calling the select2 command again when showing a dynamic dropdown tagged as advanced would cause the dropdown to forget its ajax ability (should probably be double-checked in different browsers, but verified in Firefox and Chrome)


MrCreosote:
LGTM

MrCreosote:
Note, of course, that I am far from expert in this stuff. I am a child in these matters

PR #1527, opened 2019-10-03
remove dead NJS client

briehl:
And a minor narrativeio regression about shared narratives


briehl:
Also crept in a jobmanager.list_jobs bugfix

briehl:
@Tianhao-Gu , mind taking a look? This removes the remaining NJS_Wrapper code and client (and does a little bugfix in narrative saving - non-admin users can't update workspace metadata)

PR #1530, opened 2019-10-07
job log state issues

briehl:
"canceled" isn't a valid state when the job state is "error", but "error" is.


briehl:
Also noting here that we may need to refactor how "terminated" state gets managed - whether it's from user action or admin action.

PR #1533, opened 2019-10-22
update jobmanager handling of timestamps in list_jobs

briehl:


Tianhao-Gu:
PR looks good. We should change the JSON test file just to make sure. 
https://github.com/kbase/narrative/blob/py3-update/src/biokbase/narrative/tests/data/ee2_job_test_data.json#L9-L10

The new format for both `update` and `create` is:
`2019-10-22T22:20:27.569472+00:00`

Tianhao-Gu:
👍  looks good to me.

PR #1534, opened 2019-10-24
fix app error rendering

briehl:


briehl:
Also renders the error message (most commonly a stacktrace) as a <pre> block, so it'll retain formatting.

PR #1536, opened 2019-10-29
Timestamp system variable

briehl:
the token system var removal part is covered in a separate PR, but this builds on it.

### DIFF
--- a/src/biokbase/narrative/app_util.py
+++ b/src/biokbase/narrative/app_util.py
@@ -3,6 +3,7 @@ import re
 import json
 import biokbase.narrative.clients as clients
 import biokbase.auth
+import time
 
 """
 Some utility functions for running KBase Apps or Methods or whatever they are this week.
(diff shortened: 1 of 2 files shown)


briehl:
Once this gets tested and merged, I'll propagate it down to the py3-update branch.

src/biokbase/narrative/app_util.py
```diff
@@ -77,6 +73,12 @@ def system_variable(var):
         except:
             return None
         # TODO: make this better with more exception handling.
+    elif var == 'timestamp_epoch_ms':
+        # get epoch time in milliseconds
+        return int(time.time()*1000)
```
MrCreosote: Can't you, you know, actually get the time in ms rather than multiplying sec by 1K?

briehl: Nope. Either seconds or (in python 3.7) ns. The seconds function has floating point accuracy out to ns, though, depending on the OS. On the ones I've tried (and on what the Narrative runs), it's fine.

https://docs.python.org/3/library/time.html#time.time

MrCreosote: Right, I forgot it returned floating point. Sorry for the noise.

PR #1538, opened 2019-11-04
Add option for timestamps in narrative system variables for apps

briehl:


briehl:
Following on to the other timestamp option on the develop branch.

PR #1541, opened 2019-11-06
ee2 tiemfields update

Tianhao-Gu:
update ee2 clients and fix job status due to ee2 time fields update. 

Screen Shot 2019-11-06 at 1 37 32 PM


Tianhao-Gu:
narrative should show correct job status with changes

Screen Shot 2019-11-06 at 2 10 04 PM

Tianhao-Gu:
@briehl Do you mind taking a look?

Tianhao-Gu:
`queued for` shows very long, `22days` in this case. Due to a bug in ee2 which will be fixed in https://github.com/kbase/execution_engine2/pull/95

PR #1545, opened 2019-11-08
fix due to status update

Tianhao-Gu:


Tianhao-Gu:
Screen Shot 2019-11-08 at 12 40 05 PM

nbextensions/appCell2/widgets/appCellWidget.js
```diff
@@ -2027,7 +2027,7 @@ define([
             widgets.runClock = RunClock.make({
                 prefix: 'for '
             });
-            var creation_time = new Date(jobState.created);
+            var creation_time = new Date(jobState.created).getTime();
```
briehl: What is the value in `jobstate.created`? I thought it was already ms since epoch?

briehl: If so, that doesn't need to change.

briehl: Unless this is for backward compatibility? There should probably be a function that ensures that for other parts of the job state anyway. If we're going forward with always returning ms since epoch as the times, then this is redundant, and should just be `var creation_time = jobState.created;`

Tianhao-Gu: 👍 sounds good. Yea. time fields are returned as million seconds since epoch.

PR #1546, opened 2019-11-12
Fix for PTV-1315

eapearson:
This fix addresses PTV-1315, which was a response to a public ticket which reported that a user was unable to adequately access the elements of an RNASeq reads alignment set. It ends up the problem was because the viewer responsible, kbaseGenericSetViewer, was never fully implemented and generated an empty display. Since the user who originally reported the error stopped responding to the public ticket eventually, and the ticket was closed, this fix should not be strictly interpreted as solving their specific issue, but the issue any user would encounter when attempting to view an object which would be viewed with this viewer.

Although this set of changes may seem large, it is focused on a single widget, kbaseGenericSetViewer.

This widget is supposed to handle objects of the various set types in the KBaseSets workspace type module. However, it was never fully implemented. 

This set of changes adds implementations for KBaseSets.ReadsAlignmentSet and KBaseSets.AssemblySet. There are other unimplemented sets, and some types in KBaseSets have other viewers not related to kBaseGenericSetViewer.

Most of the work was in implementing the viewers, which is done in preact. The job of kbaseGenericSetViewer.js becomes that of a wrapper around a main component, and providing the appropriate props to it.

The viewer and support components were added in a new component directory 'preact_components' (it looks like 'components' is already used by Jupyter). Each module in that directory implements and returns a single preact component.

I've evaluated these changes locally, but due to time constraints ("dev-poc" duty with limited time), have not done testing across services or browsers.


briehl:
This brings up some interesting issues. I'd love to see (P)React get used to make Narrative components, and it might make it easier to port things to Jupyterlab, when we get time for that again.

Also, Preact was removed from the Jupyter notebook codebase not long ago in favor of the full thing https://github.com/jupyter/notebook/pull/4726 Glancing through their recent code, I don't think it's really being used for much, and what is used references React.createElement and ReactDOM.render directly, rather than using JSX, so I don't know if they worked out compilation issues yet.

However, that's only on the Jupyter notebook 6+ path, which removed Python 2 compatibility. So we can't use it in this patch. And using Preact here version-locks us even tighter. Not a deal-breaker, I don't think, as your code looks like we can mostly just swap out the Preact dependency for React and be ok.

Also, putting Preact in the global space was a hacky workaround I did on the last Jupyter notebook version update. Their bundler caused some problems and it wasn't loading properly. I'd be up for fixing that issue now that we might make use of it.

briehl:
I suppose one option would be to embrace this change now, which would mean a few steps.
1. Do this on the py3-update branch.
2. Ensure that we're tracking the latest Jupyter notebook on that branch.
3. Ditch the Preact stuff and go with the React that they use.
4. Jump through the hoops to get everything to build properly.

We're still on Notebook 5.7.4 on the py3-update branch, I'll see what the latest release (6.0.1) looks like there.

briehl:
Updating to release 6.0.2 (the actual latest) seems to mostly just work. I had to tweak the `page.html` template to have their change that loads up React, ReactDOM, and `create-react-class`, something else that they include. But everything seems to just load up fine.

Maybe that's a path toward using React in the Narrative?

briehl:
Some notes for posterity - 
Spoke with @eapearson yesterday about the upcoming change to the latest version of Notebook, and its swap of Preact for React. We decided to just let this PR rest for now until that update happens in the Narrative repo (and the various branch wrangling/mangling for the upcoming release). Once that happens, this'll get changed to use React instead.

eapearson:
Should have pinged you @briehl.
The main question is whether react has been used in the Narrative yet, and in either case whether you have any information on using it in the codebase.
If not, I'll investigate further.

eapearson:
never mind, I found React and ReactDOM on the window object.

eapearson:
@briehl I have the preact->react port done - but this was done before the convention of working from feature branches directly in kbase/narrative.

Would you rather I just close this, push the branch up, and create a new PR from it?

I would copy the initial comment, and some additional comments.

eapearson:
@briehl I've made some additional progress to make the code quality tools and tests happier.
I don't see any GHA running on recent updates to this PR, but they did yesterday when I updated it.
Perhaps I should recast this as a PR from a branch?

eapearson:
@briehl It really isn't a problem to redo this on a branch directly in the kbase repo, so I'll do it.

The React stuff is "good", as in a step in the right direction, but the actual implementation is not as exciting as it could be, because Jupyter and the Narrative aren't set up for efficient React development. No JSX, no development build.

Without JSX the actual markup is quite awkward. One direction I didn't explore in this round is that, as for preact, react may be supported by the `html` library, which let's one use a markup syntax very close to JSX within Javascript string templates. I just converted dozens of components in kbase-ui from preact htm templates to full React TSX (typescript), and in most cases the conversion was trivial, just a few minutes. More complex cases, and when one wants to take better advantage of the React architecture over some rather arbitrary Javascript practices (especially involving data fetching and data flow), do require some heavier redesign. 

By lack of development support I mean for instance that the react library utilized is the minified one, so debugging is not as good. And the library is installed globally rather than as a pure module. So, for instance, dependency management involving React and ReactDOM won't work.

PR #1547, opened 2019-11-12
Fixes SCT-2291, in which a url hash (fragment) causes app cell failure.

eapearson:
This fix is very simple. I discovered it when implementing the fix for PTV-1315 and inadvertently formed urls which included a url fragment. Initially I presumed I had introduced a bug, but the error occurred in the app cell extension itself, which is a big stretch from a viewer widget.

The changes here are primary:
- use the window location pathname not the href -- all of the information the narrative needs is in the path itself; this also ensures that the fragment/hash is not present
- improved, perhaps, or at lest simplified the regex usage

I left quite a few notes in the code to address these concerns:
- there is no explicit handling of a non-match in the narrative url. it looks like the separate parsing and interpretation of the url in the back end catches most if not all cases of malformed urls, so it may not be possible to trigger this condition in the front end code; even then, I don't know how the narrativeConfig code can cause an error message to be displayed, other than a redirect.
- there is a construction of narrativeRef in the old code, but as I traced that through the system it does not appear that it is being used; leaving it as null has no effect, and I think it can be completely removed from the codebase.

### DIFF
--- a/kbase-extension/static/kbase/js/narrativeConfig.js
+++ b/kbase-extension/static/kbase/js/narrativeConfig.js
@@ -39,17 +39,56 @@ define([
     // Get the workspace id from the URL
     var workspaceId = null,
         objectId = null,
-        narrativeRef = null,
-        m = window.location.href.match(/(ws\.)?(\d+)((\.obj\.(\d+))(\.ver\.(\d+))?)?$/);
+        narrativeRef = null;
+    // m = window.location.href.match(/(ws\.)?(\d+)((\.obj\.(\d+))(\.ver\.(\d+))?)?$/);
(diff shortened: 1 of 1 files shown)


briehl:
This is great!

A non-match in the Narrative URL will simply fail to get served up by the Tornado handler extension we use here: https://github.com/kbase/narrative/blob/develop/src/biokbase/narrative/handlers/narrativehandler.py#L101-L108

This means that any path that doesn't look like /ws.###.obj.### or /### or /ws.### will throw a 404 and not even bring up the Narrative front end code. I think not catching that error will be fine, since it shouldn't even execute that code.

PR #1548, opened 2019-11-12
Update dependency versions

briehl:
Jupyter notebook -> 6.0.2
IPython -> 7.9.0
IPywidgets -> 7.5

This mainly removes the Preact dependency and replaces it with React, ReactDOM, and create-react-class, a legacy API for creating React classes without ES6.

https://reactjs.org/docs/react-without-es6.html


briehl:
This is also going to mean updating the kbase/narrbase image before merging. And bumping the version of narrbase being used elsewhere in the build process.

briehl:
Closing this, as it's in the `py3-pre-ee2` branch (including the base image update). When that's merged, we can merge `py3-pre-ee2` into `py3-update`

PR #1551, opened 2019-11-15
update ee2 calls

Tianhao-Gu:


Tianhao-Gu:
@briehl We updated some ee2 clients. By default, `check_jobs` and `check_workspace_jobs` are returning a list instead of a dictionary.

PR #1554, opened 2019-12-10
Fold in changes from develop since EE2 was worked on

briehl:
There's a lot. But this shouldn't include EE2 or any of its libraries or changes. Thus it should work against the production Narrative job service.


eapearson:
I can't really view the code changes, but did install and run it locally against prod (using the container method) and it seems fine.

PR #1555, opened 2019-12-10
PTV-966 fix float validator

briehl:
`if (min && min > value)` was being ignored if `min` is 0. So that's treated as falsy.

And fun fact, `isNaN(null)` evaluates to `false`. So it'll catch `undefined` and strings and whatnot, but not null? I ❤️ Javascript.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/appWidgets2/validators/float.js
+++ b/kbase-extension/static/kbase/js/widgets/appWidgets2/validators/float.js
@@ -29,10 +29,10 @@ define([
         if (!isFinite(value)) {
             return 'value must be a finite float';
         }
-        if (max && max < value) {
+        if (max !== null && !isNaN(max) && max < value) {
             return 'the maximum value for this parameter is ' + max;
         }
-        if (min && min > value) {
+        if (min !== null && !isNaN(min) && min > value) {
             return 'the minimum value for this parameter is ' + min;
         }
     }
@@ -68,7 +68,7 @@ define([
         });
     }
 
-    // For text values, there is 
+    // For text values, there is
     // function validate(value, spec) {
     //     try {
     //         var nativeValue = importString(value);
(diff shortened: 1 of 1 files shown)


kbase-extension/static/kbase/js/widgets/appWidgets2/validators/float.js
```diff
@@ -29,10 +29,10 @@ define([
         if (!isFinite(value)) {
             return 'value must be a finite float';
         }
-        if (max && max < value) {
+        if (max !== null && !isNaN(max) && max < value) {
```
eapearson: the purpose of the test for min/max is to disable the associated check if not supplied (or supplied as null). If I were rewriting this I'd probably wrap the entire test in a `if (typeof max === 'number') { .. }` , like
```javascript
if (typeof max === 'number') {
    if (value > max) {
        return 'Value exceeds allowed maximum of ' + max;
    }
}
```

eapearson: I can't comment on lines not changed, but rather than isNaN I think the function should just ensure that the value is the number type. isNaN will coerce values before checking if they are numbers (e.g. `isNaN("123")`)

briehl: We could also go the other way around and coerce value into being a number before testing. Or, since this is a validator, just assume that whatever's doing validation knows what it's going for, and err on the side of simplicity and specificity. I'll go ahead and do that.

eapearson: Nice. Yeah, I used to prefer validation to return a normalized (coerced, transformed) if the validation is successful, but now I prefer just the simpler validation, with the presumption that if the values pass validation, they obey the contract and are perfectly usable as-is. In TS, though, I've found that it is useful is to return the value because having the associated typing is critical from that point forward in the code.

briehl:
After chatting with @eapearson we decided this is ok.

PR #1556, opened 2019-12-11
make the default behavior to do empty input search in dyn dropdown

briehl:
DO NOT MERGE

Also pending in this PR is some looser validation, so if it gets a number or a string, it just stays as a number or a string. The idea here is that whatever the service provides should be passed along to the app. As long as its an atomic type (a number or string). Maybe booleans can be investigated in the future if we need, but expanding this to arrays or structures would just be ripe for abuse.

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
+++ b/kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
@@ -71,6 +71,10 @@ define([
             userId = runtime.userId(),
             eventListeners = [];
 
+        if (dd_options.query_on_empty_input === undefined) {
+            dd_options.query_on_empty_input = 1;
+        }
+
         /**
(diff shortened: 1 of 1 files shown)


kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -71,6 +71,10 @@ define([
             userId = runtime.userId(),
             eventListeners = [];
 
+        if (dd_options.query_on_empty_input === undefined) {
```
eapearson: This should be use a type test, like `typeof dd_options.query_on_empty_input === 'undefined'`.
I know, I've used a test against the global `undefined` value a million times, but I try not to now.

briehl: Sure, I get it. It's safer. Changed.

PR #1557, opened 2019-12-12
Minor error formatting fix

jayrbolton:
Make the "Details" section of the app widget's error output formatted with a `pre` tag so you can see the proper linebreaks and debug things more easily


jayrbolton:
I manually tested this on a few narratives and it looks good. eg: http://localhost:8888/narrative/notebooks/ws.37757.obj.1

PR #1558, opened 2019-12-13
Remove opacity reduction styles on unselected app cells

jayrbolton:
Removes the opacity reduction on unselected app cells. This one is another annoyance for me, as I'd prefer app cells to be readable without having to click on them. But I understand this change may be more controversial and may not have consensus. Nevertheless, here's a PR for it. Feel free to close if we think users wouldn't like it.


jayrbolton:
Downsides
- active app cell is less prominent
Upsides
- Can read the content of a narrative more easily just by scrolling without requiring clicks

I'm not sure the utility of an active app cell, since you can interact with any cell at any time. But maybe the emphasis on an "active" cell has more importance for a reason I don't know about

jayrbolton:
:skull:

PR #1563, opened 2019-12-18
Update to use search2 in the data search side-panel

jayrbolton:
Some small changes to start using the newer search database and API. This should get deployed and tested on CI first. I'll make some code comments to help review


kbase-extension/static/kbase/js/widgets/narrative_core/publicDataSources/searchDataSource.js
```diff
@@ -22,47 +22,18 @@ define([
 ) {
     'use strict';
 
-    function objectGuidToRef (guid) {
```
jayrbolton: No longer needed as the newer index provides the kbase ID

kbase-extension/static/kbase/js/widgets/narrative_core/publicDataSources/searchDataSource.js
```diff
@@ -22,47 +22,18 @@ define([
...
-            workspaceId: parseInt(m[1]),
-            objectId: parseInt(m[2]),
-            version: parseInt(m[3]),
-            ref: objectRef,
-            dataviewId: objectRef
-        };
-    }
-
     function parseGenomeIndexV1 (item) {
-        return {
```
jayrbolton: searchapi2 uses a totally separate database from previous search, so index versions have been reset to 1 rather than continuing from a previous version. When we update the config.json to user searchapi2, then the previous versions are no longer relevant as those indexes are not present in the database anymore

kbase-extension/static/kbase/js/widgets/narrative_core/publicDataSources/searchDataSource.js
```diff
@@ -22,47 +22,18 @@ define([
...
             genome_source: item.data.source,
             genome_source_id: item.data.source_id,
             scientific_name: item.data.scientific_name,
             domain: item.data.domain,
             num_contigs: item.data.num_contigs,
-            num_cds: item.data.cdss,
-            num_features: item.data.feature_counts.gene,
-            ws_ref: objectGuidToRef(item.guid),
+            num_cds: item.data.cds_count,
+            num_features: item.data.feature_counts && item.data.feature_counts.gene,
```
jayrbolton: item.data.feature_counts is sometimes null or undefined. When it's null, the UI displays it as "n/a"

jayrbolton:
The only enhancement I'd like to make for the refseq search feature is to be able to do prefix or partial match searching. But that's an update I can make in searchapi2 and is independent of this PR.

jayrbolton:
Also, the only place I can see where searchapi is used was in the public data refseq section of the data side-panel. Let me know if there's another place I'm missing.

briehl:
I don't know much about the details of search, so I'm gonna assume that the property name changes are fine.

Otherwise, this looks reasonable. We can merge after tests pass.

PR #1571, opened 2020-01-09
SCT-401 - fix issue where losing connection results in user being unable to save

briehl:
This scenario has been going on for years. A user logs in, does stuff, goes away for a while (or loses internet for a while), then returns to a disconnected kernel. After reconnecting or restarting the kernel, the user now can't save.

The reason for this was that there are two backend pieces that require their own state for KBase code to run. The kernel (where code gets executed, including app running and managing), and the server (where notebook/Narrative management gets done, including saving and loading). 

Restarting the kernel does nothing to modify the server state. This change updates that so that both the kernel and server state get updated at the same time.

The real change is in `kbaseNarrative.js`. Now, when the kernel signals that its ready (on page load, or after a restart), 3 things happen:
1. Push the current Narrative session info into the server.
2. Push the current Narrative session info into the kernel.
3. Forge the Jobs communication channel between the client and kernel, and have the kernel update its job state.

Previously, #2 and 3 were in separate responses to the same event, they're merged together to avoid a race condition (I'm not sure if that's ever been seen in the wild, but I think I've seen it once or twice if there's a very inconvenient network hiccup).

### DIFF
--- a/kbase-extension/static/kbase/js/api/auth.js
+++ b/kbase-extension/static/kbase/js/api/auth.js
@@ -1,14 +1,14 @@
 define([
     'bluebird',
     'jquery',
-    'util/string',
     'jqueryCookie'
-], function(Promise, $, StringUtil) {
+], function(Promise, $) {
     'use strict';
 
     function factory(config) {
-        var url = config.url;
-        var cookieName = 'kbase_session';
+        const url = config.url,
+            cookieName = 'kbase_session',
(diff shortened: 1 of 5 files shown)


briehl:
Copying notes here from Slack, as effective to-dos for this PR:

* Would be ideal if Narrative just tries to reconnect first, then falls back to restart, automatically via connection monitoring.
* Avoid need for users to do this.
* Should try to clear all "connection failed" messages in header when reconnected.
* Make more clear what Kernel menu options do (maybe a help?)

PR #1576, opened 2020-01-16
Merge changes to develop branch down to py3-update

briehl:
Develop was recently updated with the mess of changes that update to Python 3. This merges those changes back down to the `py3-update` branch. 

The goal here is to keep `py3-update` updated as EE2 work proceeds.

A further PR will extract the exporter work.

I think the idea is to keep `develop` as the primary working branch with fixes that get pushed to production. While EE2 work continues, therefore, `develop` should be periodically merged into `py3-update` and redeployed on CI. 

I'd also propose changing that branch name to `ee2` just to be clear what it's about. Make sense?


briehl:
And now that I've written this, there's a ton of little conflicts that need to get worked out. I'll probably have to close this PR and re-open it later.

PR #1583, opened 2020-02-11
Fix conda command in docs

jayrbolton:
Small readme fix for the conda commands


briehl:
That's from a really outdated conda. Thanks for fixing!

PR #1590, opened 2020-02-12
Fix issue with message handler being out of scope.

briehl:


briehl:
Error's a red herring, Travis is being weird.

PR #1593, opened 2020-02-18
Add help popover, rewrite static narrative ui

briehl:


briehl:
Also bump version to 4.1.0

PR #1600, opened 2020-03-06
Scroll fixing

FlyingUnicornSF:
@briehl, @eapearson  please review

### DIFF
--- a/kbase-extension/static/kbase/js/util/jobLogViewer.js
+++ b/kbase-extension/static/kbase/js/util/jobLogViewer.js
@@ -414,8 +414,9 @@ define([
         }
 
         function requestLatestJobLog() {
-            // while job is running 
+            // only while job is running 
             // load numLines at a time
+            // otherwise load entire log
             let autoState = fsm.getCurrentState().state.auto;
             if(autoState){
                 linesPerPage = numLines; // set it to 10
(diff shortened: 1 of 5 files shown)


src/config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "config": "dev",
```
eapearson: Probably shouldn't check in config file changes.

kbase-extension/static/kbase/js/util/jobLogViewer.js
```diff
@@ -1122,10 +1034,12 @@ define([
...
 
             jobId = arg.jobId;
-
-            var layout = renderLayout();
+            // passing jobId allows the panel to be found by Job ID
+            // with getElementsByClassName
+            var layout = renderLayout(jobId); 
             container.innerHTML = layout.content;
             layout.events.attachEvents(container);
+            panel = document.getElementsByClassName(jobId)[0]
```
eapearson: it would be better to put the job id on a data- element; e.g. `data-job-id`. 
But there is an even better method. You see, since the scope of the Dom query is the entire document, you have know way of knowing whether there is another node with `class={jobId}` or even `data-foo={jobId}`. The only way to ensure that you get the one element you want is to use a unique value, like a uuid, and the best performing way to set it is as the element's id. You would store the uuid within the component (or widget in this case). So something like `this.panelId = new Uuid(4).format()`, and then `id: this.panelId` in the markup function.

briehl: This is especially important, since it's possible for multiple job log viewers to exist on the page, all looking at the same job id. So that query will grab all of those elements with the same job id and operate on them, instead of the single current element that you expect.

briehl:
I turned our collective comments into issues for further work. Gonna merge this and work on an update. Thanks, @FlyingUnicornSF !!

PR #1601, opened 2020-03-10
JobManager Refactor

briehl:
This PR refactors the kernel-side job starting and management into two pieces.
`biokbase.narrative.jobs.jobcomm` handles all the kernel messaging, receiving and sending messages over the "KBaseJobs" channel. Given a message, it parses it into a `JobRequest` object that's then used to handle the messages.

It also gives more sensible access to other kernel components that need to send a message over the channel (at this point, only the AppManager does that) via `JobComm.send_comm_message`.

`biokbase.narrative.jobs.jobmanager` still does the work of wrangling job information (calls to JobComm just call out to its own instance of a JobManager), but now the individual functions only create and return data, they do not send messages.

So the flow goes like this:
1. Browser triggers some event (user submits a job, or opens the log browser, etc.)
2. Browser-side `jobCommChannel.js` sends a message to the kernel.
3. Kernel-side `JobComm` receives that messages and handles it.
4. Kernel-side `JobComm` (optionally) sends a message back over the channel.
5. Browser-side `jobCommChannel.js` gets that message and handles it.

This isn't very different from previous behavior (in fact, it should look to the user like nothing has changed), except it has more complete testing and more consistent error handling.


briehl:
Don't merge yet - haven't finished writing docs.

briehl:
Ready for review

src/biokbase/narrative/jobs/jobcomm.py
```diff
@@ -0,0 +1,326 @@
...
+            raise ValueError("Improperly formatted job channel message!")
+        self.request = self.rq_data.get("request_type")
+        if self.request is None:
+            raise ValueError("Missing request type in job channel message!")
+        self.job_id = self.rq_data.get("job_id")
+
+
+
+class JobComm:
+    """
```
bio-boris: Very nice detailed commentary

briehl:
Got the ok to merge from @bio-boris .

PR #1603, opened 2020-03-19
Add backward compatibility for EE2 job states

briehl:
EE2 job states changed around a bunch of keys. This is probably for the best, and the names and values are more consistent now.

Most of the rest of the app cell has been updated to handle them, but old cells are not backward compatible. This adds that compatibility in the following steps.
1. at cell start, validate the job state (if there's one cached in cell metadata).
2. if that passes validation, hooray! do nothing.
3. if it fails, force the job status listener to fire up (regardless of whether the job is stopped), and request a current job state.
4. If the old one is out of date and the new one is up to date, force a reset of the cell by reinitializing the state machine that handles its state - this forces a redraw of elements that would otherwise remain static with a new job state (like the header with a timer on it).

This is a little bit of a heavy task for the cell, and now there's a validator comparison on every job status update, but I think it's useful and should be easily forward-compatible if there are any changes.


briehl:
A couple notes
1. This is dependent on merging #1601 
2. I realize there's a bunch of dead code in the `common/jobs` module - that seemed the best place to put a validator. Code culling will be in a separate PR, as I've come across some other spots that could be cleaned out.

src/biokbase/narrative/tests/util.py
```diff
@@ -296,5 +296,43 @@ def find_free_port() -> int:
...
 
+def validate_job_state(state: dict) -> None:
+    """
+    Validates the structure and entries in a job state as returned by the JobManager.
+    If any keys are missing, or extra keys exist, or values are weird, then this
+    raises an AssertionError.
+    """
+    # list of tuples - first = key name, second = value type
+    # details for other cases comes later. This is just the expected basic set of
+    # keys for EVERY job, once it's been created in EE2.
```
bio-boris: So jobs launched not in narrative will not have run_id or cell_id, not sure if this matters here.

briehl: That's a good point! I forgot about that. It shouldn't be forced, then, yes.

briehl: I'll address that in #1601

PR #1633, opened 2020-04-09
fix issue with job manager not reinitializing after page reload

### LINKED ISSUE #1618: Job sharing can have problems
Strange things seem to happen when multiple users have an open narrative and a new app is run. Consider users A and B, both with the same Narrative open.

* User A starts a job, user B gets notification that the Narrative has been updated and should be reloaded.
* User B reloads the narrative, sees that a job was started, but doesn't get any updates (and a notification that they aren't allowed?)
* Restarting B's kernel seems to fix it.

This all points to B's JobManager not getting re-inited on page reload.

briehl:
Fixes #1618 

Also adds a log line when the JobManager initializes its set of jobs.


briehl:
Merging this now to test on CI.

PR #1644, opened 2020-04-21
PTV-1446 - fix bug preventing KBaseFeatureValues viewer apps from working

briehl:
The error was in mapping workspace and object names to UPAs. They tried to lookup `None` as a valid object name.

Also updated some `str.format()` into newer style f-strings.

### DIFF
--- a/src/biokbase/narrative/tests/test_widgetmanager.py
+++ b/src/biokbase/narrative/tests/test_widgetmanager.py
@@ -191,6 +191,36 @@ class WidgetManagerTestCase(unittest.TestCase):
         self.assertFalse(upas)
 
     @mock.patch('biokbase.narrative.widgetmanager.clients.get', get_mock_client)
+    def test_infer_upas_nulls(self):
+        """
+        Test infer_upas when None is passed to it as an object name. Fields with None
+        as input should not map to an UPA.
+        """
+        test_result_upa = "18836/5/1"
+        upas = self.wm.infer_upas("testCrazyExample", {
+            "obj_id1": None,
+            "obj_id2": None,
+            "obj_name1": "foo",
+            "obj_name2": "bar/baz",
+            "obj_names": ["a", "b", "c"],
+            "obj_ref1": "1/2/3",
+            "obj_ref2": "foo/bar",
+            "obj_refs": ["7/8/9", "0/1/2"],
+            "ws_name": "some_ws",
+            "extra_param": "extra_value",
+            "other_extra_param": 0
+        })
+        self.assertIsInstance(upas, dict)
+        self.assertNotIn("obj_id1", upas)
+        self.assertNotIn("obj_id2", upas)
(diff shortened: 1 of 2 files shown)


src/biokbase/narrative/tests/test_widgetmanager.py
```diff
@@ -190,6 +190,36 @@ def test_infer_upas_simple_widget(self):
         self.assertIsInstance(upas, dict)
         self.assertFalse(upas)
 
+    @mock.patch('biokbase.narrative.widgetmanager.clients.get', get_mock_client)
```
MrCreosote, line 193: I don't really understand the code here that well, so my review might not be worth much.

src/biokbase/narrative/tests/test_widgetmanager.py
```diff
@@ -190,6 +190,36 @@ def test_infer_upas_simple_widget(self):
...
+        })
+        self.assertIsInstance(upas, dict)
+        self.assertNotIn("obj_id1", upas)
+        self.assertNotIn("obj_id2", upas)
+        self.assertEqual(upas['obj_name1'], test_result_upa)
+        self.assertEqual(upas['obj_name2'], test_result_upa)
+        self.assertEqual(upas['obj_names'], [test_result_upa]*3)
+        self.assertEqual(upas['obj_ref1'], "1/2/3")
+        self.assertEqual(upas['obj_ref2'], test_result_upa)
+        self.assertEqual(upas['obj_refs'], [test_result_upa]*2)
```
MrCreosote, line 221: I don't understand the rules for how `get_mock_client` responds to a request.

src/biokbase/narrative/widgetmanager.py
```diff
@@ -396,7 +396,7 @@ def infer_upas(self, widget_name, params):
         obj_ref_list = list()   # list of tuples, but second is a list of upas
         ws = None
         for param in params.keys():
-            if param in param_to_context:
+            if param in param_to_context and params.get(param) is not None:
```
MrCreosote, line 399: Are there cases where `None` is a valid value for a param? For example, I've used `False, None, True` as a ternary.

briehl: Yes, `None` can be a valid value. But in this case, that block of code is trying to map values onto workspace object identifier. If it sees `None`, it should just accept that there's nothing there and not try to turn it into an object name/id.

MrCreosote: The docs for the method confused me a bit - now I look at the code it's clear that most of the params are going to be ignored.

src/biokbase/narrative/tests/test_widgetmanager.py
briehl: It's pretty simple. It combines all "clients" into a single module that returns mostly the same thing. It's intended to simulate a valid response to an expected valid request.

PR #1647, opened 2020-04-28
Update token management

briehl:
The original goal here was to make sure the `narrative_session` token gets removed on logout, but the scope needed to expand to the following:
* add samesite and secure properties to cookies (when appropriate, not for localhost use)
* remove dependency on the jquery-cookie package (it's not supported any longer, and we only use it as a quickie short hand)
* add some fairly simplistic cookie management
  * this really only gets used to set a developer's auth token cookie, to remove cookies on logout, and to fetch the auth cookie if it's present. Really simple stuff.

(TODO)
* remove dependency on jquery in the auth api - it makes a `$.ajax` request which can be pretty easily substituted with a `fetch`
* tests


briehl:
Test done, leaving `$.ajax` in for now, as converting it to either a pure XMLHTTPRequest or fetch + polyfill (we have very few IE users, but not 0, and fetch isn't supported in IE) isn't worth the time right now.

PR #1650, opened 2020-05-19
PTV-1515 Update advice to more useful messsage

bio-boris:
Update advice to more useful messsage.
It would be great to get a link to miriam's common errors page or a guide on what to do for errors linked as well, once that page exists. 
Also I guess we shouldn't use email anymore as per https://kbase.slack.com/archives/C5Z8X8DS4/p1589233536170200?thread_ts=1589233498.169700&cid=C5Z8X8DS4

Or maybe thats for outreach actually.


bio-boris:
Updated as per https://kbase-jira.atlassian.net/browse/PTV-1515

PR #1652, opened 2020-05-21
Fixes SCT-2559: Add info tab to App Cells

dakotablair:
This pull request Adds the info tab to app cells in narratives. Right now it is configured to always be present, regardless of job or cell state.


briehl:
I don't really like the `toBoolean` function, but as it's just being moved, and rewriting things that use it is out of scope, this seems fine.

PR #1662, opened 2020-06-02
Adding basic unit tests for app cell info tab.

dakotablair:
This pull request is to demonstrate tests running on code within the `nbextensions` directory without radically changing the current configuration within the repo. Most of the tests currently are configured to evaluate code under the `kbase-extension/static` directory. There is a sizable amount of code which lives under the `nbextensions` directory, but it does not seem to be configured to be tested at present. This comprises about 7% of the codebase by lines of code, so it might make sense to move all the code from here into the other folder. Before I would seriously suggest that, however, I would like a little more context on why there is a split in the first place and make sure that the move is possible and desirable.

The coverage report from karma also does not seem to properly count lines of code covered by the tests in this PR, so that should be addressed in the future.

The tests themselves also use slightly different syntax than the rest of the codebase, although this is internally consistent within this file. It would be worthwhile to adopt a convention to guide these sorts of syntax choices. In this case, the `await` syntax seems more maintainable to me than the equivalent `.then` approaches used elsewhere in the codebase. My main concern is that a convention be chosen and programmatically enforced through IDE tooling and perhaps a post commit hook, but I am flexible when it comes to the nature of that convention.

### DIFF
--- a/test/unit/karma.conf.js
+++ b/test/unit/karma.conf.js
@@ -10,7 +10,9 @@ module.exports = function (config) {
             jasmine: {
                 failFast: false,
                 DEFAULT_TIMEOUT_INTERVAL: 20000
-            }
+            },
+            requireJsShowNoTimestampsError: '^(?!.*(^/narrative/static/))',
+            clearContext: false
         },
         plugins: [
             'karma-jasmine',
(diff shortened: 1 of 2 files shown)


briehl:
This is great! I'm not sure either why the coverage of the nbextensions directory isn't showing up. Does it show up locally, and not get sent to coveralls?

PR #1694, opened 2020-06-05
WIP - Dependency updates - Epic omnibus edition

jsfillman:
Testing automated dependency updates via Dependabot merges.

Initial tests working on one environment with the following caveats:

- Python must be set to version `3.7`.
- I had to install `bower` manually, as it's now [deprecated](https://bower.io/blog/2017/how-to-migrate-away-from-bower/).

### To Do

- [ ] Test in a second fresh environment
- [ ] Look into migrating from bower to `Yarn`


briehl:
RE: `Look into migrating from bower to Yarn`
That's been on my list for a long while. There's some hangups with how dependencies get used and paths to them that's, I think, more of an issue of having time to work through some tedium. There's also a few dependencies that aren't available in NPM if I remember right. Though that might be different now.

It's also likely that those dependencies that are only available through bower can (a) be brought in with direct github links in yarn or (b) have some kind of more recent drop-in replacement.

Anyway, I'll be excited to work on that with you. Narrative dependency management really really (really) needs a refresh. Thanks for doing getting this started!

jsfillman:
Build is passing in CI. @briehl -- Go ahead and put this through its paces when you get a chance!

PR #1735, opened 2020-06-23
Working kbaseVariation widget

pranjan77:
- Added support for kbaseVariation widget
- Uses Jbrowseserver dynamic service


pranjan77:
need to fix things

PR #1744, opened 2020-07-01
restore hidden app cell info tab

briehl:


briehl:
Let's hold off on merging this until @scanon approves #1743 .

PR #1756, opened 2020-07-13
convert KB_WORKSPACE_ID to KB_WORKSPACE_NAME

briehl:
This is an environment variable that carries the name of the currently loaded narrative's workspace. It's been incorrectly named KB_WORKSPACE_ID for a long long time.


scanon:
Looks Okay To ME (LOTM)

briehl:
Noting here that if anyone uses the workspace name through code cells, that'll have to get changed. I'll also bring it up in Slack. I still this this is the right thing to do in the long run, and will remove ambiguity when we actually do support code cell usage.

PR #1757, opened 2020-07-13
update share panel - no more ws name allowed

briehl:
Previously, there was an input of `ws_name_or_id` in the sharing controls. It took either the workspace name or id of whatever workspace sharing permissions should be modified for.

Now it just takes workspace ids.


briehl:
While going through and writing tests for the sharing panel, found a couple things.
1. There's no tests right now for that widget.
2. Due to the flows in how it's written, it's really hard to write tests for. This'll need somewhat of a refactor. 
3. A fairly deep inspection of the code also showed the possibly for a number of untrapped errors, especially when trying to attach a narrative to an organization.

PR #1758, opened 2020-07-14
Fix SCT-2664: Show status when cells are collapsed

dakotablair:
This pull request makes an app cell status visible when the app cell is collapsed.

The unit tests test whether the status messages display in the toolbar menu depending on if the app cell is collapsed.


briehl:
Looks great. I don't have any comments. Let's merge and put it up on CI.

PR #1760, opened 2020-07-15
remove old side import tab (pre-staging area)

briehl:
This is a pretty ancient piece of code that's not used anymore, and not referenced anywhere. (see the code note `'This import interface is deprecated and will be replaced by the new interface (see next tab) in early 2018.'`)


ialarmedalien (approved):
Cruft removal always gets my approval!

ialarmedalien:
@briehl  should we revive this and get rid of some more cruft?

PR #1786, opened 2020-08-06
Staging fixes

briehl:
By default, Dropzone 5.1+, which we updated to in the last release, uses a 30000ms timeout for uploads.  
It woulda been cool if that default was included in their changelog.

This fixes the file upload widget to include some configurable values, including timeout (in ms), max parallel uploads (integer > 0), and max file size (in MB). Timeout's new, and the others were hard-coded.

About the timeout value - a look at the Dropzone code shows that the value is passed directly to an XMLHttpRequest.timeout value. A value of undefined, null, or 0 is the default for don't set a timeout and have it be unlimited. There's no guarantee that the Dropzone implementation will stay that way in the future, but for now, a value of 0 works. If we need to later, we can make it something ridiculous, like a year.

Todo: 
- [x] add mocked unit tests exercising the values.
- [ ] add integration tests against the service ensuring it fails as expected.
- [ ] possibly trap the timeout error that Dropzone throws in a different way (though if we're trying to set timeouts to be unlimited, that probably won't matter)


briehl:
Going to put the full integration test off to be a QA-driven test for now. This needs to get run through the wringer on CI.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeStagingDataTab.js
```diff
@@ -84,9 +84,9 @@ define([
                     userInfo: userInfo,
                     userId: Jupyter.narrative.userId
                 });
-                this.uploadWidget.dropzone.on('complete', function() {
-                    this.updateView();
-                }.bind(this));
+                // this.uploadWidget.dropzone.on('complete', function() {
```
bio-boris, line 87: Is this supposed to still be commented out?

bio-boris:
I think it's good that globus is now moved into a globus config variable, we can use this for the upcoming switch.
The Travis ci build failed .

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeStagingDataTab.js
briehl: Nope, that should be deleted. Thanks for catching.

briehl:
There's some issues with testing that I think are mostly worked through. Hopefully this works now on Travis.

briehl:
A little feature creep - this now also migrates the previously hard-coded Globus URL to a configuration variable, like @bio-boris mentioned above. And some other minor test harness changes, specifically to the kbaseCellToolbarMenu test module, as it could (sometimes) fail due to a missing mock.

PR #1806, opened 2020-08-28
Adjust how authentication cookies are set

briehl:

Auth tests (and local dev environment behavior) were failing due to auth tokens being unable to be set in the headless Chrome environment. Auth tokens are set as cookies in the browser, and there was a recent Chrome update that requires that `samesite=none` and `secure` are both set. This allows cross-origin cookies to work.

However, the only case where cookies need to be cross-origin is in a development environment that's configured to run on localhost (i.e. the narrative URL is something like http://localhost:8888). This is also running unsecure (not https), so a `secure` cookie cannot be set. So, when running a test that tries to make an auth token, that would always fail. But since we only need to have cookies like that on localhost, and they'll only communicate at the top level, we can use `samesite=Lax` and get all of the desired behavior. 

# Jira Ticket / Issue #

fixes https://kbase-jira.atlassian.net/browse/DATAUP-127

related to https://kbase-jira.atlassian.net/browse/DATAUP-116 (see also #1805 - tests are failing there because of this issue)

# Testing Instructions
Check the following:

Can also be tested by including either Firefox or FirefoxHeadless in the list of browsers that karma uses.

# Dev Checklist:

-  I have added tests that prove my fix is effective or that my feature works (N/A - the existing tests that were failing should be sufficient)
- (N/A) Any dependent changes have been merged and published in downstream modules

N/A

### DIFF
--- a/kbase-extension/static/kbase/js/api/auth.js
+++ b/kbase-extension/static/kbase/js/api/auth.js
@@ -56,7 +56,7 @@ define([
             const props = {
                 expires: TOKEN_AGE,        // gets translated to GMT string
                 path: '/',
-                samesite: 'none'
+                samesite: 'Lax'
             };
             if (Number.isInteger(cookie.expires)) {
                 props.expires = cookie.expires;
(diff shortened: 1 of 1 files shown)


briehl:
Also, now that this change is in, I searched the codebase, and the `setCookie` function is only invoked when setting a developer token cookie. The `setAuthToken` function isn't used anywhere, either, currently. As API calls, they're only there to mirror what the external kbase-ui flow does when working through the OAuth2 data flow.

PR #1897, opened 2020-10-26
DATAUP-278 bulk import UI ADR

briehl:

This PR contains an ADR for the new bulk import UI. Let's discuss!

# Jira Ticket / Issue #

### DIFF
--- a/docs/adrs/0003-bulk-import-ui.md
+++ b/docs/adrs/0003-bulk-import-ui.md
@@ -8,7 +8,7 @@ A primary goal of the current data upload refactoring project includes enabling 
 * Monitoring the state of those jobs and restarting if necessary.
 * Viewing the results of those jobs.
 
-Currently, this work is done by users using one cell at a time. An App Cell is generated for each desired data file upload, which can be time-consuming and cluttering for Narratives. The new design to support bulk operations.
+Currently, this work is done by users using one cell at a time. An App Cell is generated for each desired data file upload, which can be time-consuming and cluttering for Narratives. The new design supports bulk operations.
 
 ## Author
 
--- a/docs/adrs/0003-bulk-import-ui.md
+++ b/docs/adrs/0003-bulk-import-ui.md
@@ -40,6 +40,13 @@ The App Cell is, by far, the most complicated extension cell type. It has severa
 
 The reason for the App cell being an extension to code cells is the last point above: in essence, the primary UI components of the App cell serve as a fancy code generator. User interactions with form elements generate the code that can be used to start apps and view their results. Thus, anything that an app cell does with the UI can be done programmatically by an advanced user.
 
+> It should also be noted here that, regardless of the choice made, we may need to alter how app specs are made, interpreted, and rendered. The bulk import design has various inputs and outputs listed in horizontal rows while the existing app specs are rendered vertically. There's no notion of rows/columns in the app spec right now or how to visually display various inputs along a grid pattern. In order to effectively match the design we may need to do one of the following:
+> 1. Alter the app specs to support multiple in-line parameters while retaining backward compatibility (including documentation, etc.).
+> 2. Use a distinct app spec for importers that can include these options.
+> 3. Hard-code how all import apps look and feel in the Narrative interface itself.
+>
+> If we decide to update how app specs work, that means working in the `narrative_method_store` and `catalog` repos as well as internal documentation.
+
 ## Alternatives Considered
 
 * Implement the UI in the existing App Cell codebase
--- a/docs/adrs/0003-bulk-import-ui.md
+++ b/docs/adrs/0003-bulk-import-ui.md
@@ -1,0 +1,91 @@
+# Bulk Import UI
+
+Date: 2020-10-26
+
+A primary goal of the current data upload refactoring project includes enabling users to upload data in bulk and producing multiple new data objects with only a few user-side operations. That means introducing an interface that can handle the following steps:
+* Selecting several data files to concurrently import into new data objects.
+* Launching what would amount to several import jobs at once (i.e. a bulk import operation).
+* Monitoring the state of those jobs and restarting if necessary.
+* Viewing the results of those jobs.
+
+Currently, this work is done by users using one cell at a time. An App Cell is generated for each desired data file upload, which can be time-consuming and cluttering for Narratives. The new design to support bulk operations.
+
+## Author
+
+@briehl
+
+## Status
+
+In progress
+
+## Background
+
+App Cells (and other cells) are all built using the [Jupyter front end extension framework](https://jupyter-notebook.readthedocs.io/en/stable/extending/frontend_extensions.html). On application and page load, this framework allows the overriding and customization of most FE components. Specifically, in the Narrative, this framework is only used for Cells. 
+
+Jupyter Notebook (and thus Narrative) cells come in three types - Code, Markdown, and Raw. All KBase cell extensions operate on Jupyter Code cells. It's easiest to think of them as subclassing - an App Cell is a Code Cell specialized for running Narrative Apps. Each of these extensions is activated by creating a cell with a specific metadata component that describes how the cell should be treated. For example, adding the following block to code cell metadata tells the extension machinery to treat it as a KBase App Cell:
+```
+{
+    kbase: {
+        type: 'app'
+    }
+}
+```
+
+The App Cell is, by far, the most complicated extension cell type. It has 3 primary components that all serve different purposes:
+* A finite state machine (FSM) that's used to determine cell state.
+* A core system used to render the cell based on state.
+* A message bus that's used to communicate with the rest of the FE application, as well as the BE (kernel) to get job state.
+* An interpreter / manager for app specs - turns the app specs into sets of input form elements.
+* A code translator - turns the various sets of inputs into code that can be executed to run the App.
+
+The reason for the App cell being an extension to code cells is the last point above: in essence, the primary UI components of the App cell serve as a fancy code generator. User interactions with form elements generate the code that can be used to start apps and view their results. Thus, anything that an app cell does with the UI can be done programmatically by an advanced user.
+
+## Alternatives Considered
+
+* Implement the UI in the existing App Cell codebase
+* Implement the UI in a new Cell type
+* Don't use a cell, but some other interface
+
+## Decision Outcome
+
+## Consequences
+
+## Pros and Cons of the Alternatives
+
+### Implement the UI in the existing App Cell code base
+
+This option would entail adding a different mode for the current App Cell, likely based on some additional metadata feature.
+
+* `+` Reduce spread of additional, yet similar, cell types with similar code.
+* `+` Reuse existing cell control mechanisms currently in place with minimal modification.
+* `+` Reuse existing wrapping UI elements - cell title, icons, tabs, etc.
+* `-` Work done to the App cell needs to be made backward compatible with existing functionality.
+* `-` Any "breaking" work (i.e. refactor, etc.) will need to be complete before any releases are possible, making intermediate releases possibly challenging.
+* `-` Adding yet another mode for the App cell will increase the maintenance burden on an already complex code base - see especially the FSM.
+* `-` The draft UI is very complex and may turn the existing code into a deep series of nested logic.
+
+### Implement the UI in a new Cell type
+
+This option will create a new nbextension to serve a new cell type as an extension to the code cell.
+
+* `+` This would mean working with a fresh codebase without existing constraints.
+* `+` Intermediate, minimally functional changes, can be rapidly prototyped and released.
+* `-` To stick with the DRY principle, this would mean refactoring parts of the current app cell to make various components importable (message bus responses, job state management, tab menu, etc.) instead of their integration (really we might want to consider that anyway).
+* `-` Much of the structure of the mockup already functions similar to the existing App Cell and might become duplicated work with additional maintenance burden.
+* `-` This would mean updating other components that make use of Narratives - the Static Narrative interface, Search, and the Dashboard, for example, all introspect Narratives and handle cells based on their type.
+
+### Don't use a Cell, use some other interface
+
+The bulk importer UI work has been mocked up as a cell that becomes a part of the Narrative workflow, but that could change. This option would mean using some other interface rather than a cell to implement bulk import work. 
+
+This has been discussed before, especially in the context of how Narratives are intended to work as a repeatable record of analysis and discussion. Generally, importing files to data objects isn't very useful in trying to repeat work, especially if you don't have access to those files. This will also happen when those files expire on the server and are no longer available. In that case, including one or more import cells is misleading - those cells cannot be run again, implying that the analysis that the Narrative performs cannot be completely reproduced.
+
+* `+` Unimpeded by the constraints around the cell architecture as imposed by both the Jupyter Notebook and KBase codebases.
+* `+` Keeping all import jobs consolidated in one place, outside of cells, would reduce Narrative length and Narrative object size.
+* `+` Removing import cells into a separate interface would have a positive impact on load time.
+* `-` Entrypoint to the interface would be something entirely new.
+* `-` New work would need to be done to make use of the app state management systems and kernel communication.
+
+## References
+* JIRA ticket with the draft design: [DATAUP-268](https://kbase-jira.atlassian.net/browse/DATAUP-268)
+* [Jupyter notebook front end extensions](https://jupyter-notebook.readthedocs.io/en/stable/extending/frontend_extensions.html)
--- a/docs/adrs/0003-bulk-import-ui.md
+++ b/docs/adrs/0003-bulk-import-ui.md
@@ -31,7 +31,7 @@ Jupyter Notebook (and thus Narrative) cells come in three types - Code, Markdown
 }
 ```
 
-The App Cell is, by far, the most complicated extension cell type. It has 3 primary components that all serve different purposes:
+The App Cell is, by far, the most complicated extension cell type. It has several components that all serve different purposes:
 * A finite state machine (FSM) that's used to determine cell state.
 * A core system used to render the cell based on state.
 * A message bus that's used to communicate with the rest of the FE application, as well as the BE (kernel) to get job state.
--- a/docs/adrs/0003-bulk-import-ui.md
+++ b/docs/adrs/0003-bulk-import-ui.md
@@ -16,7 +16,7 @@ Currently, this work is done by users using one cell at a time. An App Cell is g
 
 ## Status
 
-In progress
+Accepted
 
 ## Background
 
@@ -55,7 +55,16 @@ The reason for the App cell being an extension to code cells is the last point a
 
 ## Decision Outcome
 
+The Bulk import UI will be developed in a new cell type, based on the existing app cell.
+
+Other points discussed:
+* We considered using a different interface than a cell, but decided that changing that entire paradigm right now would be too heavy of a burden on the design and product teams and would push back development efforts further. There was also no clear consensus on whether that's a desired result.
+* We discussed the implementation impacts of the importer app parameter layouts and decided to do something internal to the Narrative repo, as opposed to modifying the app specs in general. What the final implementation will be is left for later discussion.
+
+
 ## Consequences
+
+The team will need to develop a new cell type, which means building a new Jupyter notebook extension. We can mostly use the same startup code (from the `load_ipython_extension` function that's common among the other extensions). The team will also be spending time extracting components from the app cell that are embedded and making them more easily shareable with the new cell type.
 
 ## Pros and Cons of the Alternatives
 


docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Implement the UI in a new Cell type
+
+This option will create a new nbextension to serve a new cell type as an extension to the code cell.
+
+* `+` This would mean working with a fresh codebase without existing constraints.
+* `+` Intermediate, minimally functional changes, can be rapidly prototyped and released.
+* `-` To stick with the DRY principle, this would mean refactoring parts of the current app cell to make various components importable (message bus responses, job state management, tab menu, etc.) instead of their integration (really we might want to consider that anyway).
+* `-` Much of the structure of the mockup already functions similar to the existing App Cell and might become duplicated work with additional maintenance burden.
+* `-` This would mean updating other components that make use of Narratives - the Static Narrative interface, Search, and the Dashboard, for example, all introspect Narratives and handle cells based on their type.
```
eamahanna, line 75: Can we add a bit on how it may require changing importer app specs?

briehl: I added a block about possibly changing app specs and how they function.

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
```
bio-boris, line 82:
```
This would mean updating other components that make use of Narratives - the Static Narrative interface, Search, and the Dashboard, for example, all introspect Narratives and handle cells based on their type.
```
Wouldn't we have to do this anyway with both approaches?

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+
+This has been discussed before, especially in the context of how Narratives are intended to work as a repeatable record of analysis and discussion. Generally, importing files to data objects isn't very useful in trying to repeat work, especially if you don't have access to those files. This will also happen when those files expire on the server and are no longer available. In that case, including one or more import cells is misleading - those cells cannot be run again, implying that the analysis that the Narrative performs cannot be completely reproduced.
+
+* `+` Unimpeded by the constraints around the cell architecture as imposed by both the Jupyter Notebook and KBase codebases.
+* `+` Keeping all import jobs consolidated in one place, outside of cells, would reduce Narrative length and Narrative object size.
+* `+` Removing import cells into a separate interface would have a positive impact on load time.
+* `-` Entrypoint to the interface would be something entirely new.
+* `-` New work would need to be done to make use of the app state management systems and kernel communication.
```
bio-boris, line 94:
```
New work would need to be done to make use of the app state management systems and kernel communication.
```
I could see  this done completely using the frontend job clients only, ignoring the need for kernel communication

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+
+Date: 2020-10-26
+
+A primary goal of the current data upload refactoring project includes enabling users to upload data in bulk and producing multiple new data objects with only a few user-side operations. That means introducing an interface that can handle the following steps:
+* Selecting several data files to concurrently import into new data objects.
+* Launching what would amount to several import jobs at once (i.e. a bulk import operation).
+* Monitoring the state of those jobs and restarting if necessary.
+* Viewing the results of those jobs.
+
+Currently, this work is done by users using one cell at a time. An App Cell is generated for each desired data file upload, which can be time-consuming and cluttering for Narratives. The new design to support bulk operations.
```
MrCreosote, line 11: Nitpicky, but `The new design supports bulk operations`?

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+
+> It should also be noted here that, regardless of the choice made, we may need to alter how app specs are made, interpreted, and rendered. The bulk import design has various inputs and outputs listed in horizontal rows while the existing app specs are rendered vertically. There's no notion of rows/columns in the app spec right now or how to visually display various inputs along a grid pattern. In order to effectively match the design we may need to do one of the following:
+> 1. Alter the app specs to support multiple in-line parameters while retaining backward compatibility (including documentation, etc.).
+> 2. Use a distinct app spec for importers that can include these options.
+> 3. Hard-code how all import apps look and feel in the Narrative interface itself.
+>
+> If we decide to update how app specs work, that means working in the `narrative_method_store` and `catalog` repos as well as internal documentation.
```
MrCreosote, line 48: Does it really require catalog changes? I don't remember making catalog changes the last time I added fields to the spec, but that was a while ago.

It'd be great to make it easier to update NMS to add fields to the spec, but that's probably out of scope.

docs/adrs/0003-bulk-import-ui.md
MrCreosote, line 11: Nitpicky, but `The new design supports bulk operations`?

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+
+This option would entail adding a different mode for the current App Cell, likely based on some additional metadata feature.
+
+* `+` Reduce spread of additional, yet similar, cell types with similar code.
+* `+` Reuse existing cell control mechanisms currently in place with minimal modification.
+* `+` Reuse existing wrapping UI elements - cell title, icons, tabs, etc.
+* `-` Work done to the App cell needs to be made backward compatible with existing functionality.
+* `-` Any "breaking" work (i.e. refactor, etc.) will need to be complete before any releases are possible, making intermediate releases possibly challenging.
+* `-` Adding yet another mode for the App cell will increase the maintenance burden on an already complex code base - see especially the FSM.
+* `-` The draft UI is very complex and may turn the existing code into a deep series of nested logic.
```
MrCreosote, line 72: * `-` Multiple imports result in multiple import cells scattered over the narrative rather than a unified location for import operations and history

docs/adrs/0003-bulk-import-ui.md
MrCreosote, line 94: Is this just factoring out the code for reuse or is it something more?

docs/adrs/0003-bulk-import-ui.md
MrCreosote: This is true for the next option as well

MrCreosote, line 72: * `-` Import app cells support re-runability, but with the 90 day file deletion of the staging area and narrative copies, rerunning will eventually fail (true for next option as well)

MrCreosote:
Here's a crazy thought: Do we really need to support a spec based method of specifying file importers? Could we just hard to the import UI rather than generating it from specs?

The answer to this question, I guess, depends on a few other questions: 1) how many types do we import, 2) how often do we add types, and d) how much labor does it save to specify the importer in a spec vs. code.

bio-boris:
> Here's a crazy thought: Do we really need to support a spec based method of specifying file importers? Could we just hard code the import UI rather than generating it from specs?
> 
> The answer to this question, I guess, depends on a few other questions: 1) how many types do we import, 2) how often do we add types, and d) how much labor does it save to specify the importer in a spec vs. code.



> Here's a crazy thought: Do we really need to support a spec based method of specifying file importers? Could we just hard code the import UI rather than generating it from specs?
> 
> The answer to this question, I guess, depends on a few other questions: 1) how many types do we import, 2) how often do we add types, and d) how much labor does it save to specify the importer in a spec vs. code.

+1

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
+# Bulk Import UI
+
+Date: 2020-10-26
+
+A primary goal of the current data upload refactoring project includes enabling users to upload data in bulk and producing multiple new data objects with only a few user-side operations. That means introducing an interface that can handle the following steps:
```
leia-sefkin, line 5: A link to the figma designs, or else some screenshots would be helpful here for context for readers less familiar with the project

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+* A message bus that's used to communicate with the rest of the FE application, as well as the BE (kernel) to get job state.
+* An interpreter / manager for app specs - turns the app specs into sets of input form elements.
+* A code translator - turns the various sets of inputs into code that can be executed to run the App.
+
+The reason for the App cell being an extension to code cells is the last point above: in essence, the primary UI components of the App cell serve as a fancy code generator. User interactions with form elements generate the code that can be used to start apps and view their results. Thus, anything that an app cell does with the UI can be done programmatically by an advanced user.
+
+> It should also be noted here that, regardless of the choice made, we may need to alter how app specs are made, interpreted, and rendered. The bulk import design has various inputs and outputs listed in horizontal rows while the existing app specs are rendered vertically. There's no notion of rows/columns in the app spec right now or how to visually display various inputs along a grid pattern. In order to effectively match the design we may need to do one of the following:
```
leia-sefkin, line 43: A bit tangential to the ADR specifically but if there's any suggestions we have to make the design less onerous on whatever development path we choose we should consider and discuss with design.

docs/adrs/0003-bulk-import-ui.md
leia-sefkin: > - Multiple imports result in multiple import cells scattered over the narrative rather than a unified location for import operations and history 

Might be a silly question but is this a given? If we are talking about batch analysis design at the same time could we work on those cells as well so that they are more unified along the same lines?

leia-sefkin: Oof good point - I wonder if design has considered this as well

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+
+This has been discussed before, especially in the context of how Narratives are intended to work as a repeatable record of analysis and discussion. Generally, importing files to data objects isn't very useful in trying to repeat work, especially if you don't have access to those files. This will also happen when those files expire on the server and are no longer available. In that case, including one or more import cells is misleading - those cells cannot be run again, implying that the analysis that the Narrative performs cannot be completely reproduced.
+
+* `+` Unimpeded by the constraints around the cell architecture as imposed by both the Jupyter Notebook and KBase codebases.
+* `+` Keeping all import jobs consolidated in one place, outside of cells, would reduce Narrative length and Narrative object size.
+* `+` Removing import cells into a separate interface would have a positive impact on load time.
+* `-` Entrypoint to the interface would be something entirely new.
```
leia-sefkin, line 93: Because of this point if we're feeling strongly about this option we should definitely discuss with design what the implications are

docs/adrs/0003-bulk-import-ui.md
leia-sefkin: Oh nm!  I finally saw the resources link at the bottom 🤦

docs/adrs/0003-bulk-import-ui.md
MrCreosote: TBF Bill made this point later on in the doc, I just reiterated it here for the app cons

leia-sefkin:
Really nice write up! Couple questions:

* In a perfect world which option would you prefer?
* Is there anyone else at KBase outside of this proj team who should help review?

docs/adrs/0003-bulk-import-ui.md
MrCreosote: That's my understanding based on the design - each time a user triggers an import event from the staging area, a new import cell gets created, right? And presumably the cell gets created at the user's current position in the narrative which could be anywhere.

leia-sefkin: From what I've seen in the designs that's not the case, unless I'm completely misinterpreting: https://www.figma.com/proto/aPigPBY0X8iUatKKe03Eo0/concept-sketches?node-id=471%3A1398&scaling=scale-down

MrCreosote: That link goes to a bulk analysis screen - are you sure that's right?

So when a user does a 2nd import from the staging service to a narrative, do the files get added to the same importer cell as before then? Because that would mean there needs to be a way to select only the new files in the importer cell so you don't import the files from the 1st import again

leia-sefkin: ok - sorry i was reading your comment but also not really reading it or digesting it 😬  sorry. i understand where you are coming from now. i think that is definitely a valid concern we should address with design

docs/adrs/0003-bulk-import-ui.md
briehl: Probably. But the existing approach just reports as an app cell - a new approach might break things or require an update to search to get the new cells indexed properly.

docs/adrs/0003-bulk-import-ui.md
briehl: > Does it really require catalog changes? I don't remember making catalog changes the last time I added fields to the spec, but that was a while ago.

For changing spec fields, I don't think so. Hm, actually Catalog might just call out to NMS for its spec parsing needs entirely. But I think these spec changes could be larger than just adding a new field and hoping it works. I might be wrong, though. This all feels like quite a bit of scope creep to me.

> It'd be great to make it easier to update NMS to add fields to the spec, but that's probably out of scope.

Yup.

docs/adrs/0003-bulk-import-ui.md
briehl: fixed

docs/adrs/0003-bulk-import-ui.md
briehl: Partly factoring for reuse. Partly just hedging against other work that might be involved to better support batch / bulk job operations. Regardless, the second will have to be done. I'm not sure how much extra effort will be involved in factoring out the cell state management code.

briehl: I could, too, but it makes me nervous to have some cells maintained only on the front end, and some cells maintained by the kernel. Kernel-side operations also make it easier to control data flow between the front end and the EE2 service. State lookups are only a single RPC that happens in a loop, and I think it would be best to keep it that way as much as possible.

briehl:
> Here's a crazy thought: Do we really need to support a spec based method of specifying file importers? Could we just hard code the import UI rather than generating it from specs?

I feel like I put that somewhere as an option. But yes, that's a good question. Hard coding the import UI for various types would give a lot more freedom. It would also increase the work burden for creating new types and uploaders. But maybe that's ok for now?

docs/adrs/0003-bulk-import-ui.md
briehl: Completely agreed. From the mockups it feels like cell usage was a design decision that was just assumed. But it doesn't have to be. We can ask the design team if they've considered alternatives.

docs/adrs/0003-bulk-import-ui.md
MrCreosote: > Catalog might just call out to NMS for its spec parsing needs entirely

MrCreosote: > Catalog might just call out to NMS for its spec parsing needs entirely

MrCreosote: > Catalog might just call out to NMS for its spec parsing needs entirely
I think that's right.

>  This all feels like quite a bit of scope creep to me.

MrCreosote: > Catalog might just call out to NMS for its spec parsing needs entirely

I think that's right.

> This all feels like quite a bit of scope creep to me.

We're going to have to specify the layout *somewhere*, the only question here is where. AFAICT the choices are:

* in the current spec
* in some other new spec
* in the code

briehl:
> * In a perfect world which option would you prefer?

Good question! If we go with building a specialized cell interface that'll handle this, my perfect world would include a refactor of the existing app cell to deal with cases where we want customized interfaces. It would have a hierarchy of "app cells" of various types that all inherit from some parent app cell extension. That would include our bulk execution app cell type as well.

I'm not convinced we can do that in our expected timeframe, though. So I think I'd suggest doing what refactoring we can to remove what would be some common state management and communications modules, and making another cell extension.

My other perfect world would not include data upload in the Narrative working area as a cell at all. One of the main goals of the design of Narratives is reproducibility and sharing - if you can see my Narrative, you should be able to see and do everything I do **within the system**. Being able to upload the same data sits on the edge of that system, at least because you probably don't have the same data files in your staging area that I do, and those expire and are removed after time anyway. 

There needs to be a record of how data enters the system, and specifically into a Narrative, whether it's uploaded or copied. I just don't think a cell is necessarily the best place for that.

> * Is there anyone else at KBase outside of this proj team who should help review?

Probably not. I think the folks on this project team have enough insight at this point. Maybe @eapearson if he wants to chime in (don't feel pressured if you don't have anything to add, Erik!), as he did most of the original development of the app cell.

MrCreosote:
>> Here's a crazy thought: Do we really need to support a spec based method of specifying file importers? Could we just hard code the import UI rather than generating it from specs?

> I feel like I put that somewhere as an option. But yes, that's a good question. Hard coding the import UI for various types would give a lot more freedom. It would also increase the work burden for creating new types and uploaders. But maybe that's ok for now?

How much would it increase the work burden, especially if the designs call for things the DSL doesn't support?

We could pretty easily see 1) how often we add types (my guess: hardly ever) and 2) who adds those types (my guess: only KBase devs) by looking at the upload spec repo, wherever that lives.

With an appropriate uploader framework & libraries I'm guessing it could be basically just as easy to spec a basic uploader in code vs. a DSL and have the flexibility of code for cases where a DSL becomes too unwieldy because you constantly have to be adding layout / behavior options.

MrCreosote:
> I feel like I put that somewhere as an option.

You did: https://github.com/kbase/narrative/pull/1897/files#diff-386e5ce5a26ff0b71b41603e353ee4db0cfb6aa6359ada06ddc50ab473afb88bR46

Either I glossed over it or I accidentally and not on purpose stole it and presented it as my idea

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -0,0 +1,98 @@
...
+* `-` The draft UI is very complex and may turn the existing code into a deep series of nested logic.
+
+### Implement the UI in a new Cell type
+
+This option will create a new nbextension to serve a new cell type as an extension to the code cell.
+
+* `+` This would mean working with a fresh codebase without existing constraints.
+* `+` Intermediate, minimally functional changes, can be rapidly prototyped and released.
+* `-` To stick with the DRY principle, this would mean refactoring parts of the current app cell to make various components importable (message bus responses, job state management, tab menu, etc.) instead of their integration (really we might want to consider that anyway).
+* `-` Much of the structure of the mockup already functions similar to the existing App Cell and might become duplicated work with additional maintenance burden.
```
ialarmedalien, line 81: these two somewhat cancel each other out -- presumably if we were to create a new cell type, shared functionality would be abstracted out in some way so that we didn't end up with a copy-pasta codebase

docs/adrs/0003-bulk-import-ui.md
ialarmedalien: This has come up in previous design discussions but I am not sure that eliminating cell usage has been seriously considered (changing the present paradigm? what?!). I would like to see a more comprehensive examination of pros and cons of taking the import cells out of the narrative. As a user, it's annoying to look at other narratives and have to scroll through a load of cells that represent loading local files.

docs/adrs/0003-bulk-import-ui.md
briehl: From the impromptu meeting we held 10/28 9:30am PT, we decided to go with an "internal" spec. This will be a structured document that the importer cell could read that would set the layout. This avoids having to do major updates to Narrative Method Store, and avoids hard coding layouts that might be evolving rapidly over the next few months.

docs/adrs/0003-bulk-import-ui.md
leia-sefkin: Discussed with team and PMs - since this is current design pattern we are going to stick with that for now rather than rework the whole paradigm.

docs/adrs/0003-bulk-import-ui.md
leia-sefkin: As above - we discussed with PMs and decided that we don't want to rework the entire paradigm at this time.

docs/adrs/0003-bulk-import-ui.md
ialarmedalien: (this is just a minor observation, no action needed)

docs/adrs/0003-bulk-import-ui.md
briehl: Since we're sticking with cells as they are with their code submission to start jobs, we can stick with the existing management architecture.

briehl: (from meeting 10/28 9:30am PT)

briehl:
We met this morning as a group with our DMs and PMs and decided to go with making a new version of the App Cell.

briehl:
I will write up the changes we agreed on and the rest of the doc soon.

docs/adrs/0003-bulk-import-ui.md
MrCreosote: Did we make this decision? I remember talking about it but it wasn't clear to me that we actually made a decision - or rather that the engineering team made a decision.

docs/adrs/0003-bulk-import-ui.md
```diff
@@ -55,8 +55,17 @@ The reason for the App cell being an extension to code cells is the last point a
 
 ## Decision Outcome
 
+The Bulk import UI will be developed in a new cell type, based on the existing app cell.
+
+Other points discussed:
+* We considered using a different interface than a cell, but decided that changing that entire paradigm right now would be too heavy of a burden on the design and product teams and would push back development efforts further. There was also no clear consensus on whether that's a desired result.
```
MrCreosote, line 61: > but decided that changing that entire paradigm right now would be too heavy of a burden on the design and product teams and would push back development efforts further

I'm still not sure I believe this, but I'm in the minority apparently so *shrug*.

eamahanna (approved):
LGTM :)

PR #1920, opened 2020-11-03
bump version to 4.3.1, add release notes

briehl:
Just updates the release notes to encompass all of the user-facing changes, and a few of the more major non-user-facing ones.

There's no real new features here (unless I'm forgetting something?), so this seems like a patch version change to me.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,24 @@ The Narrative Interface allows users to craft KBase Narratives using a combinati
 
 This is built on the Jupyter Notebook v6.0.2 (more notes will follow).
 
+### Version 4.3.1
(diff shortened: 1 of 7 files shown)


briehl:
This has been the typical routine for adding release notes, focusing on user-facing changes. But I'm up for a change to that if it makes sense to include every modification.

bio-boris (approved):

ialarmedalien (commented):
Should this go into the `truss` branch so it gets updated when `truss` is merged into `develop`?

briehl:
@ialarmedalien once this gets merged, I'll merge `develop` -> `truss` branch to get it caught up. I think `truss` has some more in-flight things, and I think now's a good time for a prod release.

PR #1960, opened 2020-11-19
DATAUP-288 left column file type panel

briehl:

This includes a component that's, generically, a toggleable category panel. It's used here as the file type selector.

Right now the component itself works as described in the ticket, but the bulk import cell doesn't do too much toggling based on the selection. It's also hard-coded to have FASTQ and SRA file types only, right now (that's the next step to do).

Another issue I'm having is that we're on FontAwesome 4.7.0, which doesn't have the DNA icon that goes up in the header. I think upgrading to the free version of 5 is a great idea, but might also be kinda onerous as it's not backward compatible. I'm not sure what to do about that yet, but I'll ask design. I just wanted to get this PR set up first.

It should look roughly like this (ignore the rest of the incomplete layout...)
Screen Shot 2020-11-19 at 12 03 56 PM

# Jira Ticket / Issue #

# Testing Instructions
open a bulk import cell, and it should appear
- [ ] Tests pass in travis and locally 

# Dev Checklist:


nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -459,7 +466,7 @@ define([
         function renderLayout() {
             const events = Events.make(),
                 content = div({
-                    class: `${cssCellType}__layout_container kbase-extension kb-app-cell`,
```
briehl, line 462: Having the `kb-app-cell` class in there was causing some spacing problems and introducing a blue border that's not inline with the designs. The rest of the required styling is in the `kb-bulk-import__layout_container` class, so that seems superfluous anyway.

There's a couple of minor things to be tweaked to the widget container, I think, so I'll do that in another PR.

eamahanna: Can you record these todo items in the cleanup ticket in the backlog, or somewhere else?

nbextensions/bulkImportCell/categoryPanel.js
```diff
@@ -12,58 +12,162 @@ define([
...
         function renderLayout() {
             const events = Events.make(),
-                content = div({}, [
-                    renderHeader(),
-                    renderCategories()
-                ]);
+                content = [renderHeader()].concat(renderCategories(events)).join('');
             return {
                 content: content,
                 events: events
```
eamahanna, line 77: Why do we need to return events?

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -25,7 +25,7 @@ define([
     CellTabs,
     CellControlPanel,
     ConfigureWidget,
-    CategoryPanel
+    FileTypePanel
```
eamahanna, line 28: If we are calling this panel `FileTypePanel`, can you also rename the `CategoryPanel` files to reflect this? i.e. `categoryPanel.js` -> `fileTypePanel.js`

briehl: My thought was to re-use that category panel in the bulk analysis cell. Though looking at the designs, now, maybe we'll have to use something different anyway. So now I'm not sure if that needs to just be a custom, standalone component or not.

briehl: If not, I can just rename it. If so, I should really move it to the `common/cellComponents` dir.

nbextensions/bulkImportCell/categoryPanel.js
briehl: Those get bound to the layout as a whole. There's some event binding that happens behind the scenes once the whole DOM structure gets enabled.

nbextensions/bulkImportCell/bulkImportCell.js
leia-sefkin: for now rename and move it later if it makes sense? could be one of those technical health improvement items we add to the backlog

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -323,12 +323,14 @@ define([
         }
 
         /**
-         * This toggles which category (file type) should be shown
-         * // TODO: rename this and other widgets & elements to fileType
-         * @param {string} category - the category that should be shown
+         * This toggles which file type should be shown. This sets the
```
leia-sefkin, line 326: minor semantic point but isn't it more the import app or import data type vs. file type? or have i been spending too much time in the parameter/file input section

leia-sefkin (approved):
looks good overall. agree with @eamahanna that it makes sense to rename things now for consistency, but otherwise think it's safe to merge

nbextensions/bulkImportCell/bulkImportCell.js
briehl: Done, added here: https://kbase-jira.atlassian.net/browse/DATAUP-317

nbextensions/bulkImportCell/bulkImportCell.js
briehl: Well, now I'm confused. The design screenshots differentiate between FASTQ reads and SRA reads. Those are two different file types, but they both get converted into reads in the end. My feeling was that the categories in that column mapped to the different file types selected from the staging area.

nbextensions/bulkImportCell/bulkImportCell.js
briehl: Ok, renamed. We can move it around later if that's the best answer.

briehl: I'll add that to the list in DATAUP-317 to consider later

briehl:
Codacy's complaining about the module's return like:
```javascript
return {
    start: start,
    stop: stop
}
```
being an "unnecessary block". It thinks I'm treating it as a code block, not a returning object. Not sure what to do here besides ignore it.

PR #1978, opened 2020-11-25
DATAUP-322: ESLint / Prettier: npm machinery

ialarmedalien:
This PR puts in place machinery and configs for prettier and eslint. This includes:
- prettier config
- updating eslint config so it does not clash with prettier
- adding prettier to the dev dependencies and removing `pre-commit`, which had somehow found its way back in
- updating the package-lock.json with the new versions
- adding a `.gitattributes` file to ensure that line endings are always unix style.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "chromedriver": "^86.0.0",
     "cssnano": "^4.1.10",
     "eslint": "^7.13.0",
+    "eslint-config-prettier": "^6.15.0",
     "grunt": "1.3.0",
     "grunt-cli": "1.3.2",
     "grunt-contrib-concat": "^1.0.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "chromedriver": "^86.0.0",
     "cssnano": "^4.1.10",
     "eslint": "^7.13.0",
+    "eslint-config-prettier": "^6.15.0",
```
ialarmedalien, line 20: https://github.com/prettier/eslint-config-prettier - turns off rules that would conflict between eslint and prettier

briehl:
> * adding prettier to the dev dependencies and removing `pre-commit`, which had somehow found its way back in

🎉

briehl (approved):
Looks great! Failing tests aren't the fault of these changes, but ongoing drama from elsewhere.

PR #1992, opened 2020-12-02
Merge develop -> truss

briehl:
The purpose here is to get truss caught up with develop again.

There haven't been many changes, just to dependencies, so this shouldn't be too painful.


ialarmedalien:
@briehl Do you want to resolve the dependencies so this can be approved?

PR #2037, opened 2020-12-15
DATAUP-310 state management: adding app specs and toggling between them

briehl:
Let's try this again.

This has a few updates in support of bulk import cell state management. Much of this is setup for handling the state that the cell should be in, and much of that is based on what the user's inputs are, which in turn depends on what the apps are that need to be run. So this does a few things to set that up.

**1. Adds fetching of app specs at cell startup.** 
Now, when starting a bulk import cell, this fetches the app specs from the Narrative Method Store service. Fetching these means changing around the cell setup and initialization code. It also shifts around the inputs to the cell to look like this:
```
{
    importData: {
        fileType: {
            files: ['file1'],
            appId: 'SomeModule/someapp'
        },
        specs: {
            'SomeModule/someapp': { complicated app spec }
        }
    }
}
```
I added this structure, instead of just adding the spec into the fileType, as multiple file types can use the same spec. It seemed cleaner. They get stored in the cell metadata this way, too.

This kinda opened a rabbit hole that we need to go down anyway, which is adapting the `model` object to be useful and directly represent what we're storing in the metadata. Other widgets were modified to handle that.

**2. Adds a first step at the state management.**
This will, in a subsequent PR, function as the FSM does for the app cell. Right now there's just a single state - "new" - that is set for configuration.

**3. Implements the file type toggle.**
Now that there's different specs available for different apps, changing the active file type should also change the active configuration tab to reflect the app spec. This means having the configuration tab (and filePathWidget and paramsWidget) aware of the currently active spec.

# Jira Ticket / Issue #

# Testing Instructions
You'll need to create a new Bulk Import cell locally for this to operate right. You'll probably see some strange errors otherwise. Just make a new cell with the usual file selection methods.

# Dev Checklist:


kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
```diff
@@ -52,7 +52,10 @@ define([
         const model = config.model;
 
         function start(arg) {
-            const appSpec = model.getItem('app.spec');
+            let appSpec = model.getItem('app.spec');    // for plain app cell
```
briehl, line 55: I'm not crazy about how this forks. It's due to how normal app cells vs. the bulk import cell has information.

I also came to realize that the app metric info that gets represented here is stored in the cell at startup time, which means the latest app usage info isn't shown here. IMO, this tab should just take an app id and fetch all of its information on load. This does mean a couple of RPCs, but they should be fairly light weight.

kbase-extension/static/kbase/js/kbaseNarrative.js
```diff
@@ -1250,17 +1252,39 @@ define([
...
+     *     files: array of files
+     *   }
+     * }
+     * This returns a Promise that resolves into the cell that was created.
+     * @param {object} bulkInput keys = type ids, values = an object with properties
+     *  - appId - the app id to use for that file type (to be used in fetching the spec)
+     *  - files - array of files to import with that file type
      */
-    Narrative.prototype.insertBulkImportCell = function (typesToFiles) {
+    Narrative.prototype.insertBulkImportCell = function (bulkInput) {
```
briehl, line 1268: The big change here is that this now returns a Promise that resolves into the new cell. That Promise includes fetching the app specs, and additional info that the info tab currently needs.

kbase-extension/static/kbase/js/util/kbaseApiUtil.js
```diff
@@ -27,21 +25,41 @@ define([
...
+     * @param {Array} idList - a list of app ids
+     * @param {String} tag - a tag, one of 'release', 'beta', 'dev'
      */
     function getAppSpecs (idList, tag) {
         if (!tag) {
             tag = getAppVersionTag();
         }
         var nms = new NarrativeMethodStore(Config.url('narrative_method_store'));
-        return Promise.resolve(nms.get_method_spec({ids: idList, tag: tag}));
+        return Promise.all([
```
briehl, line 45: Updated this to also include the `get_method_full_info` call. This is used to get info that the app cell uses, and it seemed appropriate to put it here as well.

briehl: Some of the info is redundant, though, even in the existing app cell. That should get worked out in another PR.

kbase-extension/static/kbase/js/widgets/narrative_core/upload/stagingAreaViewer.js
```diff
@@ -50,7 +50,7 @@ define([
 
         init: function (options) {
             this._super(options);
-            this.bulkImportTypes = ['fastq_reads'];
+            this.bulkImportTypes = ['fastq_reads', 'sra_reads'];
```
briehl, line 53: Now that we're detaching some of the dummy data, we need to allow `sra_reads` as well. We can also add others as we go.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -189,34 +223,58 @@ define([
...
-                onUpdate: function (props) {
-                    Utils.setMeta(this.cell, 'bulkImportCell', props.getRawObject());
-                },
-            });
+            fileTypePanel;
         if (options.initialize) {
-            initialize(typesToFiles);
+            initialize(options.specs);
         }
+        let model = Props.make({
```
briehl, line 228: Now, instead of running `Props.make` with the test data, this can use the actual data from the cell.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -189,34 +223,58 @@ define([
...
+            data: Utils.getMeta(cell, 'bulkImportCell'),
+            onUpdate: function(props) {
+                Utils.setMeta(cell, 'bulkImportCell', props.getRawObject());
+            }
+        });
 
-        let spec = Spec.make({
-            appSpec: model.getItem('app.spec'),
+        let specs = {};
```
briehl, line 235: There's a little bit of strangeness here. The "specs" that are used by the app cell (and widgets in the BI cell) are somewhat modified from the raw document from the NMS service. Without overwhelming with detail, they get tweaked to know what default values are acceptable, and how they get assorted into "inputs" vs "parameters" vs "outputs".

To keep consistent with the app cell, the "raw" specs are kept around. There's a TODO in the app cell code to modify that, but I wanted to err on the side of consistency for now.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -189,34 +223,58 @@ define([
...
+             *   fileType1: {
+             *      param1: '' or default from spec,
+             *      param2: '' or default from spec,
+             *      ...etc.
+             *   },
+             *   fileType2: {...}
+             * }
+             *
+             */
+            const initialParams = {};
```
briehl, line 267: This works for now, but will need some updating.
Right now, it treats each app as an individual run. When, really, the "inputs" should be done as lists, as represented in the filePathWidget. But each app has those input parameters as singletons. I'm not sure the right way to do this yet.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -377,42 +440,82 @@ define([
...
          */
-        function toggleTab(tab) {
+        function toggleTab(tab, fileType) {
+            // if we're toggling the currently selected tab off,
+            // then it should be turned off.
+            if (tab === state.tab.selected && tab !== null && !fileType) {
+                tab = null;
+            }
             state.tab.selected = tab;
+            stopWidget();
```
briehl, line 450: This is another feature from the app cell that should be emulated for consistency. If the currently selected tab is clicked again, it gets shut off - it just removes that tab and leaves the cell closed.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -377,42 +440,82 @@ define([
...
+
+            if (tab !== null) {
+                if (!fileType) {
+                    fileType = state.fileType.selected;
+                }
+                runTab(tab, fileType);
```
briehl, line 456: Each tab needs to know what fileType is selected to know what app id to use, what specified inputs to use, what jobs to show, etc.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -377,42 +440,82 @@ define([
...
+            }
+            cellTabs.setState(state.tab);
+        }
+
+        function stopWidget() {
```
briehl, line 461: Just decomposed this into its own function. This stops the currently selected widget, if there is one.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -377,42 +440,82 @@ define([
...
-                workspaceInfo: workspaceInfo,
-                cell: cell,
-                model: model,
-                spec: spec,
-                jobId: undefined,
+                cell,
+                model,
+                spec: specs[typesToFiles[state.fileType.selected].appId],
+                fileType,
+                jobId: undefined
```
briehl, line 485: This update adds the correct spec and currently selected file type to the tab's `make` function.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -435,48 +538,19 @@ define([
...
+            // load current state from state list
+            // modify to handle file types panel
+            let state = States[0].ui;
+            let fileTypeState = {
+                completed: {}
             };
+            for (const fileType of Object.keys(typesToFiles)) {
+                fileTypeState.completed[fileType] = false;
+            }
+            fileTypeState.selected = Object.keys(typesToFiles)[0];
```
briehl, line 548: This really just selects the first, and only, state available. Next update will serialize this and load it from the cell metadata.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -521,21 +595,25 @@ define([
...
+            let fileTypeMapping = {};
+            let uploaders = Config.get('uploaders');
+            for (const uploader of uploaders.dropdown_order) {
+                fileTypeMapping[uploader.id] = uploader.name;
+            }
+            for (const fileType of Object.keys(typesToFiles)) {
+                fileTypesDisplay[fileType] = {
+                    label: fileTypeMapping[fileType] || `Unknown type "${fileType}"`
+                };
+            }
```
briehl, line 606: Here, we use the same configuration as the staging area to map from file type ids like `'fastq_reads'` to the label string `'FASTQ Reads'`. That's not exactly what's in the Figma mockups, so we might want to discuss if that's a problem.

nbextensions/bulkImportCell/main.js
```diff
@@ -18,19 +18,14 @@ define([
     'bluebird',
     'common/runtime',
     'common/error',
-    './bulkImportCell',
-    'kb_service/client/workspace',
-    'kb_service/utils',
-    'custom/custom'
```
briehl, line 24: Don't need any of these - they should've been deleted along with the `get_workspace_info` call

nbextensions/bulkImportCell/tabs/configure.js
```diff
@@ -62,22 +63,26 @@ define([
...
 
-            This is more or less copied over from the way that the appCell handles building the widget. This model assumes one instance of the params widget per cell, so may need some adjustment as we make the bulk import work with multiple data types
+            This is more or less copied over from the way that the appCell handles building the widget.
+            This model assumes one instance of the params widget per cell, so may need some adjustment as
+            we make the bulk import work with multiple data types
         */
         function buildParamsWidget(node) {
             const paramBus = runtime.bus().makeChannelBus({ description: 'Parent comm bus for input widget' });
+            // This is the key in the model that maps to the list of params for the current app.
+            const paramKey = `params.${fileType}`;
```
briehl, line 73: I'm not sure if there's a better way to do this. Right now in the data model, params are keyed under each filetype. The fileType is the main key for all the other data - it maps to the app spec ids, the files, and to which params are active.

nbextensions/bulkImportCell/bulkImportCell.js
leia-sefkin: don't have a good answer as this seems like a tough problem. is it worth representing here in a comment as a todo and/or a Jira ticket? 

how does batch mode handle it? is the only way to have a single app run process multiple files with a set of params for those files to update the app itself?

nbextensions/bulkImportCell/bulkImportCell.js
leia-sefkin: IIRC the design mockups were geared towards creating more separation between the FASTQ reads types, so that might be why the title is slightly different. either way agree that it's worth discussing to make sure we're on the right track

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -377,42 +440,79 @@ define([
          * 1. if there's a tab showing, stop() it and detach it
          * 2. update the tabs state to be selected
          * @param {string} tab id of the tab to display
+         * @param {string} fileType id of the filetype we're swapping to
          */
-        function toggleTab(tab) {
+        function toggleTab(tab, fileType) {
```
leia-sefkin, line 445: not sure what happened here, but on initial load of the cell i'm now getting no tab selected (rather than default of configure).

nbextensions/bulkImportCell/tabs/configure.js
leia-sefkin: i think this approach makes sense, especially as we agreed that for now we will only have one set of params per file type

leia-sefkin (approved):
this is a massive amount of work, thanks for all of the comments and breadcrumbs to help unpack what's going on here. i have a couple questions, but i don't see any blockers to getting this merged.

PR #2052, opened 2020-12-22
DATAUP-356 output objects

briehl:

Add the output objects widget viewer.

This expects to be given a list of reports UPAs / references (anything that the workspace `get_objects2` call will recognize as a ref), a workspace client, and a node to render itself in. It then makes 2 workspace calls - one to filter down the report data to get its `objects_created` field, and one to get the object info (name and type, specifically) for each of those created objects.

I considered having this get fed its data to render, but I think that might make the logic in the reports tab more complicated. The reports tab can now just be a container to make sure its widgets get put in the right place and activated at the right time.

Right now, this might look a little different from the reports widget, in that it has a caret for shrinking while the reports widget does not. That'll get updated in the reports view, so they'll both look like the standard app cell does.

# Jira Ticket / Issue #

# Testing Instructions
The Results tab is still dependent on the test data, so for now you can just open a cell, toggle over to the tab, and see it work!

# Dev Checklist:


nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -433,6 +436,12 @@ define([
             });
         }
 
+        function getWorkspaceClient() {
+            const worskpaceURL = runtime.config('services.workspace.url');
```
ialarmedalien, line 440: typo

kbase-extension/static/kbase/js/common/cellComponents/tabs/results/outputWidget.js
```diff
@@ -1,65 +1,165 @@
...
             // send parent the ready message
-            doAttach(arg.node);
-
-            return renderOutput(arg.data).catch(function (err) {
-                // do somethig with the error.
-                console.error('ERROR in start', err);
-            });
+            return doAttach(arg)
+                .catch((err) => {
+                    // do somethig with the error.
```
ialarmedalien, line 160: that `somethig` typo is proliferating!

ialarmedalien (approved):
A couple of little typos to fix, otherwise looks great!

PR #2058, opened 2021-01-04
Update staging_upload.json

bio-boris:

Resolving  feedback on 331
* Added missing importers to the list
* Changed the correct importer that pops up
* Note, the SRA reads will no longer work because it's not deployed
* However the FASTQ will still work, because that one is hardcoded to open up the new importer cell


briehl (approved):
👍

PR #2060, opened 2021-01-05
DEVOPS-431 Change ee2 endpoint to new endpoint

bio-boris:


briehl:
Is this still useful? Any reason not to merge this? @bio-boris

bio-boris:
@briehl Fixed the URL. We can merge it in when the build passes. Thanks

PR #2076, opened 2021-01-14
fix SCT-2914 / PUBLIC-1493 - movement of unselected cells

eapearson:

Fixes an issue with using the up and down arrow cell movement controls in unselected cells. Rather than the the unselected moving, the selected cell moves.

In addition, the styling of the toolbar controls in unselected cells is a bit too dim, and they don't respond well to hovering.

The fix for the cell movement is to have the cell movement commands include the index of the target (unselected) cell. The styling issue was a matter of adding classes to better support unselected cell styles, and altering the styles to make the controls a bit darker (they were double-subdued by being both a gray color and also having low opacity applied, which was probably unintentional), and in addition when hovering they appear as in selected cells (full color, background color applied.)

To support integration tests, some additional markup changes were required.

- fix movement of unselected cells
- improve styling of toolbar controls in unselected cells
- integration tests for the above, for ci and narrative-dev

# Jira Ticket / Issue #
original: https://kbase-jira.atlassian.net/browse/PUBLIC-1493
work tracking: https://kbase-jira.atlassian.net/browse/SCT-2914
- [x] Added the Jira Ticket to the title of the PR 

# Testing Instructions

# Dev Checklist:

- [ ] My code follows the guidelines at https://sites.google.com/truss.works/kbasetruss/development
- [x] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the travis build passes)

### DIFF
--- a/kbase-extension/static/kbase/js/userMenu.js
+++ b/kbase-extension/static/kbase/js/userMenu.js
@@ -82,10 +82,10 @@ define([
                             div({style: 'display:inline-block; width: 34px; vertical-align: top;'}, [
                                 span({class: 'fa fa-user', style: 'font-size: 150%; margin-right: 10px;'})
                             ]),
-                            div({style: 'display: inline-block', 'data-element': 'user-label'}, [
-                                displayName,
+                            div({style: 'display: inline-block'}, [
+                                span({'data-element': 'realname'}, displayName),
                                 br(),
-                                i({}, userName)
+                                i({'data-element': 'username'}, userName)
                             ])
                         ])
                     ]),
--- a/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
+++ b/kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
@@ -22,7 +22,7 @@ define([
 ) {
     'use strict';
 
-    var t = html.tag,
+    const t = html.tag,
         div = t('div'),
         a = t('a'),
         button = t('button'),
@@ -50,11 +50,13 @@ define([
             readOnly = Jupyter.narrative.readonly;
 
         function doMoveCellUp() {
-            Jupyter.notebook.move_cell_up();
+            const cellIndex = Jupyter.notebook.find_cell_index(cell);
+            Jupyter.notebook.move_cell_up(cellIndex);
         }
 
         function doMoveCellDown() {
-            Jupyter.notebook.move_cell_down();
+            const cellIndex = Jupyter.notebook.find_cell_index(cell);
+            Jupyter.notebook.move_cell_down(cellIndex);
         }
 
         function doDeleteCell() {
@@ -199,10 +201,8 @@ define([
         }
 
         function renderOptions(cell, events) {
-            var toggleMinMax = utils.getCellMeta(cell, 'kbase.cellState.toggleMinMax', 'maximized'),
-                toggleIcon = (toggleMinMax === 'maximized' ? 'minus' : 'plus'),
-                dropdownId = html.genId(),
-                menuItems = [];
+            const dropdownId = html.genId();
+            const menuItems = [];
 
             if (cell.cell_type === 'code') {
                 menuItems.push({
@@ -403,12 +403,12 @@ define([
                         ])),
 
                         (function () {
-                            var toggleMinMax = utils.getCellMeta(cell, 'kbase.cellState.toggleMinMax', 'maximized'),
+                            const toggleMinMax = utils.getCellMeta(cell, 'kbase.cellState.toggleMinMax', 'maximized'),
                                 toggleIcon = (toggleMinMax === 'maximized' ? 'minus' : 'plus'),
-                                color = (toggleMinMax === 'maximized' ? '#000' : 'rgba(255,137,0,1)');
+                                classModifier = (toggleMinMax === 'maximized' ? '-maximized' : '-minimized');
                             return button({
                                 type: 'button',
-                                class: 'btn btn-default btn-xs',
+                                class: `btn btn-default btn-xs kb-btn-expander ${classModifier}`,
                                 dataToggle: 'tooltip',
                                 dataPlacement: 'left',
                                 title: true,
@@ -418,10 +418,7 @@ define([
                                 id: events.addEvent({type: 'click', handler: doToggleMinMaxCell})
                             }, [
                                 span({
-                                    class: 'fa fa-' + toggleIcon + '-square-o fa-lg',
-                                    style: {
-                                        color: color
-                                    }
+                                    class: 'fa fa-' + toggleIcon + '-square-o fa-lg'
                                 })
                             ]);
                         }())
--- a/test/integration/specs/kbaseCellToolbarMenu_test.js
+++ b/test/integration/specs/kbaseCellToolbarMenu_test.js
@@ -1,0 +1,317 @@
+/*global describe, it, browser, expect, $, afterEach, beforeEach*/
+/* eslint {strict: ['error', 'global']} */
+'use strict';
+
+const {login, openNarrative, sendString, clickWhenReady} = require('../wdioUtils.js');
+
+/*
+Notes:
+
+The test narrative has 4 markdown cells, each collapsed. The initial collapsed
+state makes some tests easier, because they all have a title.
+*/
+
+const testData = {
+    common: {
+        unselectedBorder: '1px 1px 1px 5px solid rgb(204, 204, 204)',
+        selectedBorder: '1px 1px 1px 5px solid rgb(75, 184, 86)',
+    },
+    all: {
+        TEST_CASE1: {
+            cellIndex: 1,
+            title: "Narrative Cell Toolbar Testing"
+        },
+        TEST_CASE2: {
+            cellIndex: 2,
+            title: "Cell 1"
+        },
+        TEST_CASE3: {
+            cellIndex: 3,
+            title: "Cell 2"
+        },
+        TEST_CASE4: {
+            cellIndex: 3,
+            title: "Cell 2"
+        },
+        TEST_CASE5: {
+            cellIndex: 4,
+            body: "Cell 3"
+        }
+    },
+    ci: {
+        TEST_CASE1: {
+            narrativeId: 58675,
+        },
+        TEST_CASE2: {
+            narrativeId: 58675,
+        },
+        TEST_CASE3: {
+            narrativeId: 58675,
+        },
+        TEST_CASE4: {
+            narrativeId: 58675,
+        },
+        TEST_CASE5: {
+            narrativeId: 58675,
+        }
+    },
+    'narrative-dev': {
+        TEST_CASE1: {
+            narrativeId: 80970,
+        },
+        TEST_CASE2: {
+            narrativeId: 80970,
+        },
+        TEST_CASE3: {
+            narrativeId: 80970,
+        },
+        TEST_CASE4: {
+            narrativeId: 80970,
+        },
+        TEST_CASE5: {
+            narrativeId: 80970,
+        }
+    }
+};
+
+const testCases = mergeObjects([testData[browser.config.testParams.ENV], testData.all]);
+
+function mergeObjects(listOfObjects) {
+    const simpleObjectPrototype = Object.getPrototypeOf({});
+
+    function isSimpleObject(obj) {
+        return Object.getPrototypeOf(obj) === simpleObjectPrototype;
+    }
+
+    function merge(obj1, obj2, keyStack) {
+        Object.keys(obj2).forEach(function (key) {
+            const obj1Value = obj1[key];
+            const obj2Value = obj2[key];
+            const obj1Type = typeof obj1Value;
+            // var obj2Type = typeof obj2Value;
+            if (obj1Type === 'undefined') {
+                obj1[key] = obj2[key];
+            } else if (isSimpleObject(obj1Value) && isSimpleObject(obj2Value)) {
+                keyStack.push(key);
+                merge(obj1Value, obj2Value, keyStack);
+                keyStack.pop();
+            } else {
+                console.error('UNMERGABLE', obj1Type, obj1Value);
+                throw new Error('Unmergable at ' + keyStack.join('.') + ':' + key);
+            }
+        });
+    }
+
+    const base = JSON.parse(JSON.stringify(listOfObjects[0]));
+    for (let i = 1; i < listOfObjects.length; i += 1) {
+        merge(base, listOfObjects[i], []);
+    }
+    return base;
+}
+
+
+
+async function waitForCell(notebookContainer, cellIndex){
+    return await browser.waitUntil(async () => {
+        const cell = await notebookContainer.$(`.cell:nth-child(${cellIndex})`);
+        return cell;
+    });
+    // return await panel.$$('[role="table"][data-test-id="result"] > div > [role="row"]');
+}
+
+async function waitForCellWithTitle(container, cellIndex, title) {
+    const cell = await waitForCell(container, cellIndex);
+    await browser.waitUntil(async () => {
+        const titleElement = await cell.$('[data-element="title"]');
+        const text = await titleElement.getText();
+        return text === title;
+    });
+    return cell;
+}
+
+async function waitForCellWithBody(container, cellIndex, bodyText) {
+    const cell = await waitForCell(container, cellIndex);
+    await browser.waitUntil(async () => {
+        const element = await cell.$('.text_cell_render.rendered_html');
+        const text = await element.getText();
+        return text === bodyText;
+    });
+    return cell;
+}
+
+async function selectCell(container, cellIndex, title) {
+    const cell = await waitForCellWithTitle(container, cellIndex, title);
+
+    // Make sure not selected.
+    // We do this by inspecting the border.
+    const borderStyle = await cell.getCSSProperty('border');
+    expect(borderStyle.value).toEqual(testData.common.unselectedBorder);
+
+    // Click on title area to select cell.
+    const titleArea = await cell.$('.title-container');
+    await clickWhenReady(titleArea);
+
+    await browser.waitUntil(async () => {
+        const borderStyle = await cell.getCSSProperty('border');
+        return borderStyle.value === testData.common.selectedBorder;
+    });
+
+    return cell;
+}
+
+async function selectCellWithBody(container, cellIndex, body) {
+    const cell = await waitForCellWithBody(container, cellIndex, body);
+
+    // Make sure not selected.
+    // We do this by inspecting the border.
+    const borderStyle = await cell.getCSSProperty('border');
+    expect(borderStyle.value).toEqual(testData.common.unselectedBorder);
+
+    // Click on title area to select cell.
+    const bodyArea = await cell.$('.text_cell_render.rendered_html');
+    await clickWhenReady(bodyArea);
+
+    await browser.waitUntil(async () => {
+        const borderStyle = await cell.getCSSProperty('border');
+        return borderStyle.value === testData.common.selectedBorder;
+    });
+
+    return cell;
+}
+
+describe('Test kbaseCellToolbarMenu', () => {
+    beforeEach(async () => {
+        await browser.setTimeout({ 'implicit': 30000 });
+        await browser.reloadSession();
+    });
+
+    afterEach(async () => {
+        await browser.deleteCookies();
+    });
+
+    it('moves a minimized selected cell down', async () => {
+        const testCase = testCases.TEST_CASE1;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
+
+        const cell = await waitForCellWithTitle(narrativeContainer, testCase.cellIndex, testCase.title);
+
+        // Find and click the move-down button
+        const downButton = await cell.$('[data-test="cell-move-down"]');
+        await clickWhenReady(downButton);
+        await waitForCellWithTitle(narrativeContainer, testCase.cellIndex + 1, testCase.title);
+    });
+
+    it('moves a minimized unselected cell down', async () => {
+        const testCase = testCases.TEST_CASE2;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
+
+        const cell = await waitForCellWithTitle(narrativeContainer, testCase.cellIndex, testCase.title);
+
+        // Find and click the move-down button
+        const downButton = await cell.$('[data-test="cell-move-down"]');
+        await clickWhenReady(downButton);
+        await waitForCellWithTitle(narrativeContainer, testCase.cellIndex + 1, testCase.title);
+    });
+
+    it('moves a minimized unselected cell up', async () => {
+        const testCase = testCases.TEST_CASE3;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
+
+        const cell = await waitForCellWithTitle(narrativeContainer, testCase.cellIndex, testCase.title);
+
+        // Find and click the move-ups button
+        const upButton = await cell.$('[data-test="cell-move-up"]');
+        await clickWhenReady(upButton);
+        await waitForCellWithTitle(narrativeContainer, testCase.cellIndex - 1, testCase.title);
+    });
+
+     // select cell
+    it('selects a minimized cell', async () => {
+        const testCase = testCases.TEST_CASE4;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
+        await selectCell(narrativeContainer, testCase.cellIndex, testCase.title);
+    });
+
+    // select cell and move down
+    it('selects a minimized cell and moves it down', async () => {
+        const testCase = testCases.TEST_CASE4;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
+        const cell = await selectCell(narrativeContainer, testCase.cellIndex, testCase.title);
+        const downButton = await cell.$('[data-test="cell-move-down"]');
+        await clickWhenReady(downButton);
+        await waitForCellWithTitle(narrativeContainer, testCase.cellIndex + 1, testCase.title);
+    });
+
+    // select cell and move up
+    it('selects a minimized cell and moves it up', async () => {
+        const testCase = testCases.TEST_CASE4;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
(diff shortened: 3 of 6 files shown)


eapearson:
This should be pretty much ready to go. New and old integration tests pass for ci and narrative-dev. 
Code coverage falling by 0.01% on a 13.68% coverage codebase should not be a reason to hold off, imo;
local front end tests are currently cranky as heck.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
```diff
@@ -403,12 +403,12 @@ define([
                         ])),
 
                         (function () {
-                            var toggleMinMax = utils.getCellMeta(cell, 'kbase.cellState.toggleMinMax', 'maximized'),
+                            const toggleMinMax = utils.getCellMeta(cell, 'kbase.cellState.toggleMinMax', 'maximized'),
                                 toggleIcon = (toggleMinMax === 'maximized' ? 'minus' : 'plus'),
-                                color = (toggleMinMax === 'maximized' ? '#000' : 'rgba(255,137,0,1)');
+                                classModifier = (toggleMinMax === 'maximized' ? '-maximized' : '-minimized');
```
ialarmedalien, line 408: 👍

test/integration/specs/kbaseCellToolbarMenu_test.js
```diff
@@ -0,0 +1,327 @@
...
+
+*/
+
+const testData = {
+    common: {
+        unselectedBorder: '1px 1px 1px 5px solid rgb(204, 204, 204)',
+        selectedBorder: '1px 1px 1px 5px solid rgb(75, 184, 86)',
+    },
+    all: {
+        TEST_CASE1: {
```
ialarmedalien, line 28: It's a bit difficult to read the number with the string written like this -- could you change it to either `testCase1` or `TEST_CASE_1`, so the digit is clearer?

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
```diff
@@ -418,10 +418,7 @@ define([
                                 id: events.addEvent({type: 'click', handler: doToggleMinMaxCell})
                             }, [
                                 span({
-                                    class: 'fa fa-' + toggleIcon + '-square-o fa-lg',
-                                    style: {
-                                        color: color
-                                    }
+                                    class: 'fa fa-' + toggleIcon + '-square-o fa-lg'
```
ialarmedalien, line 421: `fa fa-${toggleIcon}-square-o fa-lg`?

test/integration/specs/kbaseCellToolbarMenu_test.js
```diff
@@ -0,0 +1,327 @@
...
+        TEST_CASE4: {
+            narrativeId: 80970,
+        },
+        TEST_CASE5: {
+            narrativeId: 80970,
+        }
+    }
+};
+
+const testCases = mergeObjects([testData[browser.config.testParams.ENV], testData.all]);
```
ialarmedalien, line 85: Wouldn't it be simpler to do something like:

``` js
const narrativeIdsByEnv = {
  ci: 58675,
  'narrative-dev': 80970,
},
narrativeEnv = browser.config.testParams.ENV;
if (!narrativeIdsByEnv[narrativeEnv]) {
// bail early 
}
const narrativeId = narrativeIdsByEnv[narrativeEnv];

// in the tests
it('moves a minimized selected cell down', async () => {
  const testCase = testCases.TEST_CASE1;
  await login();
  const narrativeContainer = await openNarrative(narrativeId);
  // etc.
```

?

test/integration/specs/kbaseCellToolbarMenu_test.js
```diff
@@ -0,0 +1,327 @@
...
+        return text === bodyText;
+    });
+    return cell;
+}
+
+async function selectCell(container, cellIndex, title) {
+    const cell = await waitForCellWithTitle(container, cellIndex, title);
+
+    // Make sure not selected.
+    // We do this by inspecting the border.
```
ialarmedalien, line 151: This seems like a rather brittle way to test the selection status of the cell; would it be better to check what classes the cell has instead? That way at least the tests aren't sensitive to small UI changes.

test/integration/specs/kbaseCellToolbarMenu_test.js
```diff
@@ -0,0 +1,327 @@
...
+    const bodyArea = await cell.$('.text_cell_render.rendered_html');
+    await clickWhenReady(bodyArea);
+
+    await browser.waitUntil(async () => {
+        const borderStyle = await cell.getCSSProperty('border');
+        return borderStyle.value === testData.common.selectedBorder;
+    });
+
+    return cell;
+}
```
ialarmedalien, line 190: This is almost identical to the function above. I would refactor this code to create a more general `selectCellPart` function, which has an additional param that indicates whether the title or the body should be selected.

test/integration/specs/kbaseCellToolbarMenu_test.js
```diff
@@ -0,0 +1,327 @@
...
+        const testCase = testCases.TEST_CASE4;
+        await login();
+        const narrativeContainer = await openNarrative(testCase.narrativeId);
+        const cell = await selectCell(narrativeContainer, testCase.cellIndex, testCase.title);
+        const upButton = await cell.$('[data-test="cell-move-up"]');
+        await clickWhenReady(upButton);
+        await waitForCellWithTitle(narrativeContainer, testCase.cellIndex - 1, testCase.title);
+    });
+
+    // Everything above, but cells are expanded.
```
ialarmedalien, line 271: If you're repeating a set of tests with a single variable changed, I would rewrite my testing functions to parameterise the variable; it makes the code easier to maintain and there's less chance of a copy-paste error. ISTM you could write a more generic testing function that takes in four params--expanded/collapsed state, selection state, and movement direction, plus test case data--and use that to test all combinations of parameters.

Each test also starts with `await login()`, so that can go into the `beforeEach()` function.

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: okay, nice.
btw, I'm also applying changes like this which affect integration test  style of the (two) other existing integration test files.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
eapearson: This was already in place, not changed by the PR. The change was to remove the explicit color style.
But maybe I don't know what the question is?
The `toggleIcon` in this case is the functional part of the font awesome class name, which in this case is either fa-minus-square-o or fa-plus-square-o , so basically toggles between the minus and plus icons.
I was tempted to, but did not, improve that line by making it a template string.

ialarmedalien: The comment was just about using a template string. ATM we don't have any policies either way so it's up to you. 🙂

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: An update to the PR will include a general solution to this.
I had punted on this, because I wanted to keep the test data confined to a single json-compatible structure -- probably better to keep the test data in a file at some point, as it can get pretty large for some test cases, and keeping it in the set file, while simpler for now, makes the test files bigger.

ialarmedalien: Thinking about this a bit more and looking at the other existing integration tests (which aren't in the `truss` / bulk import cell branch yet), I think it would make sense to have a utility function to get the narrative ID for a given test in a certain environment, and to skip the tests if there isn't a narrative available for that test in that env. Something like

```js
// in a utils file
function getTestNarrativeId(testId) {
  const tests = {
    kbaseCellToolbarMenu: {
      ci: 58675,
      'narrative-dev': 80970,
    },
    kbaseCellSidebarMenu: {
      ci: 12345,
    },
    ...
  };

  const env = browser.config.testParams.ENV;
  if (!tests[testId]) {
    throw new Error(`Invalid testId ${testId}`);
  }
  if (!tests[testId][env]) {
    return null;
  }
  return tests[testId][env];
}
```

```js
// in the test file
const testId = 'kbaseCellToolbarMenu'; // could derive dynamically from the file name
const narrativeId = testUtils.getTestNarrativeId(testId);
if (!narrativeId) {
  // skip the tests - no narrative available in this env
}
```

That would save from having repeated boilerplate in each integration test. 

(Going a step further, you could enclose the tests in a function, and have a `runTests` util function to which you pass the test ID and the test function; `runTests` could either execute the function (thereby running the tests) or could skip that test set if there was no narrative available.)

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: There will be an update to this. Not quite the way you suggested, but close in spirit.

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: In this case, the test is actually to ensure that the cell being targeted also has the correct content, in this case title. It is first selected by position (index) and then the text of the title area is checked. This relieves the burden on test from having to know the precise selector for locating a cell title (or body, the other variant of this method.)

This function has moved in the latest set of changes, but there is also a method to wait for a cell purely by the index.

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: Is addressed in an update to the PR.

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: The idea was to eventually move all test data into a separate json file, so they all live in a single test data object.
The test data has since been moved into a companion test data file.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
eapearson: Oh, right, well, I have tried to restrain myself from just changing code to be better. I do try to restrain myself to only changing code if
(a) it is required for the fix
(b) it is a latent bug not described in the bug report
(c) it is seriously required to have to code make sense (a big exception is any jQuery code, which never makes sense.)
(d) it is required for a test or makes a test easier or more reliable, e.g. an attribute to select on
(e) it angers eslint
I'm sure there are others ...

That said, I do sometimes just make code improvements if they bug me enough and I'm in a generous mood.
In this case, string concatenation is old fashioned, but there are literally hundreds of not thousands of locations in the codebase where they are used and can be transformed into template strings :)
It sure is satisfying to do it, though :)

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: I'm sure the integration test support will continually evolve. There are few enough tests that it is easy to refactor them all in a few minutes when we do change that support.

Since the original PR, I refactored the narrative-centric util methods into a NarrativeTesting class, which knows about the structure of test data utilized by the tests I've written. An instance is created for each test, given the test data, a test case key, the environment. Some things are assumed to be in the global test config, which I'm not comfortable with ... but it is a process.

(BTW some of my comments may have come after yours but not reflect them, because this page was stale.)

Actually, globally, I'm not sure that this is the best testing framework to use anyway. It is actually quite easy to create false positive test just by missing an await ... the promises are uncontrolled, and the test completes without error, although the promises may actual resolve to an error. It is asking a lot for devs to remember when they need to use await (which is just about always when using the WebdriverIO api).

I found earlier that the synchronous tests did not work well, especially if one needs to wrap them and mix in custom test code, but that may have changed since, and I may not have tried hard enough.

test/integration/specs/kbaseCellToolbarMenu_test.js
eapearson: I did quite a round of test refactoring to be more DRY, and fixed some false test positives.
Basically, I was getting stretched by the amount of test time to actual bug fixing time, and just stopped when the tests reported success (and a round or two of refactoring them.) That is partly because the integration testing is so new, and we are still building up support and getting experience.

eapearson:
We'll see how things are when the token is updated and the builds succeed. 
I don't think the SonarCloud stuff is anything to worry about at. The code duplication can be useful, but is also wonky at times, and for tests maybe not worth making it happy. Many tests have similar code which uses different values. Tests which hide too much of the actual test steps can be difficult to understand and debug. And a 0.01% drop in coverage trivial (and I've had trouble getting the unit tests to work at all because of the reliance on CI for some of the, so adding the trivial testing to push that get coverage neutral or positive was out of reach, but maybe it is worth pushing that until I can get them to succeed?)

test/integration/wdio.conf.js
```diff
@@ -6,6 +6,9 @@
 const testConfig = require('../testConfig');
 const fs = require('fs');
 
+const CHROME_BINARY = require('puppeteer').executablePath();
+// process.env.
```
briehl, line 10: Is this comment needed?

test/integration/wdioUtils.js
```diff
@@ -58,64 +58,18 @@ async function clickWhenReady(el) {
...
-    
-    // Ensure narrative notebook has displayed
-    // TODO: more interesting waitUntil loop to signal the 
-    // failure reason (useful for debugging tests?)
-    const container = await $('#notebook-container');
-    await container.waitForDisplayed({
-        timeout,
-        timeoutMsg: `Timeout after waiting ${timeout}ms for narrative to appear`
-    });
+function* range(start, end) {
```
briehl, line 61: Does this function get used anywhere? I can't find it in the PR.

test/integration/NarrativeTesting.js
```diff
@@ -0,0 +1,189 @@
...
+        // Ensure logged in
+        const loginButton = await $('#signin-button > div > button');
+        await loginButton.waitForDisplayed({
+            timeout,
+            timeoutMsg: `Timeout after waiting ${timeout}ms for login button to appear`
+        }); 
+
+        await clickWhenReady(loginButton);
+
+        const realnameElement = await $('[data-element="realname"]');
```
briehl, line 118: :+1: I see what that was used for now. :)

briehl (commented):
I'm going through this now and catching up on a week and a half of work. There's a lot to process here, but it looks good to me, I don't see a lot of boilerplate or repetition that some of the conversation pointed out.

@AJ would you mind taking another glance at this if you get a chance? And @eapearson would you mind fixing the `package.json` conflict? I think this can get merged soon. Thanks for all the work!

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
ialarmedalien: (c) - :grin:

Template strings can be implemented globally by `eslint`, should we wish to adopt a policy towards them at some point.

test/integration/specs/kbaseNarrativeSidePublicTab_test.js
```diff
@@ -344,173 +54,144 @@ async function openPublicData() {
...
+        await row.scrollIntoView();
+        await expect(row).toBeVisible();
+        await testRow(row, t);
     });
 
-    it('opens the public data search tab, searches for a term, should find an expected row', async () => {
-        const testCase = testCases.TEST_CASE3;
+    it('does nothing', async () => {
+        await expect(1).toEqual(1);
+    });
```
ialarmedalien, line 150: ??

test/integration/specs/narrative_basic_test.js
```diff
@@ -5,13 +5,13 @@ const env = process.env.ENV || 'ci';
 
 const allTestCases = {
     ci: {
-        CASE1: {
+        CASE_1: {
```
ialarmedalien, line 8: Can you standardise these to use `TEST_CASE_x` too, please?

test/testConfig.json
```diff
@@ -1,12 +1,14 @@
 {
     "token": {
-        "user": "wjriehl",
-        "file": "test/wjriehl.tok"
+        "user": "narrativetest",
+        "file": "test/narrativetest.tok"
```
ialarmedalien, line 4: 🎉

test/integration/NarrativeTesting.js
```diff
@@ -0,0 +1,189 @@
...
+        });
+    }
+
+    async selectCell(cell) {
+        // Make sure not selected.
+        // We do this by inspecting the border.
+        await browser.waitUntil(async () => {
+            const borderStyle = await cell.getCSSProperty('border');
+            return borderStyle.value === this.testData.common.unselectedBorder;
+        });
```
ialarmedalien, line 153: Using the cell border to determine whether or not a cell is selected seems like a rather brittle way to test it -- if you are going to use cell styling to determine the state, I would have thought it would be better to see whether or not the cell has the class `selected`. You don't want presentation changes to affect the JS tests.

test/integration/NarrativeTesting.js
```diff
@@ -0,0 +1,189 @@
...
+
+    async waitForCellWithText(container, cellIndex, selector, text) {
+        await browser.waitUntil(async () => {
+            const cell = await this.waitForCell(container, cellIndex);
+            const element = await cell.$(selector);
+            const cellText = await element.getText();
+            return text === cellText;
+        });
+        return await this.waitForCell(container, cellIndex);
+    }
```
ialarmedalien, line 175: 👍

test/integration/specs/kbaseCellToolbarMenu_test.js
```diff
@@ -0,0 +1,245 @@
...
+    });
+
+    it('selects and moves an expanded last cell down', async () => {
+        await testCellMovement({
+            caseLabel: 'TEST_CASE_5',
+            direction: 'down',
+            selectCell: true
+        });
+    });
+});
```
ialarmedalien, line 245: You have already done some lovely refactoring, and the tests in this file are prime candidates for further refactoring. You are consistently altering three dimensions (caseLabel, direction, selectCell), so they can go into a set of loops, and you can generate the text for the `it('...')` function from the variables. I would add a brief description to each test case JSON object, i.e.:

```json
        "TEST_CASE_1": {
             "cellIndex": 1,
             "title": "Narrative Cell Toolbar Testing",
             "description": "minimised first cell",        
             ...
        },
        ...
        "TEST_CASE_5": {
             "cellIndex": 5,
             "body": "Cell 4",
             "description": "expanded last cell",        
             ...
        },
```

Then you can do something like:
```js
const directions = ['up', 'down'];
const doSelectCell = [true, false];

// iterate through the test cases
// testCases is an object with keys TEST_CASE_x and values the test case data
Object.keys(testCases).forEach((testCaseId) => {
  directions.forEach((direction) => {
    doSelectCell.forEach((selectCell) => {
      let testDescription = `moves the ${testCases[testCaseId].description} ${direction}`;
      if (selectCell) {
        testDescription = `selects and ${testDescription}`;
      }
      it(testDescription, async() => {
        await testCellMovement({
          case: testCaseId,
          direction: direction,
          selectCell: selectCell
        });
      });
   });
});

File: test/integration/specs/kbaseCellToolbarMenu_data.json [hunk 17]
```diff
@@ -0,0 +1,165 @@
...
+            {
+                "title": "Cell 2"
+            },
+            {
+                "body": "Cell 3"
+            },
+            {
+                "body": "Cell 4"
+            }
+        ]
```
ialarmedalien, line 21: this information already seems to be in the `cases` object -- does it need to be repeated here?

test/integration/NarrativeTesting.js
```diff
@@ -0,0 +1,189 @@
...
+        return mergeObjects([caseData, envDefaults, envcaseData]);
+    }
+
+    /**
+    * Navigates the wdio browser to a workspace as given by its id, and waits until the 
+    * narrative container is visible.
+    * @arg {number} The narrative's workspace id
+    * @returns {Promise} The Promise value is ignored.
+    */
+    async openNarrative(workspaceId, timeoutOverride) {
```
ialarmedalien, line 89: this is very nice and very useful! 👍

test/integration/NarrativeTesting.js
```diff
@@ -0,0 +1,189 @@
...
+
+        // Note, order is least specific to most specific, so that the more
+        // specific can override the least.
+
+        // Top level test cases provide defaults (non-env-specific) per-case
+        // test data.
+        const caseData = this.testData.cases[this.caseLabel];
+
+        // Each env can establish defaults (e.g. narrative id)
+        const envDefaults = this.testData[env].defaults;
```
ialarmedalien, line 74: what happens if the env doesn't exist in `this.testData`?

test/integration/specs/kbaseNarrativeSidePublicTab_data.json
```diff
@@ -0,0 +1,291 @@
+{
+    "cases": {
+        "TEST_CASE_1": {},
+        "TEST_CASE_2": {},
+        "TEST_CASE_3": {},
+        "TEST_CASE_4": {},
+        "TEST_CASE_5": {}
+    },
```
ialarmedalien, line 8: is this necessary?

test/integration/NarrativeTesting.js
```diff
@@ -0,0 +1,189 @@
...
+    return merged;
+}
+
+class NarrativeTesting {
+    constructor({testData, caseLabel, timeout}) {
+        this.testData = testData;
+        this.caseLabel = caseLabel;
+        this.timeout = timeout;
+        this.caseData = this.getCaseData();
+    }
```
ialarmedalien, line 58: Can you add some documentation about proposed structure of the test case data, so that anyone who is writing their own integration tests can follow the same format?

test/integration/wdio.conf.js
```diff
@@ -6,6 +6,9 @@
 const testConfig = require('../testConfig');
 const fs = require('fs');
 
+const CHROME_BINARY = require('puppeteer').executablePath();
```
ialarmedalien, line 9: 👍

test/integration/specs/kbaseCellToolbarMenu_test.js
ialarmedalien: I've moved this comment to the new location of this code and clarified what I meant.

test/integration/specs/kbaseCellToolbarMenu_test.js
ialarmedalien: (N.b. There are copy-pasta errors in the descriptions for the tests, which is one reason why I would very strongly recommend using nested loops to parameterise and generate your tests.)

ialarmedalien:
Lots of big improvements, but there's still a lot of repetition that could be factored out. Some documentation on the format for the test case objects will help keep them consistent between tests.

test/integration/NarrativeTesting.js
eapearson: It fails, and you would need to create test data for that environment.
Generally, there should be always be test support for a test environment we are going to test against.
There is no support, generally, for creating or validating test data within the integration tests.
I see this type of functionality as a step in that direction.

test/integration/specs/kbaseNarrativeSidePublicTab_data.json
eapearson: This establishes the global test case data, which in this case is absent.
Ideally the test data would be consistent across environments, and most of the test data could live in the global test cases. This can be accomplished for testing certain aspects of test narratives, but creating equivalent narratives. 
For search data, though, unfortunately we load different public data in CI than Prod, which complicates testing, and other areas supporting search.
It might be possible to reconstruct the test data using a search term which produces identical results.

kbase-extension/static/kbase/js/userMenu.js
```diff
@@ -57,61 +57,62 @@ define([
...
+                                {
+                                    href: '/#people/' + userName,
+                                    target: '_blank',
+                                    'data-menu-item': 'userlabel',
+                                },
+                                [
+                                    div(
+                                        {
+                                            style:
+                                                'display:inline-block; width: 34px; vertical-align: top;',
```
ialarmedalien, line 84: can all these inline styles be moved into the relevant stylesheet?

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
```diff
@@ -622,8 +523,9 @@ define([
     }
 
     return {
-        make: function (config) {
-            return factory(config);
+        // eslint-disable-next-line no-unused-vars
```
ialarmedalien, line 526: why not just remove the `_config`?

package.json
```diff
@@ -56,9 +56,14 @@
         "prettier": "prettier --write '**/*.js'",
         "prettier_check": "prettier --check '**/*.js'",
         "prettier_debug": "prettier --debug-check '**/*.js'",
+        "eslint-check": "eslint '**/*.js'",
+        "eslint-pre-push": "scripts/eslint-pre-push.sh",
```
ialarmedalien, line 60: could you move these up so they're under the `eslint` item?

kbase-extension/static/kbase/js/userMenu.js
eapearson: But the goal was not to make as few code changes as possible, whilst adding testing attributes.

package.json
eapearson: I've alphabetized the list, which reordered other entires as well.

kbase-extension/static/kbase/js/userMenu.js
eapearson: Yes, they could be. But the point of changing that file at all was to make a small change to allow it to be testable, not to refactor it.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseCellToolbarMenu.js
eapearson: Just making eslint happy, whilst preserving the original intent (to communicate that a config object is provided). Now removed.

eapearson:
@briehl I think integration tests are failing due to an expired token.

eapearson:
Just went through a [insert favorite adjective] round of merge conflicts. Hopefully all is well afterwards. 
Let's get this puppy finished!

PR #2103, opened 2021-01-26
change hamburger menu link to 'narratives'

briehl:

A tiny PR to change the "dashboard" link to go to the new "narratives" path.
Just a little cleanup that needs to be done before staging the release.

# Testing Instructions

# Dev Checklist:

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,7 +5,8 @@ This is built on the Jupyter Notebook v6.0.2 (more notes will follow).
 
 ### Version NEXT
 - SCT-2778 - convert data slideout, public tab, refseq data source to use searchapi2/rpc api rather than searchapi2/legacy.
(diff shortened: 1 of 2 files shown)


eapearson (approved):

eapearson:
Looks fine to me. Look, no code changes, just html and markdown!
Ideally there would be an integration test to validate the link in the menu. In kbase-ui the integration test actually follows the link too, but I don't know that that is necessary, and only works if one is also testing on a kbase host rather than localhost; kbase-ui is the primary way one gets to /narratives so there is a case for it there.

briehl:
Yeah, I started to make an integration test around that, but decided against it for the same reason - it won't work if running on localhost. Also, if we're just testing that the link goes to the right path, we could just programmatically examine the `href` and make sure it has the `/narratives` path, which is what the code is. I don't know. I'm torn between it feeling like overkill, and feeling like due diligence.

PR #2111, opened 2021-02-01
Fix package lock

briehl:
`package-lock.json` got out of whack with the latest merge from `develop`. This puts it back in sync.


ialarmedalien (approved):
pre-emptive 👍 for whenever the package-lock.json file gets sorted out!

PR #2114, opened 2021-02-03
SCT-2919 - Fix tab selection weirdness ... wrong target had been selected

eapearson:
# Fix bug in tab selection in the data slide out panel

The click handler for tabs had used the wrong target -- since the tab text is overlaid on the tab itself, currentTarget needs to be used since it is the value of the element with the event handler, whereas target is the element that was acted upon. With the wrong element acted upon, the tab would not receive the "active" class, and would not appear to have been selected. The logic of the tab click handler would also exit early if the current tab was active, and would set the current tab to "active" otherwise. However, the code to deactivate a tab would never be applied to previously selected "tabs", since they had originally been selected incorrectly, and thus the tabs would act strangely since subsequent clicks on previously clicked tabs would have no effect, since they were considered active.

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/SCT-2919

# Testing Instructions

# Dev Checklist:

- [-] I have made corresponding changes to the documentation
- [-] New and existing unit tests pass locally with my changes
- [-] Any dependent changes have been merged and published in downstream modules
- [-] (Python) I have run Black and Flake8 on changed Python code manually or with a git precommit hook

- [-] [Version has been bumped](https://semver.org/) for each release


eapearson:
This should be ready to go, very small number of changes, confirmed manually.
However, I will circle back and add tests for tab selection.

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataPanel.js
```diff
@@ -191,11 +191,12 @@ define([
...
                     this.trigger('toggleSidePanelOverlay.Narrative', this.$overlayPanel);
 
-                    // NOTE - this will be missed and a widget will remain active if the panel is closed by means other than clicking this button.
+                    // NOTE - this will be missed and a widget will remain active if the panel 
+                    // is closed by means other than clicking this button.
                     // This should be re-visited at some point.
                     this.deactivateLastRenderedPanel();
 
-                    //once we've clicked it 10 times, meaning we've open and shut the browser 5x, we reveal its TRUE NAME.
+                    // once we've clicked it 10 times, meaning we've open and shut the browser 5x, we reveal its TRUE NAME.
```
ialarmedalien, line 199: Exciting KBase easter egg?!?!

ialarmedalien (dismissed):
👍

test/integration/specs/kbaseNarrativeDataPanel_test.js
```diff
@@ -0,0 +1,180 @@
...
+    it('opens the "Import" tab', async () => {
+        const testCase = testCases.TEST_CASE_1;
+        await login();
+        await openNarrative(testCase.narrativeId);
+
+        const slideoutPanel = await openDataSlideout();
+
+        const [tab, tabPanel] = await selectTab(slideoutPanel, 'import');
+        // TODO: make sure the panel content displayed correctly
+    });
```
ialarmedalien, line 136: Rather than repeating very similar code blocks, can you refactor this code so that you iterate through an array (or object, with the tab names as values) of tabs, running the `selectTab` and `selectTabByLabel` tests for each?

kbase-extension/static/kbase/js/widgets/narrative_core/kbaseNarrativeDataPanel.js
briehl: Yeah, a former KBaser snuck that in an age ago. It didn't hurt anything and was kinda cute, so we left it. :)

test/integration/specs/kbaseNarrativeDataPanel_test.js
briehl: I think the thing here is that it starts each test case from scratch.
Though I suppose the first several lines could be factored into their own little accessory function.

ialarmedalien: Yup -- I meant that each test case would start from scratch, i.e. running the login and open narrative code and then do the specific `selectTab` or `selectTabByLabel` function.

eapearson: There are upsides and downsides to table-based testing like that. I often love doing it, because you can cram a lot of test cases into fewer lines, but the downside is that you lose information in test failures without extra work. And they do test different things in this case. The isolated tests each open the narrative, open the data slideout, and navigate to a tab, which is different than opening a narrative and navigating amongst the tabs. 

These tests provide easy location of which tab fails if a bug is introduced to one of them.

There is a later test which does repeated bashing of tab selection. It is not ideal because it randomizes the tabs, making the test unreproducible, with the benefit of being accomplished in just a few lines of code. I guess that test could be frozen by generating 100 or whatever tab ids and storing in a json file. It is just a matter of that gift we all possess in very limited supply ... time.

Yeah, the 5 line tests could be bundled up into a parameterized test function. Worth? 

There are so many test cases lurking the the possible combinations in actions provided by tab switching and interaction with other narrative elements, and so many attributes one could test, it is worth considering limiting our expenditure to achieve the goal of ensuring that the major usage cases of what we are testing are covered. In this case, I think it is primarily tab operation with clicking on the tab itself, and clicking on the tab label.

We should continue to improve and add tests to cover the myriad usage cases, but, gosh, I'm just happy to have tests the provide decent coverage of features and are largely readable.

ialarmedalien: In case it wasn't completely clear, this is what I am suggesting:

```js
  const tabNames = {
    import: 'Import',
    sharedwithme: 'Shared With Me',
    ...
  }

  Object.keys(tabNames).forEach((tabId) => {
    it(`opens the "${tabName[tabId]}" tab`, async () => {
         const testCase = testCases.TEST_CASE_1;
         await login();
         await openNarrative(testCase.narrativeId);

         const slideoutPanel = await openDataSlideout();
         const [tab, tabPanel] = await selectTab(slideoutPanel, tabId);
         // TODO: make sure the panel content displayed correctly
    });
  });

eapearson: Oh, you mean to hoist the array of tab ids into a top level variable. Sure, added to the test data.
Also have added some additional tests.

ialarmedalien (approved):
👍

PR #2116, opened 2021-02-03
DATAUP-330 fixed narrative viewers tests and codebase

briehl:

This was, and still is, kinda ugly.

The core problem here was that the NarrativeViewers module would, as soon as it was loaded, make some calls to the Narrative Method Store, regardless of whether that was needed or not. So the main fix here was to make it only do that when the `createViewer` function is called. And then cache the result for 5 minutes.

A secondary problem was actually finding a code path that would run it for real so it could be tested live. Turns out this is only invoked by the decrepit `kbaseNarrativeDataCell` widget, which has since been supplanted by other mechanisms. The only narrative I could find to invoke that was an outdated one that was automatically updated to other mechanisms, yet still uses that module. It's here: https://ci.kbase.us/narrative/630 , shared with @ialarmedalien and @eapearson. It's the first cell that triggers it.

I'd like to completely remove that hunk of code, but since it's built in to some legacy narratives, I don't think we can yet. And I gotta move on to other things this sprint, so band-aid it is. I tried to keep changes to a minimum, other than some minor syntax updates (var -> const / let, arrow functions as appropriate).

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/DATAUP-330

# Testing Instructions
Try out the linked narrative, and/or run the unit tests.

# Dev Checklist:

- [x] (Python) I have run Black and Flake8 on changed Python code manually or with a git precommit hook


test/unit/spec/narrative_core/narrativeViewersSpec.js
```diff
@@ -0,0 +1,139 @@
...
+            expect(Viewers).toBeDefined();
+
+            ['getViewerInfo', 'createViewer', 'defaultViewer'].forEach(fn => {
+                expect(Viewers[fn]).toBeDefined();
+            });
+        });
+
+        it('should load viewer info as a promise', () => {
+            return Viewers.getViewerInfo().then((info) => {
+                expect(info.viewers).toBeDefined();
```
ialarmedalien, line 103: You could put these in an array and then iterate through -- saves a line and makes it easy to amend.

test/unit/spec/narrative_core/narrativeViewersSpec.js
```diff
@@ -0,0 +1,139 @@
...
+            'Module.Type2': {
+                view_method_ids: ['Viewers/missing_viewer_app'],
+                landing_page_url_prefix: 'type2',
+                name: 'Type 2'
+            },
+            'Module.Type3': {
+                view_method_ids: ['Viewers/error_viewer_app'],
+                landing_page_url_prefix: 'type3',
+                name: 'Type 3'
+            }
```
ialarmedalien, line 38: is this data used/tested anywhere?

briehl: Yes, that's the mocked result from NMS.list_categories (the mocked out type information)

test/unit/spec/narrative_core/narrativeViewersSpec.js
briehl: Ah, sure, I didn't see that. I didn't touch the older tests, just migrated to a different spot. I'll update.

ialarmedalien (approved):
🦖 👍

PR #2119, opened 2021-02-05
DATAUP-330 fix narrativeLogin tests, add a cleanup function to narrativeLogin

briehl:

This rewrites the tests for the narrativeLogin module to handle promises properly, and mock out calls to the auth service.
This also exposes the clearTokenCheckTimers function from that module. Without this, there's no way to clear the timers, and they seem to just go off anyway during the remainder of the test suite. This would cause, among other things, a redirect to the logout page if tests run for too long. 

That redirection is something that's definitely one of the causes of our test miseries, and having a tighter control on how that happens is necessary.

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/DATAUP-330

# Testing Instructions
Yet again, there should be no visible narrative changes. The only actual code change is to expose that interval clearing function, and it's only used by the test suite.

# Dev Checklist:

### DIFF
--- a/kbase-extension/static/kbase/js/narrativeLogin.js
+++ b/kbase-extension/static/kbase/js/narrativeLogin.js
@@ -295,10 +295,10 @@ define ([
          * communication with other KBase resources.
          */
         clearTokenCheckTimers();
-        var sessionToken = authClient.getAuthToken();
+        const sessionToken = authClient.getAuthToken();
         return Promise.all([authClient.getTokenInfo(sessionToken), authClient.getUserProfile(sessionToken)])
-            .then(function(results) {
-                var tokenInfo = results[0];
+            .then((results) => {
+                let tokenInfo = results[0];
                 sessionInfo = tokenInfo;
                 this.sessionInfo = tokenInfo;
                 this.sessionInfo.token = sessionToken;
@@ -306,20 +306,21 @@ define ([
                 this.sessionInfo.user_id = this.sessionInfo.user;
                 initEvents();
                 initTokenTimer(sessionInfo.expires);
-                UserMenu.make({
+                if (!noServer) {
+                    ipythonLogin(sessionToken);
+                }
+                $(document).trigger('loggedIn', this.sessionInfo);
+                $(document).trigger('loggedIn.kbase', this.sessionInfo);
+                const userMenu = UserMenu.make({
                     target: $elem,
                     token: sessionToken,
                     userName: sessionInfo.user,
                     email: results[1].email,
                     displayName: results[1].display
                 });
-                if (!noServer) {
-                    ipythonLogin(sessionToken);
-                }
-                $(document).trigger('loggedIn', this.sessionInfo);
-                $(document).trigger('loggedIn.kbase', this.sessionInfo);
-            }.bind(this))
-            .catch(function(error) {
+                return userMenu.start();
+            })
+            .catch((error) => {
                 console.error(error);
                 if (document.location.hostname.indexOf('localhost') !== -1 ||
                     document.location.hostname.indexOf('0.0.0.0') !== -1) {
--- a/kbase-extension/static/kbase/js/narrativeLogin.js
+++ b/kbase-extension/static/kbase/js/narrativeLogin.js
@@ -105,7 +105,7 @@ define ([
     function showNotLoggedInDialog() {
         const message = `
         <p>You are logged out (or your session has expired).</p>
-        <p>You will be redirected to the sign in page after closing this, or ${AUTO_LOGOUT_DELAY / 1000} seconds, 
+        <p>You will be redirected to the sign in page after closing this, or ${AUTO_LOGOUT_DELAY / 1000} seconds,
            whichever comes first.</p>
         `;
         var dialog = new BootstrapDialog({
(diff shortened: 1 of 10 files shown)


kbase-extension/static/kbase/js/narrativeLogin.js
```diff
@@ -105,7 +105,7 @@ define ([
     function showNotLoggedInDialog() {
         const message = `
         <p>You are logged out (or your session has expired).</p>
-        <p>You will be redirected to the sign in page after closing this, or ${AUTO_LOGOUT_DELAY / 1000} seconds, 
+        <p>You will be redirected to the sign in page after closing this, or ${AUTO_LOGOUT_DELAY / 1000} seconds,
```
briehl, line 108: A few whitespace changes crept in here when I saved the file.

test/unit/spec/narrativeLoginSpec.js
```diff
@@ -1,19 +1,113 @@
...
+        it('Should instantiate and have expected functions', () => {
+            expect(Login).toBeDefined();
+            ['init', 'sessionInfo', 'getAuthToken', 'clearTokenCheckTimers'].forEach(item => {
+                expect(Login[item]).toBeDefined();
+            });
+        });
+
+        it('Should instantiate on a DOM node in the happy path', () => {
+            const $node = $('<div>'),
+                fakeUser = 'some_user',
```
briehl, line 64: This blob of boilerplate should get added to the mocks module when that gets ported over from the `truss` branch.

test/unit/spec/narrativeLoginSpec.js
```diff
@@ -1,19 +1,113 @@
...
-            Login.init($node);
+        it('Should throw an error when instantiating with a bad token', () => {
+            const $node = $('<div>');
+            mockAuthRequest('token', {error: {}}, 401);
+            mockAuthRequest('me', {error: {}}, 401);
+            return Login.init($node, true)  // true here means there's no kernel
+                .then(() => {
+                    const request = jasmine.Ajax.requests.mostRecent();
+                    expect(request.requestHeaders.Authorization).toEqual(FAKE_TOKEN);
+                    Login.clearTokenCheckTimers();
```
ialarmedalien, line 109: It's difficult to tell from the tests that a DOM node has been instantiated in the first case and an error thrown in the second. What are the observable effects of the error being thrown (e.g. DOM structure changes?), and can they be tested here?

briehl: This one was painful to work through. There's some asynchronous effects that happen in the narrativeLogin code at this point. The basic flow the `.init` function does is:
1. Talk to auth to validate the token and get the user profile.
2. Start up a timer to revalidate every second or so (without making an auth call, this timer also comes in to play if the user's computer went to sleep and wakes back up - if that's over some interval, THEN it makes an auth validation)
3. Use that info to initialize the UserMenu object that builds the little profile menu in the upper-right, with the gravatar and such.

The UserMenu then makes a call to the profile service to fetch gravatar info or user-set icon, and puts the image in. It's the part that does the work of manipulating the DOM. Since it does that in its own time, and the Promise isn't resolved by `NarrativeLogin.init`, there's likely no DOM changes that can be tested yet.

So there's 2 options here. Either leave as-is, or modify both the UserMenu code to have an explicit start function that returns a Promise, and pass that through the Login.init.

I'm not particularly against the code modification, in fact it's probably the right thing to do. I'm just trying to not do too many changes all at once here. But if you think it's a good idea, I'll put it in and test it. I think that module is (gasp) not unit tested anyway.

briehl: Ugh, I think I just talked myself into it. I don't think it's a heavy change.

test/unit/spec/narrative_core/kbaseNarrativeAppPanel-spec.js
```diff
@@ -205,35 +299,38 @@ define([
...
-                        kbase_sessionid: 'narrativetest_session'
+                        token: FAKE_TOKEN,
+                        user_id: FAKE_USER,
+                        kbase_sessionid: 'fake_session'
                     });
                 }
             });
 
-            expect(appPanel.appCatalog).toBeNull();
+            spyOn(appPanel, 'spawnCatalogBrowser'); //.and.callThrough();
```
ialarmedalien, line 331: comment left in

test/unit/spec/userMenuSpec.js
```diff
@@ -0,0 +1,98 @@
...
+
+        it('Should start and create a menu in the target node with an empty user profile', () => {
+            const $elem = $('<div>');
+            const userMenu = UserMenu.make({
+                target: $elem,
+                token: TEST_TOKEN,
+                userName: TEST_USER,
+                email: TEST_EMAIL,
+                displayName: TEST_NAME
+            });
```
ialarmedalien, line 59: could have a little helper function to do this, since it's repeated three times

test/unit/spec/narrative_core/kbaseNarrativeAppPanel-spec.js
briehl: Thanks for catching!

test/unit/spec/userMenuSpec.js
briehl: Yup, done.

ialarmedalien (approved):
👍

PR #2250, opened 2021-04-21
DATAUP-312 write Python code for running

briehl:

This includes a few minor fixes, the generated Python code for running the cell, and toggles the action button to run when ready.

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/DATAUP-312

# Testing Instructions
1. You might need to make a new bulk import cell with these changes
2. Fill out all parameters and file paths
3. It should now show check marks in both file type buttons, and make the Run button available to click.
4. Clicking that run button should put a Pythonic error at the bottom of the cell, as the `run_app_bulk` function isn't written in this PR.

# Dev Checklist:

### DIFF
--- a/kbase-extension/static/kbase/js/common/cellComponents/actionButtons.js
+++ b/kbase-extension/static/kbase/js/common/cellComponents/actionButtons.js
@@ -67,7 +67,7 @@ define(['common/html'], (html) => {
                 });
             });
             bus.on('actionButton', (message) => {
-                const action = message.data;
+                const action = message.data.action;
                 runAction(action);
             });
 
--- a/kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
+++ b/kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
@@ -146,6 +146,16 @@ define([
                 });
             });
 
+            fieldWidget.bus.on('validation', (message) => {
+                // only propagate if invalid. value changes come through
+                // the 'changed' message
+                if (!message.isValid) {
+                    paramsBus.emit('invalid-param-value', {
+                        parameter: parameterSpec.id,
+                    });
+                }
+            });
+
             // The 'sync' message is a request for the current model value from the
             // input widget.
             fieldWidget.bus.on('sync', () => {
--- a/kbase-extension/static/kbase/js/common/pythonInterop.js
+++ b/kbase-extension/static/kbase/js/common/pythonInterop.js
@@ -1,5 +1,6 @@
 define([], () => {
     'use strict';
+    const indentString = '    ';
 
     function escapeString(stringValue, delimiter) {
         const delimiterRegex = new RegExp(delimiter, 'g');
@@ -13,7 +14,6 @@ define([], () => {
         return "'";
     }
 
-    const indentString = '    ';
     function makeIndent(level) {
         let retval = '';
         for (let i = 0; i < level; i += 1) {
@@ -21,14 +21,12 @@ define([], () => {
         }
         return retval;
     }
+
     function pythonifyValue(value, options, indentLevel) {
         options = options || {};
         indentLevel = indentLevel || 0;
         switch (typeof value) {
             case 'number':
-                if (value === null) {
-                    return 'None';
-                }
                 return String(value);
             case 'string':
                 return '"' + escapeString(value, options.delimiter || autoDelimiter(value)) + '"';
@@ -49,13 +47,12 @@ define([], () => {
                 if (value === null) {
                     return 'None';
                 }
-                var prefix = makeIndent(indentLevel + 1);
                 return (
                     '{\n' +
                     Object.keys(value)
                         .map((key) => {
                             return (
-                                prefix +
+                                makeIndent(indentLevel + 1) +
                                 pythonifyValue(key, options) +
                                 ': ' +
                                 pythonifyValue(value[key], options, indentLevel + 1)
@@ -89,6 +86,17 @@ define([], () => {
             });
     }
 
+    /**
+     * Builds a "nice" list of args by adding indentations and returns between each.
+     * E.g., turns ["foo", "bar", "baz"] into:
+     * "
+     *     foo,
+     *     bar,
+     *     baz
+     * "
+     * @param {array} args - the array of arguments to make pretty
+     * @returns
+     */
     function buildNiceArgsList(args) {
         const indent = indentString;
         return '\n' + indent + args.join(',\n' + indent) + '\n';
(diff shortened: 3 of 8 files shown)


kbase-extension/static/kbase/js/common/cellComponents/actionButtons.js
```diff
@@ -67,7 +67,7 @@ define(['common/html'], (html) => {
                 });
             });
             bus.on('actionButton', (message) => {
-                const action = message.data;
+                const action = message.data.action;
```
briehl, line 70: The "action" being run was the object `{action: 'someString'}` when it should be just `'someString'`.

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
```diff
@@ -146,6 +146,16 @@ define([
                 });
             });
 
+            fieldWidget.bus.on('validation', (message) => {
```
briehl, line 149: When an input widget (specifically a text field - in the test cases, the float and int fields) gets modified, it gets validated internally. If it's invalid, the model doesn't get updated, but visually it's still there. It does send a `validation` message up the tree, though. This catches that and invalidates the whole parameter set if so.

For example, if I were to make a completely valid and ready to run set of parameters, then change a numerical field from "0" to, say, "abc", that should raise issues and invalidate the run.

Instead, what's happening right now is the input widget just says "nope, that's an invalid value, and I'm not going to save it", but leaves it in the text field. So it looks ok, but really isn't. Catching the validation message at the `filePathWidget` (and `paramsWidget`) level and propagating it on up to the `configure` widget gets us around that.

I'm not really sure if this is the best behavior - I kinda feel like the user should be able to put in and serialize whatever they value they want. If it's wrong, we should just raise an error and not let them run the cell. But this solves that little hiccup for now.

kbase-extension/static/kbase/js/common/pythonInterop.js
```diff
@@ -1,5 +1,6 @@
 define([], () => {
     'use strict';
+    const indentString = '    ';
```
briehl, line 3: Went through and did some linting and moving around here.

kbase-extension/static/kbase/js/common/pythonInterop.js
```diff
@@ -13,22 +14,19 @@ define([], () => {
...
         }
         return retval;
     }
+
     function pythonifyValue(value, options, indentLevel) {
         options = options || {};
         indentLevel = indentLevel || 0;
         switch (typeof value) {
             case 'number':
-                if (value === null) {
```
briehl, line 29: `null` never shows up as `typeof` `number`, so this can never happen.

kbase-extension/static/kbase/js/common/pythonInterop.js
```diff
@@ -244,16 +251,53 @@ define([], () => {
         return pythonCode;
     }
 
+    /**
+     * Builds the call to run_app_bulk
```
briehl, line 255: Maybe we should chat about this structure. I think this'll be the most flexible down the road for whoever else wants to run it, but it duplicates all of the common parameters. I also considered adding a separate "shared_params" dict to each app that gets run in bulk, and just put it on the kernel to do the duplication before running in EE2. But... I haven't done that yet.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -141,32 +143,8 @@ define([
             busEventManager = BusEventManager.make({
                 bus: runtime.bus(),
             }),
-            typesToFiles = setupFileData(options.importData);
```
briehl, line 144: I did some rearranging here to keep the class variable declarations mostly separate from the function declarations.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -468,27 +472,68 @@ define([
          * Update the internal readiness state for the app cell parameters. When all fileTypes have
          * a "complete" state, then they are all completely filled out and ready to run.
          * @param {string} fileType - which filetype's app state to update
-         * @param {string} state - what the new ready state should be - one of complete, incomplete, error
+         * @param {string} newState - what the new ready state should be - one of complete, incomplete, error
          */
-        function updateParameterState(fileType, state) {
-            model.setItem(['state', 'param', fileType], state);
+        function updateParameterState(fileType, newState) {
```
briehl, line 477: Funny, I think we did the same thing in parallel PRs - as `state` is already a variable owned by this class. :)

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -468,27 +472,68 @@ define([
...
+            updateState(uiState);
             if (cellReady) {
-                console.log('cell ready!');
+                buildPythonCode();
+            } else {
+                clearPythonCode();
             }
         }
 
+        function buildPythonCode() {
```
briehl, line 501: follows the same pattern as other app cells. Get the cell id, make a run id, and send parameters to get Pythonified.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -729,8 +757,40 @@ define([
...
+         */
         function runAction(action) {
-            alert(action);
+            if (readOnly) {
+                console.warn('ignoring attempted action in readonly mode');
+                return;
+            }
+            switch (action) {
+                case 'runApp':
+                    cell.execute();
```
briehl, line 771: `runApp` is the only one implemented right now. Just tells the Jupyter cell to run itself.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -776,6 +836,33 @@ define([
             return uiState;
         }
 
+        /**
+         * Passes the updated state to various widgets, and serialize it in
+         * the cell metadata, where appropriate.
+         * @param {string} newUiState - change to a new UI state, if defined.
+         */
+        function updateState(newUiState) {
+            if (newUiState && newUiState in States) {
```
briehl, line 845: Tweaked this to take in a new uiState from the list of ui states - which tabs are active, etc.

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
ialarmedalien: Yeah - I saw that behaviour when I was cleaning up the tests for some of the input widgets. 😕

kbase-extension/static/kbase/js/common/pythonInterop.js
ialarmedalien, line 255: On a different topic, presumably `run_app_batch` is the current (/old) way of doing things using the BatchApp, and `run_app_bulk` is the new way, using ee2's batch endpoint.

briehl: Yeah, I don't want to just rewrite `run_app_batch` in tricky ways to maintain backward compatibility. I'd rather just deprecate it and make a new wrapper.

ialarmedalien: Yes, it might be worth thinking about / discussing -- there's also the matter of how to display the params in the job status table; we don't really want to repeat the shared params for each job in a batch. I'm also wondering about whether we store the input params in the frontend model object or in the kernel so that info is immediately accessible to the job status table and there doesn't need to be an extra call to ee2 for every job in the table.

nbextensions/bulkImportCell/bulkImportCell.js
ialarmedalien: Great minds, as they say... ;)

test/unit/spec/common/pythonInterop-Spec.js
```diff
@@ -2,114 +2,89 @@ define(['common/pythonInterop'], (PythonInterop) => {
...
-            expect(PythonInterop.pythonifyValue(jsValue)).toEqual(pyValue);
-        });
-        it('Pythonify an array of arrays', () => {
-            const jsValue = [
+        /* these are all 3-tuples:
+         * label (for the test label)
+         * input value
+         * expected output value
+         */
+        const pythonifyValueCases = [
```
ialarmedalien, line 36: Much more concise! I like a nice bit of DRY testing. :D

ialarmedalien (approved):
👍

kbase-extension/static/kbase/js/common/pythonInterop.js
briehl: I think the kernel can hold it pretty easily. It might have a slightly higher memory footprint on the server, but it's easy enough to make a request for that information. I think at kernel startup when it fetches all the job info, it can probably fetch and cache all of that pretty quickly, then we can have the FE request it as needed.

PR #2255, opened 2021-04-27
Bump chromedriver to ^90.0.0

briehl:

Integration tests are all failing hard due to an out of date chromedriver. And dependabot is only doing v 89, and those are failing, too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1534,9 +1534,9 @@
             }
         },
         "chromedriver": {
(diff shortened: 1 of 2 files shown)


ialarmedalien (approved):
I guess at least now the tests are failing for other reasons...?

briehl:
@ialarmedalien Yeah, the secret test token was expired. Trying again with a fresh one.

PR #2325, opened 2021-06-09
update python dependencies

briehl:
Batch version of a gaggle of dependabot updates. The Python deps are pretty reasonable.


ialarmedalien (approved):
👍

PR #2368, opened 2021-07-12
DATAUP-503 Retry jobs in JM, tests in JM, cached state refactor. (Rebased.)

n1mus:

The job states were being cached twice, once in _completed_job_states and once in the job object. This was refactored to consolidate those, and also remove some dead code. This means that the state that is passed over the comms was changed a little big. For example, listener_count was removed since it wasn't used anywhere. spec or app_spec were removed. child_jobs was always an empty list, and that was kept in because it is used in the FE.

retry_jobs in jobmanager.py was changed, courtesy of @ialarmedalien . Tests were updated for that.

Some of the jobcomm.py stuff and corresponding documentation are broken because it hasn't been finalized yet. So, the _retry_jobs in jobcomm.py and two of the corresponding tests. I wasn't sure what messages should be sent, i.e., whether to send the new_jobs message as well as the jobs_retried message. 

# Jira Ticket / Issue #

# Testing Instructions

# Dev Checklist:


src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
             kblogging.log_event(self._log, "list_jobs.error", {"err": str(e)})
             raise
 
-    def _create_error_state(
```
n1mus, line 214: This was never used

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
...
-
-        return {
-            "state": state,
-            "widget_info": widget_info,
-            "owner": job.owner,
-            "listener_count": self._running_jobs[job.job_id]["refresh"],
-        }
-
-    # can be removed -- unused
-    def _child_job_states(self, child_job_list, app_id, app_tag):
```
n1mus, line 380: This was called in such a way that always returned []

src/biokbase/narrative/jobs/job.py
```diff
@@ -12,7 +14,7 @@
 __author__ = "Bill Riehl <wjriehl@lbl.gov>"
 
 COMPLETED_STATUS = "completed"
-TERMINAL_STATES = [COMPLETED_STATUS, "terminated", "error"]
```
n1mus, line 15: Seems like "status" is a one-line thing, "state" is a more complex thing

src/biokbase/narrative/jobs/job.py
```diff
@@ -240,6 +238,76 @@ def state(self):
         except Exception as e:
             raise Exception(f"Unable to fetch info for job {self.job_id} - {e}")
 
+    def revised_state(self, state=None) -> dict:
```
n1mus, line 241: Formerly jm._construct_job_status

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -70,7 +68,6 @@ def initialize_jobs(self):
                 }
             )
             self._running_jobs = dict()
-            self._completed_job_states = dict()
```
n1mus, line 73: Consolidated into Job which also caches state. So you'll have to iterate jm._running_jobs and do job.terminal_state

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -79,39 +76,35 @@ def initialize_jobs(self):
...
             }
-            # if the job is in a terminal state, add it to the _completed_job_states index
-            if not refresh_state:
-                revised_state = self._construct_job_status(
-                    self.get_job(job_id), job_state
-                )
-                self._completed_job_states[job_id] = revised_state
 
-    def _create_jobs(self, job_ids):
+    def _create_jobs(self, job_ids, job_states=None):
```
n1mus, line 85: Sometimes job_states already known. Is check_jobs even expensive?

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
...
-        widget_info = None
-
-        # If there's no job, but the state is valid, then that (likely) means the job was started
-        # by either running AppManager.run_app directly without cell_id or run_id info, or that
-        # it was started outside of the biokbase.narrative.jobs setup. This could be done through
-        # direct calls to EE2.
-        #
-        # This could also be triggered by manually looking up job state for some job that doesn't
-        # exist in the Narrative. Which is borderline, but still probably ok.
-        if job is None and state is not None:
```
n1mus, line 297: jobs was never None

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
...
-                    "run_id": None,
-                }
-            )
-            return {
-                "state": state,
-                "widget_info": widget_info,
-                "owner": None,
-            }
-
-        if state is None:
```
n1mus, line 310: Forget why I got rid of this. Probably state was never None

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
...
-                }
-            )
-            if "batch_size" in job.extra_data:
-                state.update({"batch_size": job.extra_data["batch_size"]})
-
-        return {
-            "state": state,
-            "widget_info": widget_info,
-            "owner": job.owner,
-            "listener_count": self._running_jobs[job.job_id]["refresh"],
```
n1mus, line 376: Never used in narrative repo

src/biokbase/narrative/tests/data/ee2_job_test_data.json
```diff
@@ -135,5 +135,37 @@
                 "run_id": "6546971e-757a-4a26-8439-325497c6d843"
             }
         }
+    },
```
n1mus, line 138: An errored state to complement the terminated state for retry testing

src/biokbase/narrative/tests/narrative_mock/mockclients.py
```diff
@@ -34,19 +35,23 @@ def test_my_function(self):
...
-        self.config = ConfigTests()
         self.job_info = self.config.load_json_file(
             self.config.get("jobs", "job_info_file")
         )
-        self.ee2_job_info = self.config.load_json_file(
-            self.config.get("jobs", "ee2_job_info_file")
-        )
         self.test_job_id = self.config.get("app_tests", "test_job_id")
 
+    @property
```
n1mus, line 51: Made this attribute the property type so I could deep copy it to avoid testing bugs

src/biokbase/narrative/tests/narrative_mock/mockclients.py
```diff
@@ -240,7 +245,7 @@ def check_job(self, params):
         job_id = params.get("job_id")
         if not job_id:
             return {}
-        info = self.ee2_job_info.get(job_id, {})
+        info = self.ee2_job_info.get(job_id, {"job_id": job_id, "status": "unmocked"})
```
n1mus, line 248: retry_jobs adds more jobs to the narrative, so this helps test that

src/biokbase/narrative/tests/test_jobmanager.py
```diff
@@ -40,13 +41,52 @@ def create_jm_message(r_type, job_id=None, data=None):
...
+        "widget_info": None,
+        "owner": None,
+    }
+
+
+def get_dne_job_state(job_id):
+    return {"state": {"job_id": job_id, "status": "does_not_exist"}}
+
+
+def get_test_job_states():
```
n1mus, line 64: This sort of simulates that combination of job._augment_ee2_state and job.revised_state

src/biokbase/narrative/tests/test_jobmanager.py
```diff
@@ -390,23 +415,159 @@ def test_cancel_jobs(self):
             self.jm.cancel_job(JOB_COMPLETED)
 
     @mock.patch("biokbase.narrative.jobs.jobmanager.clients.get", get_mock_client)
-    def test_retry_job_good(self):
-        retry_results = self.jm.retry_jobs([JOB_TERMINATED])
+    def test_retry_jobs__success(self):
```
n1mus, line 418: These retry_jobs tests can be refactored as they sort of follow a similar pattern

src/biokbase/narrative/jobs/job.py
```diff
@@ -325,8 +393,7 @@ def is_finished(self):
         Returns True if the job is finished (in any state, including errors or canceled),
         False if its running/queued.
         """
-        status = self.status()
-        return status.lower() in TERMINAL_STATES
+        return self.state().get("status") in TERMINAL_STATUSES
```
ialarmedalien, line 396: 👍

src/biokbase/narrative/jobs/job.py
```diff
@@ -369,3 +436,82 @@ def dump(self):
         """
 
         return {attr: getattr(self, attr) for attr in [*ALL_JOB_ATTRS, "_last_state"]}
+
+    # def _create_error_state(
```
ialarmedalien, line 440: is this handled somewhere else now?

src/biokbase/narrative/jobs/job.py
```diff
@@ -369,3 +436,82 @@ def dump(self):
...
+    #     app_id = the id of the app that all the child jobs are running (format: module/method, like "MEGAHIT/run_megahit")
+    #     app_tag = one of "release", "beta", "dev"
+    #     (the above two aren't stored with the subjob metadata, and won't until we back some more on KBParallel - I want to
+    #     lobby for pushing toward just starting everything up at once from here and letting HTCondor deal with allocation)
+    #     sub_job_list = list of ids of jobs to look up
+    #     """
+
+    #     # TODO
+    #     # THIS IS A DRAFT METHOD
+    #     # NEEDS TO BE RUN/VERIFIED
```
ialarmedalien, line 487: similarly -- what's happening with this method? are you planning to reinstate it?

src/biokbase/narrative/jobs/jobmanager.py
ialarmedalien: It may be useful to just keep an index of the jobs that are completed, without having the job state attached to it -- e.g. 
```
self._completed_job_states = {
   job_id_1: True,
   job_id_2: True,
}
```

Alternatively, you could store the data in the `_running_jobs` dictionary, e.g.
```
self._running_jobs = {
   job_id_1: { "job": <job object>, "completed": False, "refresh_state": 1 },
   job_id_2: { "job": <job obj>, "completed": True, "refresh_state": 0 },
...
}
```
though you lose some of the advantages of having the list of completed jobs immediately available.

src/biokbase/narrative/jobs/jobmanager.py
ialarmedalien: 🤦

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
...
-            "errormsg": error_msg,
-            "error_code": code,
-            "job_id": job_id,
-            "cell_id": cell_id,
-            "run_id": run_id,
-            "created": 0,
-            "updated": 0,
-        }
-
-    def _construct_job_status(self, job: Job, state: dict) -> dict:
```
ialarmedalien, line 250: this method is now part of the Job object, correct?

n1mus: Yeah, I changed the name to job.revised_state

src/biokbase/narrative/jobs/job.py
n1mus: I just corrected some bugs by eye and commented it out. I didn't know what to do with it. I can delete it :sweat_smile:

src/biokbase/narrative/jobs/job.py
n1mus: That was the one in jobmanager.py that was never used. It's called in _construct_job_status when the `state` is `None`. But I think that was never the case. Maybe I should double check and then just delete this and not code hoard.

n1mus: Or should I add that error handling back in? I think the `state` was always coming from ee2, so it normally wouldn't be `None`. But am I whittling down too much

src/biokbase/narrative/jobs/job.py
ialarmedalien: I think we'll need some variant of it in the job manager to ensure that all the child jobs of a batch parent are in the running jobs list, but it's not worth keeping this code as-is. Delete!

src/biokbase/narrative/tests/test_jobmanager.py
```diff
@@ -429,8 +590,9 @@ def test_lookup_all_job_states__ignore_refresh_flag(self):
             {
                 JOB_CREATED: self.job_states[JOB_CREATED],
                 JOB_RUNNING: self.job_states[JOB_RUNNING],
-                JOB_COMPLETED: self.jm._completed_job_states[JOB_COMPLETED],
-                JOB_TERMINATED: self.jm._completed_job_states[JOB_TERMINATED],
+                JOB_COMPLETED: self.job_states[JOB_COMPLETED],
```
ialarmedalien, line 593: you could do this with a dictionary comprehension

ialarmedalien:
Looks like the tests are failing -- can you get them running?

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -661,33 +442,57 @@ def _cancel_job(self, job_id: str) -> None:
...
-                {"job_ids": job_id_list}
+                {"job_ids": checked_jobs["job_id_list"]}
             )
         except Exception as e:
             raise transform_job_exception(e)
+        # for each retry result, refresh the state of the retried and new jobs
+        orig_ids = [result["job_id"] for result in retry_results]
+        retry_ids = [result["retry_id"] for result in retry_results if "retry_id" in result]
+        for job_id in orig_ids:
+            self.get_job(job_id).reset_state()
```
ialarmedalien, line 472: I don't think we need to reset the job state here -- the goal of this bit of code is to update the job state and create the output to send to the frontend. If the job is finished (e.g. in an `error` or `terminated` state), won't this remove the cached job data?

src/biokbase/narrative/jobs/job.py
ialarmedalien: Could you call this method something like `state_for_output` -- something to indicate that this method rearranges job data into a form for transmitting it to the frontend, and edit the method docs to indicate what it's doing?

src/biokbase/narrative/jobs/job.py
ialarmedalien: It looks like this code path is for when ee2 didn't return any information about the job. It might be more appropriate to have it in the job manager module, though.

src/biokbase/narrative/jobs/jobmanager.py
ialarmedalien: It looks like this will need to be reinstated for handling error cases.

src/biokbase/narrative/jobs/job.py
```diff
@@ -240,6 +238,76 @@ def state(self):
...
+                },
+                error_code: optional - int
+            }
+        }
+        """
+        if not state:
+            state = self.state()
+        else:
+            state = self._augment_ee2_state(state)
+        state["child_jobs"] = []
```
ialarmedalien, line 278: This will need to be updated for batch jobs -- if this is a batch parent, there should be child jobs present.

src/biokbase/narrative/jobs/job.py
```diff
@@ -240,6 +238,76 @@ def state(self):
...
+        batch_size = state.get("job_input", {}).get("narrative_cell_info", {}).get("batch_size")
+        if batch_size:
+            state["batch_size"] = batch_size
```
ialarmedalien, line 281: This info has moved to the `extra_data` field; it's for the old style batch jobs -- the data is added in the `appmanager` module, but it doesn't get saved in ee2. I would be tempted just to get rid of this as it doesn't correlate with the info in `child_jobs` and caused me lots of confusion when I was working on the frontend.

src/biokbase/narrative/jobs/jobmanager.py
ialarmedalien: well, it's a call to ee2 (vs getting cached data), so it will be slower, and you can imagine a situation where ee2 does down, so it'd be useful to have that saved data.

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -79,39 +76,35 @@ def initialize_jobs(self):
...
         """
-        job_states = clients.get("execution_engine2").check_jobs(
-            {
-                "job_ids": job_ids,
-                "return_list": 0,
-                "exclude_fields": JOB_INIT_EXCLUDED_JOB_STATE_FIELDS,
-            }
-        )
+        job_ids = [job_id for job_id in job_ids if job_id not in self._running_jobs]
+        if not job_states:
```
ialarmedalien, line 91: do we need to check whether job states are present for all the jobs in `job_ids`?

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -79,39 +76,35 @@ def initialize_jobs(self):
...
-                # updates via the start_job_update message, the refresh flag will
-                # be set to True.
-                self._running_jobs[job_id] = {"refresh": 0, "job": job}
+            job_state = job_states.get(job_id, {})
+            job = Job.from_state(job_state)
+            # Note that when jobs for this narrative are initially loaded,
+            # they are set to not be refreshed. Rather, if a client requests
+            # updates via the start_job_update message, the refresh flag will
+            # be set to True.
+            self._running_jobs[job_id] = {"refresh": 0, "job": job}
```
ialarmedalien, line 105: Can this code be combined with the code in `initialise_jobs` so that we are treating new jobs consistently? Is there anywhere else where new jobs are created that this should be dealt with?

src/biokbase/narrative/jobs/job.py
```diff
@@ -240,6 +238,76 @@ def state(self):
...
+                state.update(
+                    {
+                        "status": "error",
+                        "errormsg": "Unable to build output viewer parameters!",
+                        "error": {
+                            "code": getattr(new_e, "code", -1),
+                            "source": getattr(new_e, "source", "JobManager"),
+                            "name": "App Error",
+                            "message": "Unable to build output viewer parameters",
+                            "error": "Unable to generate App output viewer!\nThe App appears to have completed successfully,\nbut we cannot construct its output viewer.\nPlease contact the developer of this App for assistance.",
```
briehl, line 299: I think we're trying to generally update to "concat the KBase team at https://kbase.us/support for assistance." or something like that.

src/biokbase/narrative/jobs/job.py
briehl: If you can verify that the state passed to `_contruct_job_status` will never be none, then just delete this, I think.

There was a case in some older version where the job lookup could fail and just return None, so that was put in to avoid something that would be catastrophic. I think that's all handled better now.

src/biokbase/narrative/jobs/job.py
briehl: Yeah, what @ialarmedalien said. This can get fixed later in the job manager.

src/biokbase/narrative/jobs/jobmanager.py
briehl: Keeping it as a `set` would be simpler, IMO. Maybe just `self._completed_jobs = <set of job ids>`?

src/biokbase/narrative/jobs/jobmanager.py
briehl: Agreed with @ialarmedalien. I'd avoid any EE2 calls if you can. You don't need to go super out of your way, though. It also depends on when `_create_jobs` gets called.

From elsewhere, it looks like `_create_jobs` only gets called during `retry_jobs`, where you don't have any job states, just the new retried ids. I'd ditch the `job_states` kwarg all together and just fetch the new state here.

This seems to have similar functionality to `initialize_jobs`, So maybe that could call this? Or both of these could call the same Job construction loop.

This also seems to return the new states. Could you make a note of that in the docstring and typing?

src/biokbase/narrative/jobs/jobmanager.py
briehl: Ha! I should read your notes before saying the same thing elsewhere... :)

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -211,221 +204,10 @@ def list_jobs(self):
...
-                try:
-                    widget_info = Job.map_viewer_params(state, params, app_id, app_tag)
-                except ValueError:
-                    widget_info = {}
-                state.update({"widget_info": widget_info})
-            child_job_states.append(state)
-        return child_job_states
-
-    def _construct_job_status_set(self, job_ids: list) -> dict:
+    def _construct_job_state_set(self, job_ids: list, states: dict = None) -> dict:
```
briehl, line 207: :+1:  on the name change

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -595,7 +376,7 @@ def cancel_job(self, job_id: str) -> bool:
             raise ValueError(f"No job present with id {job_id}")
 
         # otherwise, our job ID is fine
-        if job_id in self._completed_job_states:
+        if self.get_job(job_id).terminal_state:
```
briehl, line 379: `get_job(job_id).is_finished()` might be more appropriate here and below. Maybe that's just stylistic and I'm being picky?

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -617,12 +398,12 @@ def cancel_jobs(self, job_id_list: List[str] = []):
 
         for job_id in checked_jobs["job_id_list"]:
             if (
-                job_id not in self._completed_job_states
+                not self.get_job(job_id).terminal_state
```
briehl, line 401: Are these checks the same? Can a job be in a terminal_state and have `_check_job_terminated` be False? Or, the opposite question - can the `Job` object (the `.terminal_state` check) differ from what the `JobManager` (`_check_job_terminated`) would return?

briehl (changes requested):
Found a few little things, but overall looks good.

Mainly, it would be helpful to update typing when functions get updated (and where it's missing...).
Also, it would be nice to have `JobManager.initialize_jobs` and `.create_jobs` call the same job creation function. There's a little duplication there that may not be necessary.

src/biokbase/narrative/jobs/jobmanager.py
ialarmedalien: FYI - at some point soon-ish the `_check_job_terminated` method is going to be removed -- when cancelling a job, it first hits an ee2 endpoint that tells you whether the job has been terminated. It would be simpler just to remove that call and send a request to terminate the job. ee2 returns a 200 response if the job is no longer running, whether that be because it was successfully cancelled or because it was already finished.

src/biokbase/narrative/jobs/job.py
ialarmedalien: Could we simplify the `state()` method so that it just gets the cached state / updates the state from ee2 or a supplied state dictionary, and put all the data structure alterations into this method?

i.e.
```py
def state(self, state=None):

  if self.terminal_state:
    return self._last_state

  if not state:
    state = clients.get("execution_engine2").check_job(
      {"job_id": self.job_id, "exclude_fields": EXCLUDED_JOB_STATE_FIELDS}
    )
    except Exception as e:
  ...
  # ensures that the cached state is always up to date
  # if update_state doesn't update self.terminal_state, it should be edited so that it does
  # cache the most recent state in self._last_state
  self.update_state(state)
  return self._last_state

  def revised_state(self, state=None):
    state_for_output = self._augment_ee2_state(self.state(state))
    # etc    
```

src/biokbase/narrative/jobs/jobmanager.py
n1mus: Yeah, unclear to me. At least it's what it was doing before

src/biokbase/narrative/jobs/jobmanager.py
ialarmedalien: Here's a PR for that change: https://github.com/kbase/narrative/pull/2372

src/biokbase/narrative/jobs/job.py
n1mus: OK, so now `job.state()` is just a wrapper for the EE2 state, and `job.revised_state()` is the thing sent to the front end (more or less)

src/biokbase/narrative/jobs/jobmanager.py
n1mus: Shoot. I may have gotten confused partially because the `job.reset_state()` method was added

n1mus: @bio-boris you mentioned something about this part

ialarmedalien: You should be OK if you just run the usual `_construct_job_state_set`, because that should update the jobs. Just remove these two lines and all should be well.

Obviously you'll want to write some tests to verify that! 😉

src/biokbase/narrative/jobs/job.py
ialarmedalien: I like the idea of concat-ing the KBase team -- how far do you think we would reach? :wink:

src/biokbase/narrative/jobs/jobmanager.py
bio-boris: @n1mus I think the way we concluded was that we wouldn't be handling the state changes if they were changed by the api/backend (Like if after the cell state was set to terminated and the flag to use the cache was enabled, a retry on the backend wouldn't change that flag)

src/biokbase/narrative/jobs/job.py
briehl: Everyone! I think that was an 80's horror film...

src/biokbase/narrative/jobs/jobmanager.py
n1mus: Really? I can remove the flag to use the cache, but what it would do is retry then send the old/terminal state back to the FE. I thought I heard some mentions of retry changing the status

bio-boris: I think we were talking about if the whole cell/parent_job had a Terminal or Non Terminal state, based on whether there were jobs in certain states. So if all jobs were finished, it would be terminal. However, if someone clicked retry, it would no longer be terminal. However, if someone called the ee2 api directly, the frontend wouldn't care.

src/biokbase/narrative/jobs/jobmanager.py
n1mus: We decided that this can wait because the logic is slightly different (how `"refresh"` is initialized, the ee2 method). But there is also `register_new_job` which can be refactored in, and lines can be condensed so it looks less aesthetically like duplicate logic

briehl (approved):
Thanks for the changes.

ialarmedalien (approved):
Looks great! :tada: Thanks for all the cleanup you have done on this code -- it looks much better now.

PR #2427, opened 2021-08-19
DATAUP-551 duplicate outputs

briehl:

Check for duplicate outputs across tabs.

Screen Shot 2021-08-18 at 10 53 00 PM

This has a few fairly major changes, and **will require making a new bulk import cell.**
1. Track the output parameter ids in the bulkImportCell metadata now (just list and store them along side the fileParamIds and otherParamIds so we don't have to scrounge for them).
2. The configure tab now fetches and sends the output parameter for all of the unselected tabs to the filePathWidget of the selected tab.
3. Modifies the filePathWidget and fieldTableCellWidget to deal with the cross-tab changes.

There's some code in there to try to track the exact rows in each tab where there are duplicates, but that doesn't exactly work right because when multiple duplicates are made, they don't really get stored or flagged as such. So that code's in there, but will need some fine tuning at the level of the configuration validator to work right. I don't think we necessarily need to say what rows the duplicates are in, though, but that should get confirmed with user testing.

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/DATAUP-551

# Testing Instructions

# Dev Checklist:

### DIFF
--- a/kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
+++ b/kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
@@ -192,20 +192,39 @@ define([
             }
         }
 
+        /**
+         *
+         * @param {object} rows
+         */
         function setDuplicateValue(rows) {
             state.isDuplicate = true;
             let message = 'duplicate value';
             let rowMessage = '';
+            let tabMessage = '';
+            let tabPrefix = ' on tab';
 
-            if (rows) {
+            if (rows && rows.thisTab && rows.thisTab.length) {
                 rowMessage = ' on row';
-                if (rows.length > 1) {
-                    message += 's';
+                if (rows.thisTab.length > 1) {
                     rowMessage += 's';
                 }
-                rowMessage += ' ' + StringUtil.arrayToEnglish(rows);
+                rowMessage += ' ' + StringUtil.arrayToEnglish(rows.thisTab);
+            }
+
+            if (rows && rows.otherTabs) {
+                const tabs = Object.keys(rows.otherTabs); // don't use rows for now, as there's something funny about the reporting
+                tabMessage = StringUtil.arrayToEnglish(tabs.map((x) => `"${x}"`));
+                if (tabs.length > 1) {
+                    tabPrefix += 's';
+                }
             }
             message += ' found' + rowMessage;
+            if (rowMessage.length && tabMessage.length) {
+                message += ', and ';
+            }
+            if (tabMessage.length) {
+                message += tabPrefix + ' ' + tabMessage;
+            }
             renderMessage(MESSAGE.error, message, `.${messageBaseClass}__duplicate`);
         }
 
--- a/kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
+++ b/kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
@@ -193,7 +193,16 @@ define([
         }
 
         /**
-         *
+         * Most of the code here is wrangling the English language.
+         * This translates a structure like
+         * {
+         *   thisTab: [1, 2],
+         *   otherTabs: {
+         *     "Elsewhere": [3]
+         *   }
+         * }
+         * into text like:
+         * "duplicate values found on rows 1 and 2, and on tab "Elsewhere" row 3"
          * @param {object} rows
          */
         function setDuplicateValue(rows) {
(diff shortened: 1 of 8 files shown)


nbextensions/bulkImportCell/tabs/configure.js
```diff
@@ -116,17 +178,6 @@ define([
          * @param {DOMElement} node - the node that should be used for the left column
          */
         function buildFileTypePanel(node) {
-            const fileTypesDisplay = {},
```
briehl, line 119: Moved to global scope to make the file type display labels available to other functions.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -362,8 +368,9 @@ define([
                         inputs: typesToFiles,
                         params: initialParams,
                         app: {
-                            fileParamIds: fileParamIds,
-                            otherParamIds: otherParamIds,
+                            fileParamIds,
+                            otherParamIds,
+                            outputParamIds,
```
briehl, line 373: Track the output parameter ids for each app separately from the other types of parameter.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -250,19 +250,20 @@ define([
...
-                    return true;
-                }
                 return (
                     param.dynamic_dropdown_options &&
                     param.dynamic_dropdown_options.data_source === 'ftp_staging'
                 );
             });
+            const outputParams = appSpec.parameters.filter((param) => {
+                return param.text_options && param.text_options.is_output_name;
+            });
```
ialarmedalien, line 260: Ah good, I need this for the job status tab!

nbextensions/bulkImportCell/tabs/configure.js
```diff
@@ -65,6 +77,61 @@ define([
...
+                    outputParamIds.forEach((paramId) => {
+                        const value = allParams[type].filePaths[i][paramId];
+                        if (!value) {
+                            return;
+                        }
+                        if (!(value in unselectedOutputValues)) {
+                            unselectedOutputValues[value] = {};
+                        }
+                        const displayType = fileTypesDisplay[type].label;
+                        if (!(type in unselectedOutputValues[value])) {
```
ialarmedalien, line 125: should that be `displayType`, rather than `type`?

test/unit/spec/common/cellComponents/filePathWidget-Spec.js
```diff
@@ -68,7 +69,7 @@ define([
         });
 
         beforeEach(function () {
-            this.bus = Runtime.make().bus();
+            this.bus = Bus.make();
```
ialarmedalien, line 72: is this a more general switch that could be made?

kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
```diff
@@ -192,20 +192,46 @@ define([
...
+                if (rows.thisTab.length > 1) {
                     rowMessage += 's';
+                    plural = true;
+                }
+                rowMessage += ' ' + StringUtil.arrayToEnglish(rows.thisTab);
+            }
+
+            if (rows && rows.otherTabs) {
+                const tabs = Object.keys(rows.otherTabs);
+                // don't use rows for now, as there's something funny about the reporting
```
ialarmedalien, line 219: were you able to reproduce the strangeness in a test?

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
```diff
@@ -704,51 +726,76 @@ define([
...
-            // }
+            /* Get the parameter widgets and translate to something easier to work with -
+             * {
+             *    rowId: {
+             *      paramId1: {
+             *        widget: Object
+             *        duplicateRows: {
+             *          thisTab: [rowId, rowId],
+             *          otherTabs: {
+             *              tabId1: [rowIdx, rowIdx]
```
ialarmedalien, line 767: is it deliberate that `thisTab` uses `rowId` and `otherTabs` uses `rowIdx`?

kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
briehl: No, but I think it stems from how and when validation messages get sent, or not. And if they get sent with some invalid content, they shouldn't get stored in the data model. Some seem to seep through, though.

I'll put a TODO in to try to track it down, but I think that's going to be a deep, dark, rabbit hole of bus message mayhem.

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
briehl: Yeah, that's a typo. At that point, they should all numerical row index values, not the unique ids.

nbextensions/bulkImportCell/bulkImportCell.js
briehl: Good, I'm glad it's useful outside of this part!

nbextensions/bulkImportCell/tabs/configure.js
briehl: Yes! Oh, I wonder if that's part of the issue with rows not being right? That would overwrite this array each time it's made.

test/unit/spec/common/cellComponents/filePathWidget-Spec.js
briehl: Yeah, I realized it late last night. This could probably be made in a number of other places.

kbase-extension/static/kbase/js/common/cellComponents/fieldTableCellWidget.js
briehl: Fixed it! Your comment on the configure tab was the bug. 🎉

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
```diff
@@ -183,7 +180,11 @@ define([
             fieldWidget.bus.on('validation', (message) => {
                 // only propagate if invalid. value changes come through
                 // the 'changed' message
-                if (!message.isValid) {
+                let duplicateValues = [];
+                if (model.getItem('outputParamIds').includes(parameterSpec.id)) {
```
briehl, line 184: This is duplicated-ish from the `changed` response.
It turns out that output parameter widgets send both a "changed" and a "validation" response when changes happen. The changes are intended to be stored whether they're an error or not. Validation is to say whether that's valid or not.

The problem comes at widget startup - both of those messages are fired more or less at the same time. So catching one to do the duplication check could get overridden by the other. It's a little annoying, but clean enough, to do the check twice.

Another solution would probably be to either refactor the input widget itself (which would mean an overhaul of all the input widget behaviors), or to cache what values are actually duplicates somewhere and try to keep that updated as we go... which is also a ton more code to fix what I think is a message-spam issue.

ialarmedalien (approved):
Looks great! Let's get it merged.

PR #2431, opened 2021-08-26
DATAUP-468 verify files on load 

briehl:

The goal is to verify that staging area files still exist when loading a bulk import cell. It should show errors and prevent app starting if they're gone.

This is part 1 of a multi-part PR. There's some progress here, but some cases won't be handled yet until the other PRs are done. I just wanted to keep them smallish.

This PR does the following:
1. Fixes select2 styling to show an error when something bad is selected.
2. Adds an option to select boxes of invalid selections on start. Selecting these will force the error to happen.
3. Disables a (I think?) redundant parameter check on start. The BI cell was doing a global parameter check whenever it was started and used that to figure out what state it should be in. However, that state should be saved already, so this might be redundant.

Next PR will (hopefully) speed things up by making a front end cache for staging file information. It can get sluggish with multiple bulk import cells with lots of file data to fetch. Maybe it's worth holding off, though, and optimizing later.

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/DATAUP-468

# Testing Instructions
1. Make a Bulk Import cell with some files
2. Get those files selected
3. Save the Narrative (probably easiest to do that at this step)
4. Delete one of the files.
5a. Reset the configure tab (i.e. disable and re-enable the tab, or hit the info tab and go back, or toggle to a different data type and go back) and see the error.
5b. Reload the narrative to see the error.

# Dev Checklist:


kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
```diff
@@ -30,18 +30,20 @@ define([
      *  - initialParams - Array of objects. Each item represents a row of parameter values.
      *    So each item has an object with parameter id -> value
      *  - availableFiles - Array of strings, one for each file that should be available here.
+     *  - unavailableFiles - Set of strings, for files that the cell thinks are available, but
+     *      are no longer available through the staging service
      *  - viewOnly - boolean, if true start with the view version of input widgets
+     *  - unselectedOutputValues
```
ialarmedalien, line 36: Presumably this is in an upcoming PR?

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -602,54 +602,58 @@ define([
...
-                            // Update the execMessage panel with details of the active jobs
-                            controlPanel.setExecMessage(
-                                Jobs.createCombinedJobState(
-                                    jobManagerContext.model.getItem('exec.jobs.byStatus')
-                                )
-                            );
-                        },
-                        fsmState: (jobManagerContext) => {
-                            const fsmState = Jobs.getFsmStateFromJobs(
+            render().then(() => {
```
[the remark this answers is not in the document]
ialarmedalien: 😖

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -602,54 +598,34 @@ define([
...
+                        }
+                    },
+                });
 
-                    return BulkImportUtil.evaluateConfigReadyState(model, specs);
-                })
-                .then((appReadyState) => {
-                    const curState = model.getItem('state');
-                    const curReadyState = curState.params;
-                    const updatedReadyState = !SimpleUtil.isEqual(appReadyState, curReadyState);
```
ialarmedalien, line 635: so most of this is now no longer necessary?

ialarmedalien (approved):
👍

ialarmedalien:
There are a couple of Sonarcloud issues that should be quick and easy to fix if you want to do that prior to merging.

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
briehl: Oops, no, that should get removed.

nbextensions/bulkImportCell/bulkImportCell.js
briehl: Correct, at least I think so. 
That block runs on cell startup (and is only really relevant after a page reload) and does a check over the app inputs to see what state they're in. This is fine, except for two things:

1. Loading the configure tab would do it again.
2. It's completely irrelevant if the cell is being run, or has been run.

So having this here is just redundant at this point. The other problem is that it would require a good chunk of rearranging where information goes and where it comes from so that this could get run. The configure tab has all of the information already at hand, and it would take some data munging (and running a `list` command against the staging service) to make it work here.

If it becomes a problem, I think we should move that check out entirely to a utility function that would just take in the cell's data model and figure it out, but I would worry about efficiency.

kbase-extension/static/kbase/js/common/cellComponents/filePathWidget.js
briehl: Or, rather, filled out. :V

briehl:
I think I get why sonarcloud is so wonky about code complexity - it treats all of our more factory modules as a giant function. So if you have a bunch of little functions in there, each with nesting logic, it counts all of the logical bits as being part of one big function. Which isn't helpful.

I went through and fixes the other ones, though.

ialarmedalien:
Yeah, I'm afraid I usually ignore the code complexity issues as there isn't much we can do without a wholesale move to es6 imports/exports.

PR #2463, opened 2021-09-14
DATAUP-468 verify files on load

briehl:

Addresses a few bugs with verifying that files still exist when reloading a bulk import cell, and introduces a bit of inefficiency that'll get fixed later.

1. On page load (and on configure tab load), it was only searching for files in the currently selected data type tab. Now it should search for files on all tabs.
2. When this was only done for the configure tab, it was causing an error if a different tab was selected, and it was causing errors with managing bulk import cell state. So now a check is done when the bulk import cell itself is loaded.
3. Inefficiency - this means that all tabs are effectively checked twice if the configure tab is the first tab loaded. This isn't too much of a problem if (a) the user doesn't have many files in the staging area so they load up quick or (b) the configure tab isn't pre-selected. But an efficiency bump would be gained if we could cache the call for several minutes (and invalidate it when the user uploads more files). A later PR will unify calls to `StagingService.list` under a data provider front end cache, including calls from the Staging Area.

So this should fix the bugs, but might make things a little slower if there's 300+ files in the staging service (like i'm testing with).

# Jira Ticket / Issue #
https://kbase-jira.atlassian.net/browse/DATAUP-468

# Testing Instructions
* Make a bulk import cell with some files of multiple types.
* save the narrative
* delete 1 or more files from the staging area for each of those types
* reload the narrative
* should see some errors if those files were selected for import

# Dev Checklist:


nbextensions/appCell2/widgets/appCellWidget.js
```diff
@@ -1131,7 +1131,11 @@ define([
             }
             const state = fsm.getCurrentState();
             try {
-                FSMBar.showFsmBar({ ui: ui, state: state, job: model.getItem('exec.jobState') });
+                FSMBar.showFsmBar({
```
briehl, line 1134: I guess Prettier and a stray git commit snuck something in here.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -608,37 +612,69 @@ define([
             const meta = cell.metadata;
             meta.kbase.attributes.lastLoaded = new Date().toUTCString();
             cell.metadata = meta;
-            render().then(() => {
-                jobManager.addHandler('modelUpdate', {
-                    execMessage: (jobManagerContext) => {
-                        // Update the execMessage panel with details of the active jobs
-                        controlPanel.setExecMessage(
-                            Jobs.createCombinedJobState(
+            render()
```
briehl, line 615: Turns out this chunk where we render, then do a global parameter check to update the state with is actually usefully important. Updated this to revert that change, merge it with other recent changes, and introduce the flow that looks up missing files before validating that things are as correct as they should be.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -608,37 +612,69 @@ define([
...
-                });
-
-                // TODO: assess cell state, update job info if required
-                try {
-                    jobManager.restoreFromSaved();
-                } catch (e) {
-                    console.log(e);
-                }
+                    });
+                    return BulkImportUtil.getMissingFiles(Array.from(expectedFiles)).catch(
```
briehl, line 646: Added the `.catch` to this chunk of the Promise (rather than the next part of the Promise chain) so that it'll more easily pass on a silent empty Array if it fails.

I figure that if the staging service is down, bulk import cells should still be able to load their last known good state.

nbextensions/bulkImportCell/bulkImportCell.js
```diff
@@ -608,37 +612,69 @@ define([
...
+                                error
+                            );
+                        }
+                    );
+                })
+                .then((missingFiles = []) =>
+                    BulkImportUtil.evaluateConfigReadyState(model, specs, new Set(missingFiles))
+                )
+                .then((readyState) => {
+                    const curState = model.getItem('state');
```
briehl, line 660: This is mostly restored from the previous code state.

nbextensions/bulkImportCell/util.js
```diff
@@ -53,24 +53,88 @@ define(['common/runtime', 'StagingServiceClient'], (Runtime, StagingServiceClien
...
      * the ready state.
+     * @param {Object} model the data model from the BulkImportCell, containing all the cell info
+     * @param {Object} specs keys = app id, values = app specs
+     * @param {Set} missingFiles Set of files that the app cell expects to be present,
+     *  but are missing from the staging area
+     * @returns
      */
-    function evaluateConfigReadyState(model, specs) {
+    function evaluateConfigReadyState(model, specs, missingFiles) {
+        // given some missing files (precalcuate for now), set up options for evaluating all file inputs
```
briehl, line 126: Precalculate "for now" = it would be nice to have that optional. If left as `undefined`, we should look that up from a cache, or just make the call based on the list of input files in the `model` object.

However, it's only actually used in two places (the `configure` and `bulkImportCell` modules), and it seemed cleaner to just make both of those places look up their own missing files for now. The `Configure` module needs to keep a hold of that list of missing files to compare while the user is setting up inputs.

nbextensions/appCell2/widgets/appCellWidget.js
ialarmedalien: I sometimes wonder if Prettier changes its mind between runs.

nbextensions/bulkImportCell/util.js
```diff
@@ -53,24 +53,88 @@ define(['common/runtime', 'StagingServiceClient'], (Runtime, StagingServiceClien
...
         const evalPromises = fileTypes.map((fileType) => {
-            const filePathValues = model.getItem(['params', fileType, 'filePaths']);
             // make an array of empty options with the same length as the
             // number of file path values
-            const filePathOptions = Array.from({ length: filePathValues.length }, () => ({}));
             return evaluateAppConfig(
                 model.getItem(['app', 'otherParamIds', fileType]),
                 model.getItem(['params', fileType, 'params']),
-                {},
+                {}, // no particular options for non-file parameters righ tnow
```
ialarmedalien, line 134: misplaced space

ialarmedalien (approved):
Looks fine -- there's one misplaced space in a comment which you can fix here if you want to or put it into another PR.

briehl:
I'll get it in the next one. Thanks!

PR #2477, opened 2021-09-24
test version bump deploy

briehl:
## DO NOT MERGE

This is just here to test deployment of different version tags on CI.


briehl:
I think this needs to actually be merged in order for the "test" to work. During the PR it only makes an image with a PR tag, it doesn't make a version release image until it's merged.

PR #2651, opened 2021-12-14
DATAUP-608 add uploader docs

briehl:

Address some clarifying comments in the previous PR #2605 to add to the uploader app configuration documentation.

No code changes.

### DIFF
--- a/docs/developer/data_import.md
+++ b/docs/developer/data_import.md
@@ -29,7 +29,7 @@ The data import app selection is configured in [`staging_upload.json`](../../kba
 
 Each `app_info` object has the following accepted keys:
 * `app_id`: the app to use for importing the file in the format `module_name/app_name` (e.g. `kb_uploadmethods/import_genbank_as_genome_from_staging`). This is the only required field.
-* `app_input_param`: this is the parameter for setting the file name in the app.
+* `app_input_param`: this is the parameter for setting the file name in the app. App specs don't provide an unambiguous parameter for a file input, hence the need for this clarifying element.
 * `app_input_param_type`: describes what the `app_input_param` is expected to be. Accepts either `string` or `list` (list of strings). If the given app spec has `allow_multiple` as an option for the input, then this should be of type `list`.
 * `app_static_params`: an object with key-value pairs of other app parameters that should be preset when creating the app cell.
 * `app_output_param`: this should be the app's parameter id for the created object. This gets automatically filled with a suggested name based on the file name and the `app_output_suffix` field below. Note that this does not take into account whether the parameter has the `is_output_name` option set.
--- a/docs/developer/data_import.md
+++ b/docs/developer/data_import.md
@@ -29,12 +29,12 @@ The data import app selection is configured in [`staging_upload.json`](../../kba
 
 Each `app_info` object has the following accepted keys:
 * `app_id`: the app to use for importing the file in the format `module_name/app_name` (e.g. `kb_uploadmethods/import_genbank_as_genome_from_staging`). This is the only required field.
-* `app_input_param`: this is the parameter for setting the file name in the app
-* `app_input_param_type`: describes what the `app_input_param` is expected to be. Accepts either `string` or `list` (list of strings)
+* `app_input_param`: this is the parameter for setting the file name in the app.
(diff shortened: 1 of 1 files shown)


docs/developer/data_import.md
```diff
@@ -29,12 +29,12 @@ The data import app selection is configured in [`staging_upload.json`](../../kba
 
 Each `app_info` object has the following accepted keys:
 * `app_id`: the app to use for importing the file in the format `module_name/app_name` (e.g. `kb_uploadmethods/import_genbank_as_genome_from_staging`). This is the only required field.
-* `app_input_param`: this is the parameter for setting the file name in the app
-* `app_input_param_type`: describes what the `app_input_param` is expected to be. Accepts either `string` or `list` (list of strings)
+* `app_input_param`: this is the parameter for setting the file name in the app.
```
MrCreosote, line 32: Can you update as per https://github.com/kbase/narrative/pull/2605#discussion_r768944647

MrCreosote (changes requested):
So I realize this is just documenting the way things work, which is good, but I'm not sure when an opportunity to discuss this mechanism will pop up again so here goes.

There's a couple aspects to how this work that naively appear concerning to me regarding how apps get updated, assuming I understand the process correctly. When a dev updates and registers an app, it may or may not need spec changes here for the app to work correctly. That means there may be some period of time where the dev, beta, and / or release apps don't work, depending on when the narrative spec is updated, and to have the release app work correctly the entirety of the time the narrative with the spec change needs to be released at the same time as the app.
Am I thinking about this the right way or is the sequence of operations in my head incorrect?

docs/developer/data_import.md
briehl: I commented on that PR, but I'll just copy it here - that doesn't "override" the spec, because there isn't a specific "THIS IS THE INPUT FILE" field there. It is, by the nature of the app specs, possibly ambiguous.

briehl:
> There's a couple aspects to how this work that naively appear concerning to me regarding how apps get updated, assuming I understand the process correctly. When a dev updates and registers an app, it may or may not need spec changes here for the app to work correctly. That means there may be some period of time where the dev, beta, and / or release apps don't work, depending on when the narrative spec is updated. 

That's correct. It's a bit of a pain, but with the ambiguous nature of the app specs, it's what we've got. Another option would be to make dev/beta/release versions of the config, or some kind of flag if the individual uploaders differ.

But, yeah, this isn't the best place to think about or redesign that. A JIRA ticket would save the conversation better.

docs/developer/data_import.md
MrCreosote: Okay, how about adding some text along those lines saying that the app doesn't provide an unambiguous input file marker hence the need for this field?

MrCreosote:
> That's correct. It's a bit of a pain, but with the ambiguous nature of the app specs, it's what we've got.

To me, naively it seems like the right thing to do is to make the app specs unambiguous and have the entire specification for the app behavior and rendering in one place, so we're not forced into a situation where there are split, possibly conflicting specs in different code bases that need to be updated in lockstep to avoid Weird Stuff (TM). Anyway, I'll make a ticket which I presume we won't be getting to any time soon

MrCreosote:
https://kbase-jira.atlassian.net/browse/DATAUP-671

docs/developer/data_import.md
briehl: Ok, done.

briehl:
> To me, naively it seems like the right thing to do is to make the app specs unambiguous and have the entire specification for the app behavior and rendering in one place...

Yeah, I've been kinda hoping for that for years. Maybe in the magical distant future when we get to redesign the SDK and apps?

MrCreosote:
> Maybe in the magical distant future when we get to redesign the SDK and apps?

It doesn't seem *that* hard:
* update the NMS with any needed fields like `input_file: true` or whatever
* update the Narrative to understand those fields and prefer them over the current internal specs
* update `spec.json`s and deploy apps as needed
* drop the specs and spec code

There's only a handful of apps, so it's not like CoaC or anything

MrCreosote (approved):
LGTM

PR #2723, opened 2022-01-26
Remove "import specification" from the staging area options

briehl:

As the title says - remove the option to load an XSV file as an import specification. This lets us keep working in the codebase  on `develop` but prevents unfinished features from leaking out to the upcoming 5.0.1 release (and others after that).


ialarmedalien (approved):
merge away!

PR #2881, opened 2022-04-11
Version bump to 5.0.3

briehl:

See release notes for changes.
The only other change here is removing some never-used code from `src/biokbase/narrative/__init__.py`.


src/biokbase/narrative/__init__.py
```diff
@@ -13,22 +13,4 @@ def version():
 #   no args = print current version
 #   1 arg = self-modify version to arg value
 if __name__ == "__main__":
-    import os
```
briehl, line 16: This is some fancy self-updating thing that never gets used. We might consider a helper script that helps out releases. But not yet.

ialarmedalien (approved):
LGTM

PR #2882, opened 2022-04-11
Updating javascript dependencies

ialarmedalien:

Implementing most of the most recent batch of dependabot updates.


package.json
```diff
@@ -129,21 +129,21 @@
...
-        "datatables.net": "1.11.3",
+        "datatables.net": "^1.11.5",
         "datatables.net-bs": "^1.11.5",
         "datatables.net-buttons-bs": "1.4.2",
         "dropzone": "5.7.0",
         "file-saver": "1.3.4",
         "font-awesome": "4.7.0",
         "google-code-prettify": "1.0.5",
-        "handlebars": "4.0.5",
+        "handlebars": "^4.7.7",
```
ialarmedalien, line 141: read through the release notes and checked in the UI

package.json
```diff
@@ -129,21 +129,21 @@
...
         "jquery": "2.2.4",
         "jquery-nearest": "1.3.1",
-        "jquery-ui": "1.12.1",
+        "jquery-ui": "^1.13.1",
         "jquery.tipsy": "1.0.3",
-        "js-yaml": "3.3.1",
+        "js-yaml": "3.14.1",
```
ialarmedalien, line 146: We use this module for the `requirejs-yaml` plugin (now archived/unmaintained); `js-yaml` is now on v4.x, but `requirejs-yaml` uses the 3.x API, so we cannot update any further

briehl: 👍

briehl (changes requested):
I think just add the list of updates to the release notes and it's good to go!

briehl (approved):
Thanks!

PR #2951, opened 2022-05-05
DATAUP-754 Dependabot, python

n1mus:

Did recommended upgrades for:
* `idna`
* `pyopenssl`
* `pymongo`
* `jupyter-console`
* `beautifulsoup4`
* `plotly`
* `pygments`
* `html5lib`

Did intermediate version upgrade instead for:
* `cryptography` (in order to satisfy `pyopenssl` upgrade but avoid python3.6 support drop)

Rejected recommended upgrades for:
* `jinja2` (and actually downgraded)
* `pillow`
* `pytest` (so negating a recent dependabot PR rebase)

Additional non-dependabot downgrades:
* `jsonschema` (just the start of a rabbit hole)

# Jira Ticket / Issue #

# Testing Instructions

# Dev Checklist:


ialarmedalien:
It looks like something strange is going on in the docker image wrt: the backend tests -- have a look at the logs for the [`run_unit_tests`](https://github.com/kbase/narrative/runs/6302254595?check_suite_focus=true) backend tests step. Unfortunately `pip` has some quirks and when it installs a list of modules, it doesn't always catch all the dependencies of the listed modules, plus whether or not the deps get installed can be dependent on where in the `requirements.txt` file a module is listed (😖 ).

Previously we had a similar issue with `pytest` not being able to find one of its component modules which was fixed by moving `pytest` to the end of the file. It may be that you just need to add `pluggy` to the `requirements.txt` file.

ialarmedalien:
It looks like something strange is going on in the docker image wrt: the backend tests -- have a look at the logs for the [`run_unit_tests`](https://github.com/kbase/narrative/runs/6302254595?check_suite_focus=true) backend tests step. Unfortunately `pip` has some quirks and when it installs a list of modules, it doesn't always catch all the dependencies of the listed modules, plus whether or not the deps get installed can be dependent on where in the `requirements.txt` file a module is listed (😖 ).

Previously we had a similar issue with `pytest` not being able to find one of its component modules which was fixed by moving `pytest` to the end of the file. It may be that you just need to add `pluggy` to the `requirements.txt` file.

briehl:
That's silly. Maybe the version of `pluggy` gets overwritten by something else? I ran things in my local env, and it seemed to test ok, I'll try in a fresh env.

ialarmedalien:
looks like there were a couple of issues:
https://github.com/kbase/narrative/runs/6302185929?check_suite_focus=true#step:7:1099
https://github.com/kbase/narrative/runs/6302185929?check_suite_focus=true#step:7:1798

so I'd revert those two deps back to the old version and see if that changes anything.

If you also rebase on develop, there's a test fix which should get the FE unit tests passing again

ialarmedalien:
redo the current commits so that commit 1 updates the release notes with the updates from dependabot, and then, in a separate commit, see if you can add `idna`. If that works, update `pyopenssl`, and after that, `pymongo`. If it doesn't, try updating `pyopenssl` or `pymongo`. If none of that works, we can just merge the first commit with the updated release notes.

n1mus:
I did not realize that the narrative still runs on python3.6. It was hard to tell from the Dockerfile and I guess it's not common to have narrative containers around since the base image is so large.

briehl:
We really need to update the base image. I think maybe we should start one from scratch. I'm not sure that there are things in the KBase base image that are necessary. Though there might be a new one these days.

ialarmedalien:
ISTR that digging into the narrative image and following the trail of the various docker images that it is based on takes you down through all nine circles of hell.

briehl:
Yup, been there. I think we could safely put it off for now.
Well, "safely" anyway.
Until something dire breaks and no longer installs.

PR #2952, opened 2022-05-05
DATAUP-639 add "exact_match_on" support to dynamic dropdown

briehl:

When `exact_match_on` is given to an app spec's `dynamic_dropdown_options` map, it's intended to do the following:
1. Automatically do a dynamic search using the `initialValue` as search term
2. The results will be a list of objects. Search out the first one who's `exact_match_on` key exactly matches the value (case insensitive).
3. Use that to populate the field and get the value out of it.

For example, if we're doing the scientific name search (which is the first case this is supposed to work for), and the given initialValue is "Escherichia coli K-12", then when this dropdown initializes, it starts by doing a search with "Escherichia coli K-12" as the input. When it gets a response, it looks for the EXACT match, of the term given by the `exact_match_on` field, case insensitive. In this case, it'll find something like:

```
{
  scientific_name: "Escherichia coli K-12",
  id: "83333",
  ncbi_taxon_id: "83333",
  .... other stuff ....
}
```

Our `exact_match_on` field is `scientific_name`, so match against that. This is the hit, so autoselect the option with that result. it then does the usual bit of validation and processing - the real "value" returned by this field is the taxon id, so 83333 gets returned.

If nothing is found, this becomes invalid and shows an error. A user can fix it by doing a new search, or just clearing the field.

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/DATAUP-639

# Testing Instructions

# Dev Checklist:

- [x] [Version has been bumped](https://semver.org/) for each release

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
+++ b/kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
@@ -49,6 +49,7 @@ define([
                 displayValue: undefined,
                 value: undefined,
                 descriptionFields: new Set(),
+                exactMatchError: false,
             };
         let parent, container, ui;
         if (typeof ddOptions.query_on_empty_input === 'undefined') {
@@ -170,6 +171,9 @@ define([
 
         function validate() {
             return Promise.try(() => {
+                if (model.exactMatchError) {
+                    return Validation.validateFalse(model.value);
+                }
                 let selectedItem = getControlValue();
                 const validationConstraints = {
                     min_length: spec.data.constraints.min_length,
@@ -324,6 +328,7 @@ define([
             };
 
             if (result.isValid) {
+                model.exactMatchError = false;
                 const newValue =
                     result.parsedValue === undefined ? result.value : result.parsedValue;
                 model.value = newValue;
@@ -337,20 +342,21 @@ define([
                 }
                 changeMsg.newDisplayValue = model.displayValue;
                 changeMsg.newValue = newValue;
+                channel.emit('changed', changeMsg);
             } else if (result.diagnosis === Constants.DIAGNOSIS.REQUIRED_MISSING) {
                 model.value = spec.data.nullValue;
+                channel.emit('changed', changeMsg);
+            } else if (result.diagnosis === Constants.DIAGNOSIS.INVALID && model.exactMatchError) {
+                result.errorMessage = `An exact match was not found for "${model.value}". Please search again.`;
             }
-            channel.emit('changed', changeMsg);
-            channel.emit('validation', {
-                errorMessage: result.errorMessage,
-                diagnosis: result.diagnosis,
-            });
+            channel.emit('validation', result);
         }
 
         /**
          * Clears the current selection and updates the model.
          */
         function doClear() {
+            model.exactMatchError = false;
             $(ui.getElement('input-container.input')).html('');
             model.value = spec.data.nullValue;
             channel.emit('changed', {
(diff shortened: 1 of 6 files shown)


kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -488,83 +494,102 @@ define([
...
-                            viewData.text = config.initialDisplayValue;
-                        } else if (
-                            config.initialDisplayValue !== null &&
-                            typeof config.initialDisplayValue === 'object'
-                        ) {
-                            viewData = Object.assign(viewData, config.initialDisplayValue);
-                            if (!viewData.text) {
-                                viewData.text = viewData.id;
-                            }
+        async function render() {
```
briehl, line 497: Turned this function into `async` which tweaked the formatting. I'll point out the actual changes below.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -488,83 +494,102 @@ define([
...
+                        break;
+                    }
+                }
+                if (match) {
+                    data.push(Object.assign(match, { selected: true }));
+                    model.value = match.id;
+                } else {
+                    model.exactMatchError = true;
+                }
+            }
```
briehl, line 543: Here's the changed block.

Basically, if we have an initial value, we DON'T have a display value yet (which would get populated by the dropdown), and there's an `exact_match_on` field, then we need to do the search. Just use the existing mechanism for that.

If we find a matching result (I'm guessing most of the time there will just be one hit if so), then add the `selected` boolean to it so Select2 knows what to do. Otherwise, trigger the error condition.
This makes it pretty specific to creating a new bulk import cell, as only the uploaders make use of that right now.

If a match gets found,

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -488,83 +494,102 @@ define([
...
+
+            dropdown.select2({
+                allowClear: true,
+                disabled: config.isViewOnly,
+                templateResult: formatObjectDisplay,
+                templateSelection: selectionTemplate,
+                data,
```
briehl, line 550: Turns out we can seed select2 with all the intro data we want, and the AJAX command will just updated it (with the `processResults` callback).

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -170,6 +171,11 @@ define([
 
         function validate() {
             return Promise.try(() => {
+                if (model.exactMatchError) {
```
briehl, line 174: If we're in a state where we have an exactMatchError, then force the validation to fail and put in a custom, helpful (hopefully?) error message.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -337,20 +344,19 @@ define([
...
             } else if (result.diagnosis === Constants.DIAGNOSIS.REQUIRED_MISSING) {
                 model.value = spec.data.nullValue;
+                channel.emit('changed', changeMsg);
             }
-            channel.emit('changed', changeMsg);
-            channel.emit('validation', {
-                errorMessage: result.errorMessage,
-                diagnosis: result.diagnosis,
-            });
+            channel.emit('validation', result);
```
briehl, line 352: I don't much like the duplicate `channel.emit('changed')` bit, but it shouldn't emit `changed` when there's an error. IMO it's clean enough, but I bet I'm missing something obvious.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -488,83 +494,102 @@ define([
...
-                                    failure(err);
-                                });
-                        },
+                        }
+                        return data;
                     },
-                    placeholder: {
-                        id: 'select an option',
+                    transport: function (params, success, failure) {
+                        model.exactMatchError = false;
```
briehl, line 570: Whenever a user triggers a search, we should clear the error, as any result they select should be valid now.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
```diff
@@ -488,83 +494,102 @@ define([
...
-                                    dropdown.trigger('change');
+                data.push(viewData);
+            } else if (model.value && !model.displayValue && ddOptions.exact_match_on) {
+                const searchResults = await fetchData(model.value);
+                // verify that we have an exact match
+                let match = null;
+                for (const result of searchResults) {
+                    if (
+                        model.value.toLowerCase() === result[ddOptions.exact_match_on].toLowerCase()
+                    ) {
```
ialarmedalien, line 532: do we need to do any of that tedious stuff like trimming off whitespace or ensuring that there's some content in the match string, or is that done earlier on?

ialarmedalien (approved):
LGTM!

kbase-extension/static/kbase/js/widgets/appWidgets2/input/dynamicDropdownInput.js
briehl: Yes, certainly. Don't know how I missed that. I'm just making silly assumptions that the database call probably won't honor.

PR #2960, opened 2022-05-09
job ts implementation

n1mus:

* Please include a summary of the change and which issue is fixed.
* Please also include relevant motivation and context.
* List any dependencies that are required for this change.

# Jira Ticket / Issue #

- [ ] Added the Jira Ticket to the title of the PR (e.g. `DATAUP-69 Adds a PR template`)

# Testing Instructions

# Dev Checklist:


src/biokbase/narrative/jobs/job.py
```diff
@@ -240,16 +267,11 @@ def __getattr__(self, name):
...
-            self._acc_state["job_input"] = self._acc_state.get("job_input", {})
-            self._acc_state["job_input"][name] = value
+            self._update_state({"job_input": {name: value}})
         elif name in NARR_CELL_INFO_ATTRS:
-            self._acc_state["job_input"] = self._acc_state.get("job_input", {})
-            self._acc_state["job_input"]["narrative_cell_info"] = self._acc_state[
-                "job_input"
-            ].get("narrative_cell_info", {})
-            self._acc_state["job_input"]["narrative_cell_info"][name] = value
+            self._update_state({"job_input": {"narrative_cell_info": {name: value}}})
```
n1mus, line 274: Doesn't trigger an output_state update but hopefully wouldn't spam output_state updates?

src/biokbase/narrative/jobs/job.py
```diff
@@ -341,16 +364,18 @@ def _update_state(self, state: dict) -> None:
...
         if self._acc_state is None:
             self._acc_state = state
         else:
-            self._acc_state.update(state)
+            merge_inplace(self._acc_state, state)
+
+        self.last_updated = time.time_ns()
 
-    def state(self, force_refresh=False, exclude=JOB_INIT_EXCLUDED_JOB_STATE_FIELDS):
+    def refresh_state(self, force_refresh=False, exclude=JOB_INIT_EXCLUDED_JOB_STATE_FIELDS):
```
ialarmedalien, line 378: I would think that `refresh_state` would only be called if `force_refresh` is true -- otherwise the caller would use `static_state`

src/biokbase/narrative/jobs/job.py
```diff
@@ -359,9 +384,9 @@ def state(self, force_refresh=False, exclude=JOB_INIT_EXCLUDED_JOB_STATE_FIELDS)
             state = self.query_ee2_state(self.job_id, init=False)
             self._update_state(state)
 
-        return self._internal_state(exclude)
+        return self.static_state(exclude)
 
-    def _internal_state(self, exclude=None):
+    def static_state(self, exclude=None):
```
ialarmedalien, line 389: maybe change this to `current_state` or `cached_state` to better reflect what is being returned?

src/biokbase/narrative/jobs/job.py
```diff
@@ -199,7 +226,7 @@ def __getattr__(self, name):
                 # and need the state refresh.
                 # But KBParallel/KB Batch App jobs may not have the
                 # batch_job field
-                self.state(force_refresh=True).get(
+                self.refresh_state(force_refresh=True).get(
```
ialarmedalien, line 229: I am wondering if it would be better to get rid of the "magic" in these attributes, so that they just return the current value, rather than automatically updating. Explicitly requesting that jobs be updated fits better with "the principle of least surprise", and it would also make the code simpler because you don't have to worry about getting into some loop of endless updates. The only thing is that you'd need to take a look at where these various attributes have been called and ensure that there's an appropriate job refresh called.

src/biokbase/narrative/jobs/job.py
```diff
@@ -274,11 +296,11 @@ def was_terminal(self):
             return self._acc_state.get("status") in TERMINAL_STATUSES
 
     def is_terminal(self):
-        self.state()
+        self.refresh_state()
```
ialarmedalien, line 299: I don't think this function (`is_terminal`) is called anywhere apart from in tests. There doesn't seem to be much point in keeping it around.

src/biokbase/narrative/jobs/job.py
```diff
@@ -199,7 +226,7 @@ def __getattr__(self, name):
                 # and need the state refresh.
```
ialarmedalien, line 226: Comment about the methods `query_ee2_state` and `query_ee2_states` -- I don't find it very intuitive that they're part of the job object, given that one deals with multiple jobs. I think it'd be better to have them in the Job Manager, along with the other ee2 job status query that fetches jobs by workspace ID.

src/biokbase/narrative/jobs/job.py
n1mus: Maybe you're right. It's ancient logic though and maybe in the scope of another, less urgent PR?

src/biokbase/narrative/jobs/job.py
n1mus: Sensible

src/biokbase/narrative/jobs/job.py
n1mus: I actually kind of agree with what you're arguing here. But I took a look at changing it and it looks like `Job` has invocations like `cls.query_ee2_states` but `Job` doesn't know about `JobManager` so I think we should just keep it as is for now

src/biokbase/narrative/jobs/job.py
n1mus: Maybe call it `latest_state` or something ...

src/biokbase/narrative/jobs/job.py
n1mus: Though I want to keep this a tiny PR!

src/biokbase/narrative/jobs/job.py
n1mus: I guess you could also move the invoking method, `from_job_ids`, to `JobManager`, but I kind of like it (because I wrote it hehe ... a class factory method). But I'd argue not in scope of this ticket!

n1mus: But the singular `query_ee2_state` is also invoked in a `Job` instance method
I guess I have a weird habit of putting meta-instance methods all as class methods into that class ... Like I might merge `JobManager` into `Job` if it weren't so ginormous. I wonder if that's an anti-pattern

src/biokbase/narrative/jobs/job.py
ialarmedalien: Can you take a look and see how much it would take to remove the "magic" attributes? There are only a few that do trigger updates, and we can also fix the lack of a `status` attribute at the same time. It should be easy enough just to make a separate PR with this content.

src/biokbase/narrative/jobs/job.py
```diff
@@ -341,16 +330,18 @@ def _update_state(self, state: dict) -> None:
         # Check if there would be no change in updating
         # i.e., if state <= self._acc_state
         if self._acc_state is not None:
-            if {**self._acc_state, **state} == self._acc_state:
+            if merge(self._acc_state, state) == self._acc_state:
```
ialarmedalien, line 333: why merge instead of just checking whether `{**self._acc_state, **state}` would be the same as `self._acc_state`?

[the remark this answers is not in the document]
ialarmedalien: It might be an idea to remove the `updated` field from the job state data as it's not accurate and the FE doesn't use it.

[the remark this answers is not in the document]
ialarmedalien: It might be an idea to remove the `updated` field from the job state data as it's not accurate and the FE doesn't use it.

src/biokbase/narrative/jobs/job.py
```diff
@@ -341,16 +330,18 @@ def _update_state(self, state: dict) -> None:
...
                 return
 
         state = copy.deepcopy(state)
         if self._acc_state is None:
             self._acc_state = state
         else:
-            self._acc_state.update(state)
+            merge_inplace(self._acc_state, state)
+
+        self.last_updated = time.time_ns()
```
ialarmedalien, line 342: can you alter this so that all jobs updated that are updated from a batch of ee2 data get the same timestamp?

ialarmedalien, line 342: can you alter this so that all jobs updated that are updated from a batch of ee2 data get the same timestamp?

src/biokbase/narrative/tests/test_jobcomm.py
```diff
@@ -485,7 +497,22 @@ def test_start_job_status_loop__no_jobs_stop_loop(self):
     # Lookup all job states
     # ---------------------
 
+    def _check_pop_last_checked(self, output_states, last_checked=TEST_EPOCH_NS):
+        """
+        For STATUS responses, each output_state will have an extra field `last_checked`
+        that is variable and is not in the test data. Check that here and delete before
+        other checkd
```
ialarmedalien, line 504: `s/checkd/checks/`

ialarmedalien, line 504: `s/checkd/checks/`

ialarmedalien:
This is looking good!

Can you add in functionality such that if there are no updated jobs in response to a request, the backend returns an error?

Thanks!

src/biokbase/narrative/jobs/job.py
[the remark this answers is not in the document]
n1mus: Hmm. Good question. Thinking this through, I think this is a "natural" improvement because if the `state` is super minimal (like doesn't have `"job_input"`, and `_acc_state` __has__ accrued lots of information, like a `"job_input"`, then a superficial merge like `{**self._acc_state, **state}` would trigger a change indication

src/biokbase/narrative/jobs/job.py
n1mus: Curious ... why?
I was thinking about race conditions ...
I assuaged myself that all backend requests execute synchronously ...
The `ts`/`last_checked` and `last_updated` come from the backend so the STATUS requests are in effect atomic wrt to each other.
Am I missing something?

src/biokbase/narrative/jobs/job.py
n1mus: Yeah I think there's too much magic ...

src/biokbase/narrative/jobs/job.py
n1mus: I think if we just update `last_updated` when the jobs are updated from a batch of ee2 data though, there would be less chance of spam, because I'm actually hazy on the full behavior of the ee2 state and how it might spam `_acc_state` with updates

ialarmedalien: I want to ensure that when the frontend receives a batch of job data, all will have the same timestamp. If the timestamp is applied individually within each job object using `time.time_ns()`, there's the possibility that timestamps will vary across a batch -- esp. if batches are large.

src/biokbase/narrative/jobs/job.py
```diff
@@ -10,6 +10,7 @@
 import biokbase.narrative.clients as clients
 from biokbase.narrative.app_util import map_inputs_from_job, map_outputs_from_state
 from biokbase.narrative.exception_util import transform_job_exception
+from biokbase.narrative.jobs.util import merge, merge_inplace
```
ialarmedalien, line 13: Can you add `updated` to the `JOB_INIT_EXCLUDED_JOB_STATE_FIELDS` below --  it's not accurate or useful.

src/biokbase/narrative/tests/test_jobmanager.py
```diff
@@ -36,6 +37,7 @@
     BATCH_CHILDREN,
     BATCH_ERROR_RETRIED,
     BATCH_PARENT,
+    BATCH_RETRY_RUNNING,
```
ialarmedalien, line 40: is this used in this file? if not, can you remove it from the imports?

The VSCode python syntax highlighting / checking should indicate where you have unused imports.

n1mus: Oops yeah I should run flake8

src/biokbase/narrative/jobs/job.py
n1mus: Looks like `force_refresh` is just for getting `child_jobs` and `retry_ids`

src/biokbase/narrative/jobs/job.py
n1mus: Looks like this won't be a quick fix as it's at least somewhat baked into the tests

ialarmedalien: Does that mean you aren't going to do it or just that I should wait to see it turn up in the PR?

src/biokbase/narrative/tests/test_jobmanager.py
ialarmedalien: You can enable flake8 linting directly in vscode if you turn on / configure the various python extensions, so you can catch these things as you're editing. 🎉

n1mus: Heh yeah I have that ... you should see my code from when I used vim

src/biokbase/narrative/jobs/job.py
n1mus: Yeah ... I don't think there's enough time

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -307,10 +310,16 @@ def _construct_job_output_state_set(
...
+            now = time.time_ns()
             for job_id in jobs_to_lookup:
                 job = self.get_job(job_id)
                 if job_id in fetched_states:
-                    output_states[job_id] = job.output_state(fetched_states[job_id])
+                    fetched_state = fetched_states[job_id]
+                    # pre-emptively try a job state update
+                    # so can mark the bolus of fetched (but changed) states
+                    # with a simultaneous timestamp
+                    job._update_state(fetched_state, now)
```
n1mus, line 321: Didn't want to rename this PR to keep things simple in hopes of saving time

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -340,7 +349,22 @@ def get_job_states(self, job_ids: List[str], ts: int = None) -> dict:
         """
         job_ids, error_ids = self._check_job_list(job_ids)
         output_states = self._construct_job_output_state_set(job_ids)
-        return self.add_errors_to_results(output_states, error_ids)
+
```
n1mus, line 352: If I had more time I'd think about moving this logic to jobcomm.py ... seems like it might belong there instead

ialarmedalien: I think it belongs better here -- jobcomm.py is a fairly simple wrapper to convert between data structures and messages, whereas the job manager does various manipulations involving job data.

src/biokbase/narrative/jobs/jobmanager.py
```diff
@@ -307,10 +310,16 @@ def _construct_job_output_state_set(
...
             # if the job fetch failed, add an error message to the output
             # and return the cached job state
+            now = time.time_ns()
             for job_id in jobs_to_lookup:
                 job = self.get_job(job_id)
                 if job_id in fetched_states:
-                    output_states[job_id] = job.output_state(fetched_states[job_id])
+                    fetched_state = fetched_states[job_id]
+                    # pre-emptively try a job state update
+                    # so can mark the bolus of fetched (but changed) states
```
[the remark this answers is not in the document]
n1mus: That looks like what I wrote :sweat_smile:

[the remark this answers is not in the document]
n1mus: I should change it if you think it can be better written

src/biokbase/narrative/jobs/job.py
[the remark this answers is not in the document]
n1mus: Well ... I can't really think of one either. So I was originally thinking of this (or anything like it)
```
    def __setattr__(self, name, value):
        if name in STATE_ATTRS:
            self._update_state({name: value})
        elif name in JOB_INPUT_ATTRS:
            self._update_state({"job_input": {name: value}})
        elif name in NARR_CELL_INFO_ATTRS:
            self._update_state({"job_input": {"narrative_cell_info": {name: value}}})
        else:
            object.__setattr__(self, name, value)
```
But all the jobs *should* start out with the nested dict `"job_input"` set, and that nested dict *should* never change ...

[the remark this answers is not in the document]
ialarmedalien: This can be changed back to using `{**self._acc_state, **state}` since we've established there are no cases where the recursive merge would be used.

src/biokbase/narrative/jobs/jobcomm.py
```diff
@@ -514,6 +515,14 @@ def send_comm_message(self, msg_type: str, content: dict) -> None:
         Sends a ipykernel.Comm message to the KBaseJobs channel with the given msg_type
         and content. These just get encoded into the message itself.
         """
+        # For STATUS responses, add a last_checked field
+        # to each output_state. Note: error states will have
+        # the last_checked field too
+        if msg_type == MESSAGE_TYPE["STATUS"]:
+            now = time_ns()
+            for output_state in content.values():
+                output_state["last_checked"] = now
```
ialarmedalien, line 524: why not add this timestamp when the job manager is putting together the list of jobs, instead of adding an extra iteration through the job state data here?

src/biokbase/narrative/jobs/jobcomm.py
```diff
@@ -199,10 +203,7 @@ def _get_job_ids(self, req: JobRequest) -> List[str]:
         if req.has_batch_id():
             return self._jm.update_batch_job(req.batch_id)
 
-        try:
-            return req.job_id_list
-        except Exception as ex:
-            raise JobRequestException(ONE_INPUT_TYPE_ONLY_ERR) from ex
+        return req.job_id_list
```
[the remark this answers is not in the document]
n1mus: Because `req.job_id_list` should only reasonably throw one kind of exception, and that's already `JobRequestException(ONE_INPUT_TYPE_ONLY_ERR)`? Right ..?

src/biokbase/narrative/jobs/jobcomm.py
n1mus: Wasn't sure since the CANCEL_JOBS request also responds with a STATUS message.

n1mus: I decided not to filter the STATUS response for CANCEL_JOBS though because I figured in theory they should all get updated, whether successfully or just coming back with an error

src/biokbase/narrative/jobs/job.py
```diff
@@ -103,7 +104,7 @@ def __init__(self, ee2_state, extra_data=None, children=None):
         if ee2_state.get("job_id") is None:
             raise ValueError("Cannot create a job without a job ID!")
 
-        self._update_state(ee2_state)
+        self.update_state(ee2_state)
```
ialarmedalien, line 107: 👍

src/biokbase/narrative/jobs/job.py
```diff
@@ -129,9 +130,9 @@ def from_job_ids(cls, job_ids, return_list=True):
             return jobs
 
     @staticmethod
-    def _trim_ee2_state(state: dict, exclude: list) -> None:
-        if exclude:
-            for field in exclude:
+    def _trim_ee2_state(state: dict, exclude_fields: list) -> None:
```
ialarmedalien, line 133: 👍

src/biokbase/narrative/jobs/jobcomm.py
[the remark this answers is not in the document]
ialarmedalien: Yes, that's right... I am confused as to how that other code got in. Must have been a merge mishap. Anyway, it's fixed now.

src/biokbase/narrative/jobs/jobcomm.py
ialarmedalien: Because everything is asynchronous, the FE doesn't have any way of knowing what triggered a job status message -- whether it was a cancel request, a status request, or the BE job loop. That's why I say it's better to put the timestamp on in the job manager, so that all job state objects that the FE receives have a timestamp on them.

src/biokbase/narrative/tests/job_test_constants.py
```diff
@@ -32,6 +32,13 @@ def generate_error(job_id, err_type):
     return error_strings[err_type]
 
 
+def trim_ee2_state(ee2_state, exclude_fields):
```
ialarmedalien, line 35: don't we have this code somewhere else?

src/biokbase/narrative/tests/narrative_mock/mockclients.py
```diff
@@ -244,7 +245,13 @@ def list_objects(self, params):
     # ----- Execution Engine (EE2) functions -----
 
     def check_workspace_jobs(self, params):
-        return self.job_state_data
+        ee2_states = self.job_state_data
+        if params.get("exclude_fields"):
+            for ee2_state in ee2_states.values():
+                trim_ee2_state(ee2_state, params["exclude_fields"])
+        if params.get("return_list"):
+            ee2_states = list(ee2_states.values())
```
ialarmedalien, line 253: Do any of those params ever change? There's only one place where `check_workspace_jobs` gets called, and the params are always the same, so...

src/biokbase/narrative/tests/test_job.py
```diff
@@ -400,11 +402,73 @@ def test_job_update__invalid_job_id(self):
...
+            self.assertEqual(last_updated, job.last_updated)
+
+            trim_ee2_state(ee2_state, OUTPUT_STATE_EXCLUDED_JOB_STATE_FIELDS)
+            job.update_state(ee2_state)
+            self.assertEqual(last_updated, job.last_updated)
+
+            job.update_state({})
+            self.assertEqual(last_updated, job.last_updated)
+
+            # job has init ee2 state
```
ialarmedalien, line 439: This test looks suspiciously spaghetti code-like. Does it need to be done as this long series of transitions or can it be split into separate tests?

src/biokbase/narrative/tests/test_job.py
```diff
@@ -757,7 +821,7 @@ def test_in_cells__batch__same_cell(self):
         batch_job, child_jobs = batch_fam[0], batch_fam[1:]
 
         for job in child_jobs:
-            job.cell_id = "hello"
+            job._acc_state["job_input"]["narrative_cell_info"]["cell_id"] = "hello"
```
ialarmedalien, line 824: Was this the only place you could find where an attribute was changed (other than via the `update_state` method)?

src/biokbase/narrative/tests/job_test_constants.py
n1mus: Yep, `Job._trim_ee2_state`. I just got tired of using that in tests when usually we use independent testing functions

src/biokbase/narrative/tests/narrative_mock/mockclients.py
n1mus: No, but I thought it might be a good idea to implement the `"exclude_fields"` param since here I'm paying closer attention to when state updates are triggered

src/biokbase/narrative/tests/test_job.py
n1mus: I thought it followed a very similar pattern throughout and so could flow in a single function. The punchline is `last_updated` defined at the top never changes throughout these tests. Is there a benefit to making tests methods small?

src/biokbase/narrative/tests/test_job.py
n1mus: I checked every field in `job.__setattr__` that was from the `"job_input"`. I didn't check anything in the outer level of the ee2 state.

n1mus: But `update_state` is the only place `_acc_state` is mutated

ialarmedalien: you left a TODO comment about whether the attribute setter was ever used in `job.py` -- seems as though you've answered it here!

src/biokbase/narrative/jobs/jobcomm.py
n1mus: I think one allure of putting everything into JobComm is less tests surgery ... But putting it deep into the stack, at the origin of the STATUS response ds, seems less googly-eyed

n1mus: Okay I tried putting all the filtering/last_checked logic at the source `_construct_job_state_set` but the tests were complaining so I'm abandoning that effort for the sake of time. Is the current placement of the filtering/last_checked good enough?

src/biokbase/narrative/tests/test_job.py
ialarmedalien: It's much easier to read, understand, and update/edit a couple of stanzas of code than it is a long series of stanzas. Unless there is a specific need to test a sequence of modifications (e.g. there's something going on elsewhere that changes state as a result of these mods), it's best to make tests as simple as possible to assist future codebase editors and maintainers.

src/biokbase/narrative/tests/narrative_mock/mockclients.py
ialarmedalien: has adding the `exclude_fields` param changed the output of the function?

src/biokbase/narrative/tests/test_job.py
n1mus: Ah. But what if it's two long stanzas of highly repetitive code? With a common punchline that is accentuated by more repetition?

src/biokbase/narrative/tests/narrative_mock/mockclients.py
ialarmedalien: Today's "good idea to implement" is tomorrow's "why on earth did someone write this?". YAGNI. 😄

src/biokbase/narrative/tests/test_job.py
ialarmedalien: If it's highly repetitive, it suggests that the repetition could be abstracted out into a function... or that it could be replaced by individual tests that validate the atomic operations involved.

PR #3000, opened 2022-06-01
DATAUP-718 config tab performance boost part 2

briehl:

This adjusts selectInputs (so, all the dropdowns that aren't for workspace objects) to only render up to 20 items at once at startup. If there are more than that (like, 500+ file path options), then don't render any and only put them in the DOM when the dropdown is clicked.

The 20 is kinda arbitrary, but seemed like a decent cutoff. I don't have real statistics, but there's a touch of a performance hit when having to render all dropdown items each time, which now happens for the >20 item dropdowns.

Still, this means that for having 500 file paths in a bulk import cell, we're not shoving 25,000 `<option>` elements in the DOM. So that's helpful.

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/DATAUP-718

# Testing Instructions

# Dev Checklist:

- [n/a] [Version has been bumped](https://semver.org/) for each release
- [n/a] [Release notes](/RELEASE_NOTES.md) have been updated for each release (and during the merge of feature branches)

### DIFF
--- a/kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
+++ b/kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
@@ -17,7 +17,8 @@ define([
     // Constants
     const t = html.tag,
         div = t('div'),
-        select = t('select');
+        select = t('select'),
+        MAX_OPTIONS = 20;
 
     /**
      *
@@ -206,6 +207,19 @@ define([
 
         // LIFECYCLE API
 
+        function buildAllOptions() {
+            return model.availableValues.map((item) => buildOption(item));
+        }
+
+        function buildOption(item) {
+            return {
+                value: item.value,
+                id: item.value,
+                text: item.display,
+                disabled: model.disabledValues.has(item.value),
+            };
+        }
+
         function start(arg) {
             return Promise.try(() => {
                 parent = arg.node;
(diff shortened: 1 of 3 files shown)


kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
```diff
@@ -221,6 +216,19 @@ define([
...
+            return model.availableValues.map((item) => buildOption(item));
+        }
+
+        function buildOption(item = {}) {
+            return {
+                value: item.value,
+                id: item.value,
+                text: item.display,
+                disabled: model.disabledValues.has(item.value),
+            };
```
ialarmedalien, line 229: do you need the extra function here? I find it just as easy to parse
```js
   return model.availableValues.map((item) => {
      return {
         value: item.value,
         ...
      }
   })
```

but that might just be me!

kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
```diff
@@ -18,7 +18,7 @@ define([
     const t = html.tag,
         div = t('div'),
         select = t('select'),
-        option = t('option');
+        MAX_OPTIONS = 20;
```
ialarmedalien, line 21: I wonder if there are likely to be examples of app cells with numerous select elements with more than 20 options and if it's worth checking for that -- i.e. only doing this if the selects are in a bulk cell. Hmmm, on second thoughts, probably not worth it.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
```diff
@@ -164,7 +150,16 @@ define([
         }
 
         function setModelValue(value) {
+            // This might be a bit of a cheat. I'm not sure what'll happen if we just
+            // remove the current value from disabled values before updating.
+            // WE'LL SEE!
```
ialarmedalien, line 155: 😱

ialarmedalien (approved):

kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
briehl: Yeah, `buildOption` gets used elsewhere as a convenience function in the case where we want to start up with a single item already selected.

kbase-extension/static/kbase/js/widgets/appWidgets2/input/selectInput.js
briehl: For the record, it works fine. This is just in there to force the update early. The order of events when clicking the little "clear selection x" button seems to be (before this bit):
1. delete selection
2. set model value 
3. open dropdown
4. validate new value
5. get new list of disabled / enabled values from widget controller

So it was deleting the value, showing the error that you need a new value at the same time as opening the dropdown, and not actually showing that the recently deleted value is now available. You'd have to close the dropdown first so the select2-rendered bits would get the update, THEN it would become available again. It was really annoying, especially since we know that once the item is cleared or changed, it should instantly become available as a selection again.

This **might** have some longer term issues, but was an easier fix than figuring out how to make select2 not open the dropdown when hitting the clear button.

PR #3002, opened 2022-06-03
June JavaScript dependency updates

briehl:

Thanks, `dependabot`! Um, I think.

Leaving eslint as-is, because I think we decided the last version bump would have changed practically everything?


ialarmedalien:
eslint should be OK -- it's Prettier that sometimes causes commit chaos.

ialarmedalien (approved):

PR #3178, opened 2023-01-26
CE-136 JSON export download logging

eapearson:

This is a set of speculative changes to support logging of JSON downloads from the Narrative front end. This is primarily for discussion and exposition of one way to implement this in a manner quite similar to how front end logging has been implemented in the past.

The overall effort is to be able to log object JSON downloads, as triggered by the Narrative front end via the data panel's export functionality. These log entries should be available in a timely manner to KBase staff. In our specific case they are needed as part of Narrative object usage statistics. 

The logging works by sending Python code to the back end for creating and emitting a log entry in JSON format via the standard Python logging API.

The log entries are currently appended to /tmp/kbase-narrative-ui.json. This is the same location (/tmp) but a different file name and format than the current logging mechanism uses. 

I've separated most of the code so that current logging works unimpeded, both in the JS front end and Python back end code.

After getting the initial functionality working without too much trouble, I invested a couple hours trying to get more context into the log. One such piece of information is the narrative object reference. I tried but failed to get the current narrative version (e.g. after it obtains a new version when saved), but left that code in place anyway.

The log format is not something I put a tremendous amount of thought or any research into. At a minimum I want to satisfy our immediate needs - the time, username, and downloaded object ref. Being straightforward to parse is another important feature, so JSON. There is additional structure to make the log entries a bit more robust in the presence of other types of log entries, but clearly it is just a starting point.

Ultimately, we may want to use a different logging handler to send the log entries to a network endpoint for ingestion into a service and/or database.

# Jira Ticket / Issue #

- CE-136 https://kbase-jira.atlassian.net/browse/CE-136

- [ ] Added the Jira Ticket to the title of the PR (e.g. `DATAUP-69 Adds a PR template`)

# Testing Instructions

# Dev Checklist:

### DIFF
--- a/src/biokbase/narrative/common/kblogging.py
+++ b/src/biokbase/narrative/common/kblogging.py
@@ -300,19 +300,45 @@ def log_ui_event(event: str, data: dict, level: str):
 
     
     message = json.dumps({
-        "id": log_id,
-        "event": event,
-        "timestamp": epoch_time_millis(),
-        "username": kbase_env.user,
-        "client_ip": kbase_env.client_ip,
+        # If logs are combined, we need to tie log entries to 
+        # a specific version of a service in a specific environment.
+        "service": "narrative",
+        "version": biokbase.narrative.version(),
         "environment": kbase_env.env,
-        "narrative": {
-            "workspace_id": int(workspace_id),
-            "object_id": int(object_id),
-            "ref": kbase_env.narrative_ref,
-            "version": biokbase.narrative.version()
-        },
-        "data": data
+        # General log entry; information that any log entry
+        # will need. 
+        # id helps create a unique reference to a log entry; perhaps
+        # should be uuid, service + uuid, or omitted and only created
+        # by a logging repository service. 
+        "id": log_id,
+        "timestamp": epoch_time_millis(),
+        # The actual, specific event. The event name is a simple
+        # string, the data is a dict or serializable class whose
+        # definition is implied by the event name.
+        "event": {
+            "name": event,
+            # the event context captures information common to instances 
+            # of this kind of event. As a narrative ui event, part of the context
+            # is the narrative object, the current user, and the current user's
+            # browser. Clearly more could be captured here, e.g. the browser model,
+            # narrative session id, etc.
+            "context": {
+                # I tried to update kbase_env to reflect the current narrative ref,
+                # but no no avail so far. The kbase_env here is not the same as the 
+                # one used when saving a narrative and getting the new version, so it does
+                # not reflect an updated narrative object version.
+                # If that isn't possible, we can store the ws and object id instead.
+                "narrative_ref": kbase_env.narrative_ref,
+                # Log entries for authenticated contexts should identify the user
+                "username": kbase_env.user,
+                # Log entries resulting from a network call can/should identify
+                # the ip address of the caller
+                "client_ip": kbase_env.client_ip
+                # could be more context, like the jupyter / ipython / etc. versions
+            },
+            # This is the specific data sent in this logging event
+            "data": data
+        }
     })
 
     if level == 'debug':


kbase-extension/static/kbase/js/util/Logging.js
```diff
@@ -0,0 +1,86 @@
+define([], () => {
+    /**
+     * 
+     * @param {string} code - executable Python code in a string
+     * @returns {Promise<string>} - A promise which will result in a string the value
+     *                              of which was that returned by the last statement
+     *                              in the Python code.
+     */
+    async function executePython(code) {
```
briehl, line 9: This seems reasonable. I think there's some other error catching / wait-until-kernel-is-alive code elsewhere that we can either repurpose or just make use of.

briehl (commented):
This is a good start. I think the overall flow of event -> kernel message -> logging -> log destination will work fine. 

Would could also create a dedicated kernel channel for logging, like we do for job messaging, but I don't really think that's necessary for a starting point.

Also noting that I'd like to see the rather ancient KBError / KBFail messages make use of something like this, but that's out of scope for the prototype.

I guess my only concerns are:
1. where do the logs go?
2. what format should logs be?
3. do we keep a parallel log on the file system / make that an option for running in dev mode?
4. can we unify how all the logs are sent and processed so we can combine some older stuff and remove some cruft?

I know that expands the scope of this, so we can probably tackle all of that piecemeal. Once we can answer 1 and 2 with some devops input, we can satisfy some metric needs and build up the rest.

kbase-extension/static/kbase/js/util/Logging.js
eapearson: Ok, I'll hunt for that.

eapearson:
> I guess my only concerns are:
> 
> 1. where do the logs go?

Exactly -- I think that informs a lot of the rest. E.g. if we use a 3rd party logging service, the where is pretty much solved (unless we decide to hide it behind a service!), the format will be at least constrained, the metadata will be taken care of.

> 2. what format should logs be?
> 3. do we keep a parallel log on the file system / make that an option for running in dev mode?

I don't think in general we want to duplicate the logs. I mean, there is an argument for logging to file + service if the files are actually preserved, to serve as a backup in case the service is down and we could then backfill the logging service. But that also might be a "nice to have".

For development we could use a container to receive logs - if we use a 3rd party logging service, it should be easy to stand up such a service. Complicates development slightly but simplifies the Narrative code + configuration. Logging services should have nice tools for accessing the logs, filtering them, etc. Then again, Python logging allows one to retarget logs easily, though we'd need some sort of configuration to be able to switch it. 

> 4. can we unify how all the logs are sent and processed so we can combine some older stuff and remove some cruft?
> 
> I know that expands the scope of this, so we can probably tackle all of that piecemeal. Once we can answer 1 and 2 with some devops input, we can satisfy some metric needs and build up the rest.

It surely is sensible to have all logging be consistent, but we may want to limit the initial implementation (map) to these logs as it only affects a single activity within the Narrative -- logging from the JSON download, which is not a common user action.

src/biokbase/narrative/common/kblogging.py
```diff
@@ -252,3 +260,97 @@ def __init__(self, is_fatal, where="unknown location", what="unknown condition")
...
+        
+def log_ui_event(event: str, data: dict, level: str):
+    # We use a separate logger, configured to save the 
+    # entire message as a simple string ... and that string
+    # is a JSON-encoded message object.
+    # The resulting log file, then, is y it is a JSON stream format, since it
+    # contains multiple objects in sequence.
+    ui_log = get_logger("narrative_ui_json")
+    log_id = str(uuid.uuid4())
+    if not _has_handler_type(ui_log, logging.FileHandler):
```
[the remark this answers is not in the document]
eapearson: Yeah, I just wanted to make sure that this prototyping work actually produced a JSON log in a minimal fashion with the simplest implementation -- a file in the same place we are currently logging (/tmp). And I agree it should be global. I figured, get it working in this case, then later switch over existing logging entry points to use it too.

I'm sure on the table still is logging to a file in a mounted directory and having an external process grab those logs. I don't think that is a _good_ solution for Narrative logging, but it should be considered, as it is a traditional logging technique.

PR #3207, opened 2023-02-16
UIP-19 fix build image issues

briehl:

A couple issues that show up downstream of the previous PR.
1. We still need some additional Docker build args for the Narrative containers, so we can't use the KBase-provided build actions. These were modified here.
2. There was some leftover debugging script that's now causing an issue.
3. The parallel PR-build script (which uses the reusable docker build actions) was made manual, but the non-manual bit was left commented out for possible further use.
4. Also fixes a bug in how image names are referenced - `narrative` vs `narrative-develop`, for example. Tests were using `narrative` while the previous PR fixed the issue and created `narrative-develop` properly. But that was masked by already having a `narrative` image for that PR while it was in progress. Tricksy!

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/UIP-19

Rest of the usual stuff isn't relevant, and was removed.

### DIFF
--- a/.github/workflows/build_test_pr.yaml
+++ b/.github/workflows/build_test_pr.yaml
@@ -16,6 +16,11 @@ jobs:
       GHCR_USERNAME: ${{ secrets.GHCR_USERNAME }}
       GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
 
+  trivy-scans:
+    if: (github.base_ref == 'develop' || github.base_ref == 'main' || github.base_ref == 'master' ) && github.event.pull_request.merged == false
+    uses: kbase/.github/.github/workflows/reusable_trivy-scans.yml@main
+    secrets: inherit
+
   run_unit_tests:
     uses: ./.github/workflows/unit_test.yml
     needs: run_build_and_push
--- a/.github/workflows/pr_build.yml
+++ b/.github/workflows/pr_build.yml
@@ -1,16 +1,17 @@
 ---
 name: Pull Request Build, Tag, & Push
 on:
-  pull_request:
-    branches:
-      - develop
-      - main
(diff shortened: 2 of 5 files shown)


briehl:
The various Trivy test fails should get fixed in the next PR when the dependencies get updated.

briehl:
OR possibly the one after that when the narrative-base-image gets an update.

.github/workflows/pr_build.yml
```diff
@@ -1,16 +1,17 @@
...
+  # pull_request:
+  #   branches:
+  #     - develop
+  #     - main
+  #     - master
+  #   types:
+  #     - opened
+  #     - reopened
+  #     - synchronize
+  #     - closed
```
ialarmedalien, line 14: do you want to just delete this or is it being temporarily disabled?

.github/workflows/build_test_pr.yaml
```diff
@@ -16,6 +16,11 @@ jobs:
       GHCR_USERNAME: ${{ secrets.GHCR_USERNAME }}
       GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
 
+  trivy-scans:
+    if: (github.base_ref == 'develop' || github.base_ref == 'main' || github.base_ref == 'master' ) && github.event.pull_request.merged == false
+    uses: kbase/.github/.github/workflows/reusable_trivy-scans.yml@main
+    secrets: inherit
```
ialarmedalien, line 22: you could remove the reference to the `master` branch as I think we switched the narrative over to using `main`

ialarmedalien (approved):
Looks good -- a couple of minor pieces that may or may not need editing

PR #3275, opened 2023-05-16
PTV-1877 - wrong url for data upload/download guide in bulk import cell's "info" tab

eapearson:

This fixes an issue reported by a user in https://kbase-jira.atlassian.net/browse/PUBLIC-2378, reported as a broken link. The link in question should be to the Data Upload and Download Guide. The url for this guide changed when the single kbase.us public information site was reimplemented as separate documentation and marketing sites. It seems that most import apps invoked by the Import panel, and subsequently managed by the bulk import cell, include hardcoded links to this documentation.

Now, the wisdom of that aside, at least some apps still, after over a year, have now-obsolete links to that documentation. Although the response in the public ticket is that the app has not been updated, there is indeed a fix in this PR that will repair such obsolete links. The fix is to simply replace such links with the (currently) correct one before the link is rendered. These links appear in the "description" field for an app, so that is the field which has the fix applied.

In addition to this fix, another issue is that some of these links are not set up to open in a new browser window. This is a bad experience for users, as it causes the Narrative to be unloaded and replaced with the documentation. If a narrative is actively being edited, as will be the case for a user who just inserted the import cell and is attempting to read the documentation, the browse will also pop up an alert warning the user that they are about to leave the current page. This warning is issued because the Narrative contains unsaved changes. So another "fix" is to add 'target="_blank"' to such links.

Finally, since the description field is expected to contain html, it is subjected to DOM Purify sanitize. While the source of the content is the description field from the app spec, it is still wise to sanitize this content since it does originate outside the codebase and the narrative runtime. Although apps are curated, it is quite possible for a reviewer to be unfamiliar with unsafe html and for something to slip through. And, in any case, it is easy to accomplish.

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/PTV-1877

# Testing Instructions

# Dev Checklist:

- [-] My code follows the guidelines at https://sites.google.com/lbl.gov/trussresources/home?authuser=0
  - there don't seem to be specific coding guidelines at this url
- [-] (Python) I have run Black and Flake8 on changed Python code manually or with a git precommit hook
- [-] Any dependent changes have been merged and published in downstream modules

- [-] [Version has been bumped](https://semver.org/) for each release
- [-] [Release notes](/RELEASE_NOTES.md) have been updated for each release (and during the merge of feature branches)

### DIFF
--- a/kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
+++ b/kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
@@ -158,9 +158,7 @@ define(['domPurify', 'common/format', 'common/html', 'util/string'], (
                             if (
                                 node
                                     .getAttribute('href')
-                                    .match(
-                                        new RegExp('https?://kbase.us/data-upload-download-guide')
-                                    )
+                                    .match(/http[:]\/\/kbase.us\/data-upload-download-guide/)
                             ) {
                                 node.setAttribute(
                                     'href',
@@ -168,9 +166,7 @@ define(['domPurify', 'common/format', 'common/html', 'util/string'], (
                                 );
                             }
                             // Ensure all docs links open into a new window (or tab).
-                            if (
-                                node.getAttribute('href').match(new RegExp('https://docs.kbase.us'))
-                            ) {
+                            if (node.getAttribute('href').match(/https:\/\/docs.kbase.us/)) {
                                 node.setAttribute('target', '_blank');
                             }
                         }
--- a/kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
+++ b/kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
@@ -1,5 +1,13 @@
-define(['common/format', 'common/html', 'util/string'], (format, html, string) => {
+define(['domPurify', 'common/format', 'common/html', 'util/string'], (
+    DOMPurify,
+    format,
+    html,
+    string
+) => {
     'use strict';
+
+    // Note - would destructure in the arguments, but that would require a change to eslint rules.
+    const { sanitize } = DOMPurify;
 
     const cssBaseClass = 'kb-info-tab',
         div = html.tag('div'),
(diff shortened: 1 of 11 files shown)


eapearson:
hello @briehl, trying to get this small PR into shape, going bat-blank crazy with the tests.

Can you see why the isolated tests are reporting failure? They succeeded earlier today. I did update chrome launcher to fix the broken integration tests, don't know if that would affect the isolated tests, but think not since the report shows 54 tests completed, no errors, and the workflow proceeds to coverage, as if all tests succeeded. The summary reports an error condition (1) so it fails overall.

Oh, and now the integration tests are failing with timeouts. I don't want to spend too much precious "small bug" time just getting tests to work - so any help much appreciated :)

eapearson:
So in a separate PR the unit tests were fixed, when that PR is merged, I'll update this PR with those changes, and tests should pass.

briehl:
Tests all seem to pass now, thanks for fixing them up in the other PR. Yeah, the integration test timeouts are annoying, but I think that's a mix of all the tools and environments (including GHA's) causing issues. Mostly, that failure is confined to a single test that fails intermittently, so as long as that's the one that still fails, it hasn't been high priority to fix.

kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
```diff
@@ -1,6 +1,14 @@
-define(['common/format', 'common/html', 'util/string'], (format, html, string) => {
+define(['domPurify', 'common/format', 'common/html', 'util/string'], (
+    DOMPurify,
```
briehl, line 2: Nice inclusion of DOMPurify here.

briehl:
> It seems that most import apps invoked by the Import panel, and subsequently managed by the bulk import cell, include hardcoded links to this documentation.

YUCK. That's the problem to fix right there.

kbase-extension/static/kbase/js/common/cellComponents/tabs/infoTab.js
```diff
@@ -119,7 +127,75 @@ define(['common/format', 'common/html', 'util/string'], (format, html, string) =
...
+                 *
+                 * At present, the fixes include:
+                 * - for the app info description:
+                 *  - replace the old data upload/download guide link with the current one
+                 *  - ensure that all KBase doc links open in a new window/tab
+                 *
+                 * @param {Node} node A DOM node
+                 * @returns {string} the input node, with fixes applied.
+                 */
+                const fix = (node) => {
```
briehl, line 154: I think this is fine for now to patch over the bug, but it's really a band-aid over the real problem. The actual problem is that our docs URLs should be expected to be stable or have a redirect, and that one apparently does not. I mentioned in the ticket that fixing it here is fine, and those older narratives that use this would benefit from the update. However, these should probably not be hardcoded in the apps or in the Narrative, but configured somewhere. 

Not a concern right now, just noting my opinion.

briehl (approved):

PR #3276, opened 2023-05-17
PTV-1878  fix failing front end unit tests, PTV-1879 fix integration tests

eapearson:

There are a few front end unit tests which consistently fail. This complicates any current work from the develop branch. This is an effort to address each of the test suite issues.

Here is the top Jira comment explaining:

There are just a few issues to resolve, but when they are fixed, the front end unit tests pass consistently. Before these fixes, there were some consistent failures, as well as non-deterministic ones.

The consistent failures are:

* dynamic dropdown tests which fail because a button in the select2 control with the text "x" is captured when the test extracts the text for comparison. The test for the text is exact, so the "x" in the test causes the comparison to fail. Using a "contains" test would be appropriate here. After all, the test should be that some given text is displayed to the user, not that it is all  that is displayed.
 
* A test of the big "+" button in the data list panel asserts that it is a "block" display; however, buttons are, by default "inline-block" unless the button style is overridden. It may have been overridden in the past, but not not. In any case, the meaning of the test is that the button is displayed to the user, not the specific display style. Since the button visibility is controlled by jquery hide() and show(), it is important to know that when hidden the display style is "none", and when shown it is whatever display style the button has or had before being hidden. Therefore, the propre test is that the display style is not "none".

The non-deterministic failures are caused by the "isolated tests". These tests are those that are known to destabiilize the unit tests. Unfortunately, in the karma.local.conf.js config file, a filter on these excluded files removes them from the exclusion list. Removing that filter  re-enables the exclusion of those tests, and the seemingly "random" test failures seem to have been quelled. Presumably the GHA test scripts do not use karma.local.conf.js.

Also note that this does not address the failing integration tests. That is due to the karma launcher being out of sync with the GHA-installed Chromium. I addressed that in the previous PR, but that just resulted in integration test timeouts. And, besides, the better fix is to do something like use the Chromium binary from puppeteer, and keep the puppeteer and chrome launcher dependencies in sync. (I've done this before and it works fine.)

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/PTV-1878

# Testing Instructions

# Dev Checklist:


test/unit/karma.local.conf.js
```diff
@@ -9,28 +9,25 @@ module.exports = function (config) {
     });
     config.preprocessors = {};
     config.reporters = ['kjhtml', 'brief'];
-    config.exclude = config.exclude
-        .filter((file) => {
-            return file.indexOf('test/unit/spec/') === -1;
-        })
```
ialarmedalien, line 15: is there something else that performs this filtering?

ialarmedalien (approved):

eapearson:
BTW @briehl , @ialarmedalien I think the reason for the integration test timeouts is an expired token. I can run the tests fine locally, although two of them need updating for changes in in public data (I think).

So, I could either update this PR with a few small changes (update the chrome launcher, fix two tests), and then someone updates the test token, or I can do that in a separate PR. The advantage of doing it here is that all tests will pass green, and then I can apply this set of changes to PR #3275.

eapearson:
So I went ahead to try to fix the integration tests too. It looks like it is just a few updates to node dependencies. The chrome driver is out of sync with the node version (122) installed, and refuses to run. Updating to to chromedriver 122 fixes this. The wdio library version installed (8.3.9) is broken wrt the latest chrome - the scrollIntoView DOM method no longer works correctly, causing several tests to fail. Updating to 8.10.2 solves this problem.

That said, the tests are still finicky and, at least locally, still prone to timeout failures.

And at present, the testing token is expired (I think), so all tests time out, and they fail after about 1/2 hour.

test/unit/spec/narrative_core/kbaseNarrativeDataList-spec.js
```diff
@@ -230,7 +230,10 @@ define([
             expect($addDataButton).toBeDefined();
             expect($addDataButton.length).toEqual(1);
             expect($addDataButton.is('button')).toBeTruthy();
-            expect($addDataButton.css('display')).toEqual('block');
+            // What we are testing is that the button has been displayed - it is controlled
```
briehl, line 233: 👍

briehl (approved):
Looks good. Thanks for the fixes.

PR #3320, opened 2023-07-27
PTV-1810 fix obj input 1

briehl:

This is the first of probably 3 or 4 PRs related to some fixes for PTV-1810.
The short version is with the recent updates to that bug, input objects are stored in app cells as UPAs and passed to the app framework that way. The problem here is that apps don't always expect UPAs (ALTHOUGH THEY SHOULD). Older ones can expect object names, or references of the form ws_name/obj_name. 

The goal of this chain of PRs is to update the Narrative backend code to detect those requirements and parse UPAs back into lesser, worse formats for apps that want them. Backward compatibility and all that.

This will mainly require updating `biokbase.app_util.transform_object_value` - an old bit of code that, with the lightest poking, reveals ancient misguided decisions enabled by Python's "anything goes!" flexibility around input types. Again, the goal isn't to fully fix all of that, which would be a very large effort, but to put in an effective patch without creating more chaos. Currently, that function implicitly expects that an object name is provided, and transforms based on that. This update will explicitly expect that any object-ish string (name, reference, reference-path, UPA) is provided, and transform based on the given parameter spec.

Sounds simple, but it's less so.

PR strategy:
1. (this PR) move the `system_variable` and `strict_system_variable` functions to their own module to avoid what will become circular dependencies in the next PR.
2. Make use of the UPA API to detect whether inputs are UPAs or not (currently some hand-crafted code in there), add missing tests.
3. Adjust the rest of that code to fit our needs, add thorough unit and integration tests.

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/PTV-1810
- [X] Added the Jira Ticket to the title of the PR (e.g. `DATAUP-69 Adds a PR template`)

# Testing Instructions

# Dev Checklist:

- [X] My code follows the guidelines at https://sites.google.com/lbl.gov/trussresources/home?authuser=0
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [n/a] (JavaScript) I have run Prettier and ESLint on changed code manually or with a git precommit hook
- [X] (Python) I have run Black and Flake8 on changed Python code manually or with a git precommit hook


briehl:
Codacy is whining about `assert` being used throughout test code. Which is what it's for. Ugh.

src/biokbase/narrative/app_util.py
```diff
@@ -162,6 +99,31 @@ def _untransform(transform_type, value):
 
 
 def app_param(p):
+    """
```
briehl, line 102: Added a bit of notes here that weren't exactly intended for this PR, though still helpful.

src/biokbase/narrative/system.py
```diff
@@ -0,0 +1,64 @@
+import os
```
briehl, line 1: This is the bulk of the change - migrate `system_variable` and `strict_system_variable` over here.

Really, these should be separate functions - `get_workspace`, `get_workspace_id`, etc. But that's out of scope for this little project. It was originally done with strings to map directly to various app spec requests.

src/biokbase/narrative/tests/test_system.py
```diff
@@ -0,0 +1,90 @@
+import os
```
briehl, line 1: Redid the tests here as `pytest` style, since we're using `pytest` anyway and it would be nice to migrate there.

src/biokbase/narrative/upa.py
```diff
@@ -5,20 +5,19 @@
 """
 
 import re
-
-from .app_util import system_variable
+from .system import system_variable
 
 external_tag = "&"
```
briehl, line 12: The various typing things also weren't exactly intended to pop into this PR, but they're benign and generally helpful.

src/biokbase/narrative/system.py
```diff
@@ -0,0 +1,64 @@
...
+            return None
+        # TODO: make this better with more exception handling.
+    elif var == "timestamp_epoch_ms":
+        # get epoch time in milliseconds
+        return int(time.time() * 1000)
+    elif var == "timestamp_epoch_sec":
+        # get epoch time in seconds
+        return int(time.time())
+    else:
+        return None
```
ialarmedalien, line 63: can just do `return None` here - no need for final `else`.

src/biokbase/narrative/system.py
```diff
@@ -0,0 +1,64 @@
...
+        token - returns the current user's token credential
+        user_id - returns the current user's id
+
+    if anything is not found, returns None
+    """
+    var = var.lower()
+    if var == "workspace":
+        return os.environ.get("KB_WORKSPACE_ID", None)
+    elif var == "workspace_id":
+        ws_name = os.environ.get("KB_WORKSPACE_ID", None)
```
ialarmedalien, line 36: Doesn't `os.environ.get("blahblahblah")` return None by default if it can't find the value?

briehl: Yup. I really didn't modify that code, just copy-pasted. I'll make it better.

src/biokbase/narrative/system.py
briehl: You're right. Fixed.

src/biokbase/narrative/system.py
briehl: Fixed up. Also swapped `BaseException` over to `Exception` for those try/catch blocks. That #TODO in there still holds - we should do some better error trapping there.

PR #3343, opened 2023-08-10
UIP-36 add user information fetching to the Auth client

briehl:

This adds resources around the `/me` path of the auth service. This wasn't needed before as all the info that the Narrative backend needed was in the `/token` endpoint. But now it needs the user's anonymous id. This adds a `UserInfo` wrapper class (along with a couple accessories) and a `get_user_info` endpoint.

It's not used anywhere yet, but that'll change in the next PR.

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/UIP-36

# Testing Instructions

# Dev Checklist:

- [n/a] (JavaScript) I have run Prettier and ESLint on changed code manually or with a git precommit hook
- [x] (Python) I have run Black and Flake8 on changed Python code manually or with a git precommit hook

- [n/a] [Version has been bumped](https://semver.org/) for each release
- [n/a] [Release notes](/RELEASE_NOTES.md) have been updated for each release (and during the merge of feature branches)


src/biokbase/auth.py
```diff
@@ -41,6 +41,48 @@ def __init__(self, info_dict: dict, token: str = None):
         self.token = token if token is not None else info_dict.get("token")
 
 
+class UserRole:
```
briehl, line 44: These utility wrapper classes are basically a way to make it easy to set defaults, and not carry dicts around with semi-random keys and force whoever consumes them to do the `.get("key", "default")` shuffle.

We could add more specific attribute getters, setters, etc., but IMO, YAGNI ™️ .

src/biokbase/auth.py
```diff
@@ -41,6 +41,48 @@ def __init__(self, info_dict: dict, token: str = None):
...
+    def __init__(self, role_info):
+        self.role_id = role_info.get("id")
+        self.description = role_info.get("desc")
+
+
+class PolicyId:
+    def __init__(self, policy_info):
+        self.policy_id = policy_info.get("id")
+        self.agree_date = policy_info.get("agreedon")
```
ialarmedalien, line 53: _The least feared of the local mafia bosses was Frankie Deltorino, better known as "the Agree Don"._

src/biokbase/auth.py
```diff
@@ -41,6 +41,48 @@ def __init__(self, info_dict: dict, token: str = None):
...
+
+
+class Identity:
+    def __init__(self, ident_info):
+        self.ident_id = ident_info.get("id")
```
ialarmedalien, line 58: could this just be `self.id`?

src/biokbase/auth.py
```diff
@@ -67,6 +109,20 @@ def get_auth_token() -> Optional[str]:
...
+def get_user_info(token: str) -> UserInfo:
+    """
+    Returns the UserInfo object with information about the user who
+    owns the valid token. If the token is invalid (i.e. the status_code
+    of the request is not ok), an HTTPError exception is raised.
+    """
+    headers = {"Authorization": token}
+    r = requests.get(token_api_url + endpt_me, headers=headers)
+    if r.status_code != requests.codes.ok:
+        r.raise_for_status()
```
ialarmedalien, line 121: could just do `r.raise_for_status()` here

src/biokbase/auth.py
```diff
@@ -67,6 +109,20 @@ def get_auth_token() -> Optional[str]:
...
+    user_info = UserInfo(r.json())
+    return user_info
```
ialarmedalien, line 123:
```py
return UserInfo(r.json())
```

src/biokbase/auth.py
```diff
@@ -41,6 +41,48 @@ def __init__(self, info_dict: dict, token: str = None):
...
+        self.user = user_dict.get("user")
+        self.created = user_dict.get("created")
+        self.last_login = user_dict.get("lastlogin")
+        self.display_name = user_dict.get("display")
+        self.roles = [UserRole(role) for role in user_dict.get("roles", [])]
+        self.custom_roles = user_dict.get("customroles", [])
+        self.policy_ids = [
+            PolicyId(policy) for policy in user_dict.get("policyids", [])
+        ]
+        self.user = user_dict.get("user")
```
ialarmedalien, line 79: repeated

src/biokbase/auth.py
```diff
@@ -41,6 +41,48 @@ def __init__(self, info_dict: dict, token: str = None):
...
+
+
+class UserInfo:
+    """
+    A container for user information that comes from the Auth service. Does not
+    hold the token itself.
+    """
+
+    def __init__(self, user_dict: dict):
+        self.user = user_dict.get("user")
```
ialarmedalien, line 70: is this a user ID? user name? Would be good to ensure that `Identity.provider_user` and `UserInfo.user` are both of the same type, either IDs or names.

MrCreosote: Can I ask why you need the identity provider user name? That really just an informational thing to help the user know what links to what. It's not something I'd expect to be used in any other KBase code

briehl: I'm not sure we need any of it. I'm kinda just trying to be thorough and future-proofing here. Really, all that's ACTUALLY needed at this time is the user id, display name, and anon id.

src/biokbase/auth.py
briehl: Yup. Changed.

src/biokbase/auth.py
MrCreosote: In that case I would strongly recommend not making it part of the local state and adding a comment to the effect that the provider username should only be used in the context of the user trying to figure out which accounts are linked to their KBase account

briehl: I don't think any of that's actually necessary. I'll just trim that object down to what's needed and we can expand in the future if we need to. Which I doubt, as this is the first change to the auth client in ~6 years.

ialarmedalien (approved):

PR #3443, opened 2024-03-28
Fixes PTV-1909 - Narrative-based sign out does nothing

eapearson:

The Sign Out option of the login menu (upper right corner) does nothing when selected. Well, actually, it redirects to `/`, meaning, the root of Europa (previously kbase-ui), which in turn shows the default Narrative Navigator.

The latter behavior is not addressed in this ticket, but is due to the failure to remove the cookie ... for after the cookie is removed and the token revoked the ipythonLogout function is called, which redirects the browser to `/`, which will be the root of Europa, which will redirect to `/narratives`. When signout succeeds, redirecting to /narratives results in the sign in page. But when signout fails, the Narrative Navigator is indeed displayed.

Back to the problem that brought us together here. Which is that Narrative assumes the KBase auth cookie `kbase_session` is in host-mode (only works on a specific domain), which is how kbase-ui treated the auth cookie, but Europa now uses the auth cookie in domain-mode. In this mode, the cookie is available on the given domain and any subdomains. A cookie can only be removed when it is accessed in the same mode. (Well, removing a cookie is updating it with a settings that makes it invalid.) So, when the Narrative attempts to remove the cookie in host-mode, it does not affect the cookie in domain-mode, and the auth cookie remains untouched.

A second problem is that the code that revokes the token, executed immediately after the code to remove it from the browser, was using an invalid api method call. The url path for the auth revoke method was incorrect, so the token was not being revoked.

Finally, a small issue with usage of the auth revoke method was that it is asynchronous (returns Promise), but was not being called with promise chaining nor await (which is what I chose). So I made the containing function async and used await. This allowed me to catch the error, and ultimately fix the bug. I’m leaving this in, as it also makes that function more sensible (as previously the rest of the function would merrily proceed as the auth revoke function churned away in the background).

---

Well, now finally! In order to get the Narrative to build, I had to deal with the partially updated package.json.

I fixed these issues, but don’t consider a set of changes to be retained, as I know the desired fix at least for the node version is in the base image. Still, it won’t build without these changes.

There were multiple issues:

- several dependency updates require node >= 18; the image is node 16, so it won’t build as-is
- the `msw` version updated to does not provide a UMD build, and the `install-npm` script fails (and cannot succeed!) because the expected file to be moved does not exist, and there is no candidate.
- the narrative image has node 16, which is incompatible with the package.json

So here is what I did:
 - reverted `msw` to the last working version `0.35.0` (I stopped using `msw` in kbase-ui when this happened, and created a Python integration testing tool)
 - set “engine” in `package.json` to 20, rebuilt `package-lock.json` with nodejs 20
 - updated `Dockerfile` to install nodejs 20.12.0

I know there is an effort to add nodejs 20 to the Narrative base image

# Jira Ticket / Issue #

Related Jira ticket: https://kbase-jira.atlassian.net/browse/PTV-1909

# Testing Instructions

# Dev Checklist:


.github/workflows/integration_test.yml
```diff
@@ -47,10 +47,10 @@ jobs:
       shell: bash -l {0}
       run: docker pull ${{ env.APP_IMAGE_TAG }}
     -
-      name: Use Node JS 16
-      uses: actions/setup-node@v3
+      name: Use Node JS 20
+      uses: actions/setup-node@43
```
briehl, line 51: I think this should be `setup-node@v4`

briehl (changes requested):
Looks good! Thanks for this.
Just a typo in one of the github action scripts.

.github/workflows/integration_test.yml
eapearson: Ah, shite, indeed. @v4 it should be!

eapearson:
@briehl Addressed the code analysis flags, and the typo you noticed.

briehl (approved):

briehl:
There's a number of errors in unit tests here due to other things. I think we should merge this now and fix those in a future PR or two.

PR #3525, opened 2024-05-14
Attempt to fix finicky codecov

ialarmedalien:
Updating codecov call and passing in secret in the vain hope that it will fix the test coverage upload that keeps failing.

ETA: it worked... but that might just be a coincidence.


briehl (approved):
I could've sworn those changes were already in there, but I guess it was missing the `secrets` declaration. Urgh.

Thanks for working on this. Let's try it out in other PRs!

PR #3663, opened 2024-09-24
Fix WebdriverIO configs

briehl:

The last batch of JS updates played some havoc with the webdriverio configs.
Mainly, in a previous major version of wdio, they started wrapping the most popular browser drivers into wdio itself, which made the `wdio-chromedriver-service` obsolete. Which, it hasn't been updated in 2 years and has been marked obsolete on Github, so that's a good thing anyway.

But the configs needed a little patching and updating to tell wdio we're not using that service anymore.

Eventually, we'll need to do the same for the standalone-selenium-service (which was put in to talk to firefox), and do some fiddling with browserstack (which I don't think we're actually using).

Chrome alone gets used in GHA, so that gets priority here.

There's a bunch of other little bits of polish to add to the config situation, too, but that's a later project. This does need to fixed, though to get integration tests passing again, and so I can put in more to handle this auth business.

### DIFF
--- a/test/integration/specs/narrative_basic_test.js
+++ b/test/integration/specs/narrative_basic_test.js
@@ -1,12 +1,12 @@
 'use strict';
-const { login, makeURL } = require('../wdioUtils');
+const { login } = require('../wdioUtils');
 const env = process.env.ENV || 'ci';
 const testData = require('./narrative_basic_data.json');
 const testCases = testData.envs[env];
 
 describe('Narrative tree page with login', () => {
     beforeEach(async () => {
-        await browser.setTimeout({ implicit: 30000 });
+        await browser.setTimeout({ pageLoad: 30000 });
         await browser.reloadSession();
         await login();
     });
@@ -16,7 +16,7 @@ describe('Narrative tree page with login', () => {
     });
 
     it('opens a narrative', async () => {
-        await browser.url(makeURL(`narrative/${testCases.CASE_1.narrativeId}`));
+        await browser.url(`/narrative/${testCases.CASE_1.narrativeId}`);
         const loadingBlocker = await $('#kb-loading-blocker');
         const loadingText = await loadingBlocker.getText();
         await expect(loadingText).toContain('Connecting to KBase services...');
--- a/test/integration/wdio.conf.js
+++ b/test/integration/wdio.conf.js
@@ -4,8 +4,6 @@
 const testConfig = require('../testConfig');
 const fs = require('fs');
 
-const CHROME_BINARY = require('puppeteer').executablePath();
-
 /**
  * Given a preset key, return set set of common configuration keys for a given service, os, and browser
  * This is useful because testing services support a limited number of browser dimensions, which differs
@@ -176,7 +174,6 @@ const serviceConfigs = {
         logPath: 'selenium-standalone-logs',
         // drivers
     },
-    chromedriver: {},
     browserstack: {
         browserstackLocal: true,
         opts: {},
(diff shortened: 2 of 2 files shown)


test/integration/specs/narrative_basic_test.js
```diff
@@ -16,7 +16,7 @@ describe('Narrative tree page with login', () => {
     });
 
     it('opens a narrative', async () => {
-        await browser.url(makeURL(`narrative/${testCases.CASE_1.narrativeId}`));
```
briehl, line 19: `makeURL` (which just prepends the BASE_URL) isn't necessary. wdio does that for us. Was fiddling around here when I found that - will update other specs in a later PR.

test/integration/specs/narrative_basic_test.js
```diff
@@ -1,12 +1,12 @@
 'use strict';
-const { login, makeURL } = require('../wdioUtils');
+const { login } = require('../wdioUtils');
 const env = process.env.ENV || 'ci';
 const testData = require('./narrative_basic_data.json');
 const testCases = testData.envs[env];
 
 describe('Narrative tree page with login', () => {
     beforeEach(async () => {
-        await browser.setTimeout({ implicit: 30000 });
```
briehl, line 9: `implicit` timeouts shouldn't be used, according to docs.

test/integration/wdio.conf.js
```diff
@@ -4,8 +4,6 @@
 const testConfig = require('../testConfig');
 const fs = require('fs');
 
-const CHROME_BINARY = require('puppeteer').executablePath();
```
briehl, line 7: wdio fetches whatever version you tell it to, now, and does the pathing itself.

This might also make the puppeteer dependency obsolete, will check in a later PR.

test/integration/wdio.conf.js
```diff
@@ -196,21 +193,19 @@ function makeCapabilities(config) {
         case 'chromedriver':
             return (() => {
                 const args = [
-                    '--disable-gpu',
-                    '--no-sandbox',
+                    'headless',
+                    'disable-gpu',
```
briehl, line 197: Updating args according to docs.

test/integration/wdio.conf.js
```diff
@@ -196,21 +193,19 @@ function makeCapabilities(config) {
...
                     `window-size=${config.WIDTH},${config.HEIGHT}`,
                 ];
-                if (config.HEADLESS === 't') {
-                    args.push('--headless');
-                }
                 return {
                     browserName: 'chrome',
+                    browserVersion: '128',
```
briehl, line 202: latest version of Chrome. Not sure yet if we can add a spread of these (or if we need to, I think our policy is to stick to the latest).

That should probably get configured elsewhere rather than as a magic number, but... later.

test/integration/wdio.conf.js
```diff
@@ -196,21 +193,19 @@ function makeCapabilities(config) {
...
                     acceptInsecureCerts: true,
                     maxInstances: 1,
                     'goog:chromeOptions': {
                         args,
-                        binary: CHROME_BINARY,
                     },
+                    'wdio:enforceWebDriverClassic': true
```
briehl, line 208: The latest wdio uses the w3-specced webdriver BiDi protocol. There's something strange with either our configuration, or the mash of versions of things, that prevents that protocol from loading pages from our app. So forcing it back to webDriverClassic fixes that (at least, until they ditch that all together).

test/integration/wdio.conf.js
```diff
@@ -322,7 +317,7 @@ const wdioConfig = {
     // and 30 processes will get spawned. The property handles how many capabilities
     // from the same test should run tests.
     //
-    maxInstances: 10,
+    maxInstances: 1,
```
briehl, line 320: It'd be really nice to have this run parallel, but those runs seem to clobber each other. I suspect shenanigans with cookies getting globally added and removed, and communication with the single narrative application we have running. I think it's worth trying to fix that by ensuring that each test spec only uses a single narrative, but this PR's just about returning to status quo.

test/integration/wdio.conf.js
```diff
@@ -392,7 +387,6 @@ const wdioConfig = {
     // Services take over a specific job you don't want to take care of. They enhance
     // your test setup with almost no effort. Unlike plugins, they don't add new
     // commands. Instead, they hook themselves up into the test process.
-    services: [[testParams.SERVICE, serviceConfigs[testParams.SERVICE]]],
```
briehl, line 395: These told wdio to use the chromedriver service, which doesn't exist anymore, and was crashing.

ialarmedalien (approved):

PR #3732, opened 2026-07-07
Remove hardcoded 14-day token lifetime from auth cookie handling

dauglyon:
## Summary
KBase is extending the login token lifetime from **two weeks to four weeks**. `api/auth.js` defined `DEFAULT_TOKEN_LIFE = 14 * 24 * 60 * 60 * 1000` and used it as a fallback cookie expiration in `setCookie` whenever no `expires` was supplied. After the change, that fallback would silently keep cookies pinned to the old 14-day value.

Rather than bump the constant to 28 days, this removes it. Cookie expiry should always come from the auth service (the token's real expiration), never a hardcoded client-side lifetime.

## Why this is safe
- The production **login** path (`setAuthToken`) and **logout** path (`removeCookie`) already pass an explicit `expires`, so they are unaffected.
- The **only** caller relying on the 14-day fallback was the localhost-only dev token-injection dialog.

## Changes
- `kbase-extension/static/kbase/js/api/auth.js` — remove `DEFAULT_TOKEN_LIFE`; `setCookie` now requires an integer `expires` and throws if one isn't provided. Docstring updated.
- `kbase-extension/static/kbase/js/narrativeLogin.js` — the dev token-injection dialog now looks up the pasted token's real expiration via `getTokenInfo` and passes it to `setCookie`; an invalid/unvalidatable token is reported instead of silently setting a bad cookie.
- `test/unit/spec/api/authSpec.js` — the direct `setCookie` call (backup-cookie simulation) now passes an explicit `expires`.

## Context
Companion to kbase/ui#259, which updates the user-facing "two weeks" copy. Reviewed narrative + kbase-ui (all plugins); this was the only hardcoded lifetime assumption in narrative. All other expiry logic already reads the server-provided value.


briehl (approved):
Ok, lots of Narrative infrastructure updates later, this builds and tests fine.

Let's merge and try it on CI.

